PR #2, opened 2018-03-21
docs: add instructions to readme

weblancaster:


README.md
```diff
@@ -1,19 +1,124 @@
...
+For more on this topic, and examples we recommend
+[Better Specs](http://www.betterspecs.org/).
+
+```
+import React from "react";
+import Badge from ".";
+
+describe('Badge', () => {
+  it("render button with children", () => {
+    const badge = mount(<Badge>Wow!</Badge>);
```
Poltergeist: Is this example functional? Or do we also need to import mount here?

weblancaster: I will change this to a real example.

```
it("default", () => {
    expect(renderer
    .create(<Badge>default</Badge>)
    .toJSON()).toMatchSnapshot()
  });
```

PR #3, opened 2018-03-23
feat: script to generate component

weblancaster:
Add script to generate component based on template.
Run npm run create:component -- NewComponentName

Closes DCOS-21786

### DIFF
--- a/scripts/create-component
+++ b/scripts/create-component
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+components_path=components
+template_path=scripts/templates/component/*
+
+mkdir -p components/$1
+cp -r $template_path ${components_path}/$1
+
+FILES=$(find ${components_path}/$1 -regex '.*Component.*')
+for file in $FILES;
+do
+  mv -v $file "${file/Component/$1}";
+done
--- a/scripts/create-component
+++ b/scripts/create-component
@@ -1,13 +1,63 @@
-#!/usr/bin/env bash
+#!/usr/bin/env node
+let fs = require("fs-extra");
+let path = require("path");
+let componentName = "Component";
+const templatePath = "templates/component/";
 
-components_path=components
-template_path=scripts/templates/component/*
+var program = require("commander");
 
-mkdir -p components/$1
-cp -r $template_path ${components_path}/$1
+function readDir(dirPath) {
+  const dir = fs.readdirSync(path.resolve(__dirname, dirPath));
+  dir.forEach(location => {
+    if (
+      fs.lstatSync(path.resolve(__dirname, dirPath, location)).isDirectory()
+    ) {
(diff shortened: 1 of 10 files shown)


scripts/templates/component/stories/Component.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import Component from "../Component";
+
+const ComponentReadme = require("../README.md");
+
+storiesOf("Badge", module)
```
DanielMSchmidt: This needs to be Component then

scripts/create-component
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
DanielMSchmidt: Can we stop writing bash scripts and write this in JS? Bonus points for a small test case 👍

scripts/templates/component/Component.test.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+import Component from '.';
```
DanielMSchmidt: How does that work if component is not an index file?

scripts/templates/component/Component.test.tsx
```diff
@@ -0,0 +1,7 @@
...
+
+describe('Component', () => {
+  it.skip('test something', () => {});
```
DanielMSchmidt: This is exactly the same as ```it('test something')```

scripts/templates/component/Component.test.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+import Component from '.';
+
+describe('Component', () => {
```
DanielMSchmidt: Sorry I am whining about this right now, but aren't we using double quotes in dcos-ui? Would love to have it the same

Poltergeist:
I'll pick this up and do some changes.

tslint.json
```diff
@@ -6,6 +6,11 @@
     "tslint-config-prettier",
     "tslint-plugin-prettier"
   ],
+  "linterOptions": {
+    "exclude": [
+      "./scripts/*"
```
DanielMSchmidt: Let's make this scripts/templates/* 👍

scripts/create-component
```diff
@@ -1,13 +1,63 @@
-#!/usr/bin/env bash
+#!/usr/bin/env node
+let fs = require("fs-extra");
```
weblancaster: no big deal but should we use const on the requires?

tslint.json
weblancaster: Especially now we will be adding js scripts.

PR #4, opened 2018-03-23
feat: add GH pages

weblancaster:
Use storybook-deployer to make UI Kit accessible through URL.
https://dcos-labs.github.io/ui-kit/

Closes DCOS-21776

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "commitlint": "./node_modules/.bin/commitlint -e",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
+    "build:storybook": "build-storybook",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "commitlint": "./node_modules/.bin/commitlint -e",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
+    "build:storybook": "build-storybook",
+    "deploy:storybook": "./node_modules/.bin/storybook-to-ghpages",
```
DanielMSchmidt: I think you can leave out `./node_modules/.bin`, npm adds this automatically to your path at the beginning, so it's the same with and without it 👍

Poltergeist: I fixed this!

PR #5, opened 2018-03-23
DCOS-21772: add CI/CD

DanielMSchmidt:
- [X] [Green build without deployment](https://jenkins.mesosphere.com/service/jenkins/job/Frontend/job/UI-Kit%20Pipeline/view/change-requests/job/PR-5/2/)
- [X] [Green build with deployment](https://jenkins.mesosphere.com/service/jenkins/job/Frontend/job/dcos-labs/job/UI-Kit%20Pipeline/view/change-requests/job/PR-5/15/console)
- [ ] Get Jenkins to report the status here

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
-FROM node:slim
+FROM node:8.9.4-alpine
+
+# For deploying the gh-pages
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh
 
 ENV CI true
 COPY package*.json ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM node:slim
+
+ENV CI true
+COPY package*.json ./
+RUN npm install --ignore-scripts
+
+COPY . .
+EXPOSE 6006
+CMD ["npm", "start"]
+
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,77 @@
+#!/usr/bin/env groovy
+
+@Library('sec_ci_libs@v2-latest') _
+
+def master_branches = ["master", ] as String[]
+
+pipeline {
+  agent {
+    dockerfile {
+      args  '--shm-size=2g'
+    }
+  }
+
+  environment {
+    JENKINS_VERSION = 'yes'
+  }
+
+  options {
+    timeout(time: 1, unit: 'HOURS')
+  }
+
+  stages {
+    stage('Authorization') {
+      steps {
+        user_is_authorized(master_branches, '8b793652-f26a-422f-a9ba-0d1e47eb9d89', '#frontend-dev')
+      }
+    }
+
+    stage('Initialization') {
+      steps {
+        ansiColor('xterm') {
+          retry(2) {
+            sh '''npm install'''
+          }
+        }
+      }
+    }
+
+    stage('Unit Test') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run test'''
+        }
+      }
+    }
+
+    stage('Build') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run dist'''
+        }
+      }
+
+      post {
+        always {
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:slim
```
weblancaster: Let's use a specific version of node e.g `FROM node:8.9.4-alpine`

DanielMSchmidt: Can you explain why?

DanielMSchmidt:
Deployment doesn't seem to work because the repo is not cloned. I am honestly unsure if it wouldn't be better making it a `postrelease` hook instead 🤔 

https://jenkins.mesosphere.com/service/jenkins/job/Frontend/job/UI-Kit%20Pipeline/view/change-requests/job/PR-5/5/

(working around is possible but not so nice)

Dockerfile
weblancaster: Sure, There is no way to know what version of node we are using when `FROM node:slim` 
Would be nice to always install the version required/suggested and easy to know from the dockerfile

weblancaster:
@DanielMSchmidt we want to have the storybook build/GH pages updated on each PR merged.
The UI Kit page should have the most recent from master. Basically a "nightly" build.

Let's try to figure out a way to have it done. I have a couple ideas how we can handle that so I will set a meeting for us including @Poltergeist

weblancaster:
BTW, this is really good, thanks for opening the PRs! 🎉 ❤️

Dockerfile
DanielMSchmidt: Well, I don't have such a such a strong position there, I just like to keep these things up-to-date :) But I'll change it and we make sure to update it on a regular basis

DanielMSchmidt:
@weblancaster I think the only thing we still need is to store a GH token for this repo in our Jenkins secrets 👍

Dockerfile
weblancaster: 🙏

Jenkinsfile
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      post {
+        always {
+          archiveArtifacts 'dist/*'
+        }
+      }
+    }
+
+    stage('Deploy'){
```
weblancaster: We should only deploy when a PR is merged so probably a separated hook. Seems like here will deploy everytime there is a PR successful merged or not merged, right @DanielMSchmidt?

Jenkinsfile
```diff
@@ -0,0 +1,77 @@
...
+      when {
```
weblancaster: We should only deploy when a PR is merged so probably a separated hook. Seems like here will deploy every time there is a PR successful merged or not merged, right @DanielMSchmidt?

DanielMSchmidt: Nope, the when clause with the expression ```env.BRANCH_NAME == "master"``` you highlighted here causes Jenkins to skip the step if it is run from a PR and not on master

DanielMSchmidt: You can see it in action in the builds I triggered: 

[Here you can see the deploy is not done](https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-labs%2FUI-Kit%20Pipeline/detail/PR-5/2/pipeline/)
bildschirmfoto 2018-03-27 um 09 11 36



[Here you can see I commented the line out, so it's done](https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-labs%2FUI-Kit%20Pipeline/detail/PR-5/15/pipeline)
bildschirmfoto 2018-03-27 um 09 11 39

weblancaster: Awesome, thanks for explaining

DanielMSchmidt: No worries 👍

DanielMSchmidt:
@weblancaster I don't know if we need hold merge. It is done from a code perspective, only permissions in Jenkins / GH need to be updated :)

PR #7, opened 2018-03-23
Setup jest

DanielMSchmidt:
We can't use [story shots](https://github.com/storybooks/storybook/tree/master/addons/storyshots) because of other plugins we use. Maybe we
should take the time to decide if
a) this is important to us and
b) we can find a way around this

Closes DCOS-21785


components/badge/__tests__/badge.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
weblancaster: Can you rename the file to `badge.test.tsx` please?

DanielMSchmidt: feel free to do it yourself 👍 (because I am OOO for today ;))

weblancaster: I will do, thanks.

weblancaster:
@DanielMSchmidt let's not use `story shots` for now maybe one day when we find more value but at the moment we are pretty good with Jest Snapshot/Unit tests

DanielMSchmidt:
@weblancaster I agree with you, although storyshots also uses jest & snapshot testing, just fully automated from your storybook stories 🙈 But yeah, let's improve tooling when we need it

DanielMSchmidt:
I also added tshint as pretest step.

PR #13, opened 2018-04-03
feat: implement badge component

weblancaster:
This is the implementation of `badge` component.

**How to test:**
- `rm -rf node_modules && npm i`
- `npm run storybook`


components/badge/README.md
```diff
@@ -1,51 +1,23 @@
...
-
-```
-<!-- badge: Default -->
-<span class="badge badge-rounded">
-  Default
-</span>
-```
+|Do|Don't|
+|---	|---	|
+|Use rounded rectangles.|Use fully rounded corners (we no longer use this style).|
```
DanielMSchmidt: I don't get this section. If it is a component giving me a badge, why should I care about if it uses fully rounded corners? I just take it and place it somewhere, right?

components/badge/style.ts
```diff
@@ -1,37 +1,48 @@
 import { css } from "emotion";
-import { getColors, getFonts } from "../../shared/style";
+import { getColors, getFonts } from "../../shared/styles/core";
```
DanielMSchmidt: I love that we have colors at one place under control 👍

components/badge/style.ts
```diff
@@ -1,37 +1,48 @@
...
   "default": css`
     background-color: ${getColors().greyLight};
     border-color: ${getColors().greyLight};
     color: #1b2029;
   `,
   "success": css`
-    background-color: #14c684;
-    border-color: #14c684;
-    color: #ffffff;
+    background-color: ${getColors().green};
```
DanielMSchmidt: Why wouldn't we explode the colors, like this


```js
const { green, white, yellow } = getColors();
...
const theme = {
	"success": css`
		background-color: ${green}
	`
}
```

components/badge/README.md
```diff
@@ -1,51 +1,23 @@
 # Badges
 
-Badges are a lightweight method of annotating or labeling content. Wrap content in the `.badge` class to add an inline badge. The font-size and line-height of a badge inherit from that of it's parent's properties.
+Badges are a lightweight method of annotating or labeling content. Wrap content in the .badge class to add an inline badge. The font-size and line-height of a badge inherit from that of it’s parent’s properties.
```
Poltergeist: Here we are talking about a class, but probably we want to mention a component.

components/badge/README.md
Poltergeist: This also seems like a CSS description which we do not need anymore since we use a component, right?

components/badge/README.md
```diff
@@ -1,51 +1,23 @@
...
+|Success|Represents something positive.|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-success.png)|
+|Primary|Represents something more significant than a default.|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-primary.png)|
+|Danger|Represents something in a danger or error state.|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-danger.png)|
+|Warning|Represents something that we want to warn the user about but not quite extreme as a danger state.|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-warning.png)|
+|Outline|Outline badges for when we want the density of the badge to be lighter e.g. when next to data in a table cell|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-outline.png)|
+|With Icon|If need be we can add icons and tooltips to add further context.|![Preview](https://raw.githubusercontent.com/dcos-labs/ui-kit/master/components/badge/assets/badge-icon.png)|
```
Poltergeist: Do we need to mention this here? Badge is composable which means it takes every possible children component.

components/badge/style.ts
Poltergeist: 👍 However, we should not name them green, red, blue, dark-green, darker-green ... There are a lot of shades of green :-). Maybe we can use something like https://github.com/ryanzec/name-that-color .

components/badge/style.ts
Poltergeist: Like that idea.

components/badge/style.ts
```diff
@@ -1,37 +1,48 @@
 import { css } from "emotion";
-import { getColors, getFonts } from "../../shared/style";
+import { getColors, getFonts } from "../../shared/styles/core";
 
-const badgeAppearances = {
+const badgeTheme = {
   "default": css`
     background-color: ${getColors().greyLight};
     border-color: ${getColors().greyLight};
     color: #1b2029;
```
Poltergeist: Maybe we can use a named color here as well.

components/badge/style.ts
weblancaster: @Poltergeist this are the same name patterns used in DCOS UI/CNVAS and what the designers use

components/badge/README.md
weblancaster: Yes seems like an old description. I will talk to Lee because this was in the docs.

components/badge/README.md
weblancaster: Agree. I have copied the docs Lee provided to us I will talk to him so we are aligned.

components/badge/README.md
weblancaster: Agree, but we don't want every possible child.

components/badge/style.ts
weblancaster: hah I forgot I can use `destructuring`. Thanks.

Poltergeist:
I am okay with anything else them theme. Theme describes something different like a set of colors. As soon as we add the option for theming the theme prop is going to be very confusing as well. On that agenda, `appearance` is okay with me.

components/badge/README.md
Poltergeist: Well since the badge component is a `span` this is excluding already a lot of elements and only including inline elements like `em` and `strong` for example. But then again we do not care about which children are passed to the component this is the concern of the user not of the library. And there is no sane way to check for valid components.

components/badge/style.ts
Poltergeist: Okay, maybe we keep that change for later.

components/badge/style.ts
Poltergeist: If this is not a color named from Design we could name this `ebonyClay`. It just feels cleaner to have all colors named.

components/badge/README.md
weblancaster: Yes agree, although we want to document and have an example.

Having this will only help.

Poltergeist:
Let's get this merged! Awesome work. And our first component is done :-)

PR #14, opened 2018-04-04
Change build steps for a smoother publish

Poltergeist:
This PR makes improvements in the way we build the `dist` folder which now we include the type definitions in the package.

**How to test:**
- `npm run dist`
- Note the dist folder contain the `js` and `d.ts` files from each raw `ts/tsx`.

<details><summary><strong>chore(dependencies): update and add dependencies</strong></summary><hr/>

Update dependencies to latest versions.
  - `@types/react-dom`

Add new dev dependency
  - `rimraf`
    This is a OS independent alternative to `rm -rf`

Move dependencies
  - `node` to dev
  - `npm` to dev

The move was necessary because they also got installed when the user installed the package.
</details>

---
Commit:
b4555aa31f3270d1e4c5273e5989eb9a59349319
<details><summary><strong>chore(export): change export syntax style</strong></summary><hr/>

This changes the syntax of the exports to be less verbose.
</details>

---
Commit:
340802eaef7f45e4ecdb9e293846ae770649f76a
<details><summary><strong>chore(namespace): add `@dcos` as a namespace to the package name</strong></summary><hr/>

</details>

---
Commit:
ab69a1f9103e66d9e80b1fb0340f32fe4cc29eca
<details><summary><strong>chore(tsconfig): change tsconfig to improve build</strong></summary><hr/>

This adds declarations to the build which unfortunately also means that we need to disallow JS.
Also this changes the libs which are included and removes some redundancy.
The build also only includes components which are linked in the `index.js` file.
The build will exclude test and story files as they are not needed as a dependency.
</details>

---
Commit:
2f591e82dec163ff0f4c2e5d6e5320576bd9a0a0
<details><summary><strong>chore(git-hook): change git hook setup script</strong></summary><hr/>

The git hook is now installed by running
`npm run setup:git-hook`
this was necessary as by installing the package npm was trying to setup the git hooks as well in the
other project.
</details>

---
Commit:
f76bb7c362a7cbd614baab01653bfb2c23a216a3
<details><summary><strong>chore(dist): change dist npm script to use tsc instead of webpack</strong></summary><hr/>

The dist script now uses the typescript compiler instead of webpack by that we are now supporting a
handful of options for the user to include a component
1. `import Badge from "@dcos/ui-kit/dist/components/badge";`
   Direct importing will only include the code necessary for that component into the build of the
   application.
2. `import { Badge } from "@dcos/ui-kit";`
   Destruct import will include the complete ui-kit library in the build of the application this is
   only recommended if most of the components are actually used in the application to prevent build
   bloat as tree shaking will also not properly work.

2.5. `import { Badge } from "@dcos/ui-kit/dist/components";`
     This is basically the same import as in 2. but more verbose.

close DCOS-21892 DCOS-21891
</details>

---
Commit:
ad08a923b4783a9ee0ef333b625aca7073bd5452
<details><summary><strong>chore(npm-scripts): clean up npm scripts</strong></summary><hr/>

Change `npm` scripts to be alphabetical sorted, also spreading some scripts into pre scripts.
</details>

---
Commit:
4892c3bb0483c3d8b60dcfee4073ae5c1f45cc66
<details><summary><strong>chore(jenkinsfile): change artifact check</strong></summary><hr/>

This changes the glob jenkins is checking after the build step from `dist/*` to `dist/**/*`
</details>

---
Commit:
fdb46f696b4082764462ee20193799f7753a871b
<details><summary><strong>chore(tsconfig): add tsconfig.dist.json</strong></summary><hr/>

This adds a tsconfig for the dist creation and also readds `allowJs` option to `tsconfig.json` so
tsconfig.json is used for storybook and tsconfig.dist.json is used to create the bundle.
</details>

---
Commit:
5cda098b31221b0cf46ee5c0045ee689af143369


Poltergeist:
I will need to add a second tsconfig. one for `storybook` and one for `dist`

weblancaster:
Hey @Poltergeist couple things I noticed and made me a bit concern if this is the ideal solution we have

1. A lot of unnecessary code is shipped in dist (e.g shared, component styles).
2. Modules required in the files becomes a dependency in DCOS UI, one example in this case is in the component folder we have a style file that requires emotion, when using this way the component style isn't built as a single source.
3. The code isn't optimized, ofuscated/minified.

Although something that I liked is that the fact of definitions created and shiped together.

package.json
```diff
@@ -1,51 +1,43 @@
...
+    "create:component": "./scripts/create-component",
     "deploy:storybook": "storybook-to-ghpages",
-    "clean": "rm -rf dist",
+    "dist": "tsc -p tsconfig.dist.json",
+    "predist": "npm run clean",
+    "lint": "tslint -c tslint.json --project tsconfig.json",
+    "prerelease": "npm run dist",
     "release": "standard-version",
-    "create:component": "./scripts/create-component"
+    "setup:git-hooks": "./scripts/pre-install",
```
weblancaster: why was this moved from automatic setup `preinstall` to a custom script `setup:git-hooks`?

tsconfig.dist.json
```diff
@@ -0,0 +1,21 @@
+{
```
weblancaster: I don't understand why we need another `tsconfig`

weblancaster: I don't understand why we need another `tsconfig`

Poltergeist:
We need another tsconfig since declarations and allowJs are excluding each other and we need allowJs for storybook. And declarations for the dist.

package.json
Poltergeist: As you might have seen already in the commit message of commit f76bb7c and the PR description the `preinstall` script is not only getting run when the user is running `npm install` within the `ui-kit`  directory but also if you install the package as a dependency. If the `scripts` directory is not included in the distribution this command will fail which will fail the install. If we include the scripts directory in the distribution the git hook gets installed unnecessarily. I thought we might not want this. this is why I rather used a custom script.

tsconfig.dist.json
Poltergeist: We need another `tsconfig` since declarations and `allowJs` are excluding each other and we need `allowJs` for storybook. And declarations for the dist.

Poltergeist:
Hey @weblancaster I would like to address your concerns 

> 1. A lot of unnecessary code is shipped in dist (e.g shared, component styles).

Do you mean unnecessary files as in we do not want to require these files in DCOS UI directly, or that the code in these files are not needed? Because that might be a wrong assumption. If you have an idea how we can combine certain files in a simple way I am all ear.  

> 2. Modules required in the files becomes a dependency in DCOS UI, one example, in this case, is in the component folder we have a style file that requires emotion, when using this way the component style isn't built as a single source.

Modules defined in the dependency section of the `package.json` do become a dependency in DCOS UI. The difference if we would build a complete package into a single source would mean that this package would also include everything needed, like react and emotion. So I don’t see any difference by delivering the components in that way or in a complete package. Regarding the dependencies.

> 3. The code isn't optimized, ofuscated/minified

This is not really a problem since the code gets optimized and minified on the build of DCOS UI. 

This is the Badge component as it would turn up in a compiled project

```JS
function(e, t, n) {
  "use strict";
  var r =
      (this && this.__extends) ||
      (function() {
        var e =
          Object.setPrototypeOf ||
          ({ __proto__: [] } instanceof Array &&
            function(e, t) {
              e.__proto__ = t;
            }) ||
          function(e, t) {
            for (var n in t) t.hasOwnProperty(n) && (e[n] = t[n]);
          };
        return function(t, n) {
          function r() {
            this.constructor = t;
          }
          e(t, n),
            (t.prototype =
              null === n
                ? Object.create(n)
                : ((r.prototype = n.prototype), new r()));
        };
      })(),
    o =
      (this && this.__importStar) ||
      function(e) {
        if (e && e.__esModule) return e;
        var t = {};
        if (null != e)
          for (var n in e) Object.hasOwnProperty.call(e, n) && (t[n] = e[n]);
        return (t.default = e), t;
      };
  Object.defineProperty(t, "__esModule", { value: !0 });
  var a = o(n(0)),
    i = n(29),
    l = (function(e) {
      function t() {
        return (null !== e && e.apply(this, arguments)) || this;
      }
      return (
        r(t, e),
        (t.prototype.render = function() {
          var e = this.props,
            t = e.appearance,
            n = e.children;
          return a.createElement("span", { className: i.badge(t) }, n);
        }),
        (t.defaultProps = { appearance: "default" }),
        t
      );
    })(a.PureComponent);
  (t.Badge = l), (t.default = l);
}  
```

And this is the style how it would turn up in a compied project:
```JS
function(e, t, n) {
  "use strict";
  var r =
      (this && this.__extends) ||
      (function() {
        var e =
          Object.setPrototypeOf ||
          ({ __proto__: [] } instanceof Array &&
            function(e, t) {
              e.__proto__ = t;
            }) ||
          function(e, t) {
            for (var n in t) t.hasOwnProperty(n) && (e[n] = t[n]);
          };
        return function(t, n) {
          function r() {
            this.constructor = t;
          }
          e(t, n),
            (t.prototype =
              null === n
                ? Object.create(n)
                : ((r.prototype = n.prototype), new r()));
        };
      })(),
    o =
      (this && this.__importStar) ||
      function(e) {
        if (e && e.__esModule) return e;
        var t = {};
        if (null != e)
          for (var n in e) Object.hasOwnProperty.call(e, n) && (t[n] = e[n]);
        return (t.default = e), t;
      };
  Object.defineProperty(t, "__esModule", { value: !0 });
  var a = o(n(0)),
    i = n(29),
    l = (function(e) {
      function t() {
        return (null !== e && e.apply(this, arguments)) || this;
      }
      return (
        r(t, e),
        (t.prototype.render = function() {
          var e = this.props,
            t = e.appearance,
            n = e.children;
          return a.createElement("span", { className: i.badge(t) }, n);
        }),
        (t.defaultProps = { appearance: "default" }),
        t
      );
    })(a.PureComponent);
  (t.Badge = l), (t.default = l);
}
```

If I compare that to the source of the compiled library, this will be similar code. 

And if that means that we are not able to have the type definitions and no source maps which will result in an unreadable error message like `e is not a function` I rather take the unminified code in the dev environment.

package.json
weblancaster: Let's just make it automated then so we don't have another step to remember/do when setting up this repository.

weblancaster:
@Poltergeist and I took offline.

tsconfig.dist.json
weblancaster: I pushed a commit addressing this one.

bstavroulakis:
Works as seen on TV

PR #19, opened 2018-04-10
chore(commit-msg): fix commit msg hook

Poltergeist:
This fixes the issue where a commit is not created if the commit msg is not fulfilling the
convetion. Which means that the commit lint is telling you to run `git commit --amend` to change the
message which means that the changes are appended to the previous commit. Also the message entered
before is lost, which means if someone has entered a rather long commit message that work would have
been lost.

### DIFF
--- a/scripts/commit-msg
+++ b/scripts/commit-msg
@@ -19,7 +19,7 @@ then
     "\nUse the following command to amend changes and edit the commit message."\
     "\n\n> git commit --amend\n"
 
-  exit -1
+  exit 0
 fi
 
 info "Commit message looks good"


scripts/commit-msg
```diff
@@ -19,7 +19,7 @@ then
     "\nUse the following command to amend changes and edit the commit message."\
     "\n\n> git commit --amend\n"
 
-  exit -1
+  exit 0
```
weblancaster: @Poltergeist I had chosen to use `exit -1` because I believe we don't want people committing messages that don't follow the guidelines.

Although I'm fine with this too.

PR #21, opened 2018-04-12
feat: add onClick event to badge

weblancaster:


components/badge/badge.test.tsx
```diff
@@ -50,4 +50,14 @@ describe("Badge", () => {
     .create(<Badge><StringComponent /></Badge>)
     .toJSON()).toMatchSnapshot()
   });
+
+  it("can have onClick event", () => {
+    const foo = () => {
```
DanielMSchmidt: I would love to see this inlined instead of using a random variable here :)

components/badge/badge.tsx
```diff
@@ -3,18 +3,34 @@ import { badge } from "./style";
...
+}
 export class Badge extends React.PureComponent<IBadgeProps, {}> {
   public static defaultProps: Partial<IBadgeProps> = {
     appearance: "default"
   };
 
   public render() {
-    const { appearance, children } = this.props;
+    const { appearance, onClick, children } = this.props;
+    const isOnClickOn = !!onClick;
```
DanielMSchmidt: Would you mind going with `Boolean(onClick)` here? I know it's more characters but it's also more intuitive IMHO

weblancaster: 👍

components/badge/badge.test.tsx
weblancaster: using arrow functions inline decrease performance and we have a rule in tslint to enforce nobody use it.

weblancaster:
@Poltergeist I hope I understood correctly your points.

> There is no test if the onClick event does work.

Correct, I added a quick snapshot test that contains the onclick and would add the tests later.

> We are adding a feature to a component which does not fit into the description. 

```  
Badges are a lightweight method of annotating or labeling content. 
```

Agree, although that doesn't mean it's not interactive/can't have actions.
I spoke with @leemunroe and he said it's ok `badges` in some cases having actions/events.

> We are adding unnecessary complexity by flexibility to a clean component.

Agree, despite the `onclick` code be small it is still code. (I have suggestions to your PR though)

> We are breaking a11y as we are not using semantic elements (a span has no native on click)

Yes I'm not following a11y standards, just keep in mind that there is a way to make it accesible even though it's not a button https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role (go to examples where there is a span being used)

> A span cannot be provided with a tab index which means that the badge cannot be focused and triggered via the keyboard.

It can have tab index, take a look at the example above for more info.

> I do have prepared a proposal which we could do instead it is shown on dcos-labs/ui-kit#23 we are adding the feature by creating a new component called `BadgeButton`.

I have feedback about your changes and I will leave my comments there.

PR #23, opened 2018-04-12
Add BadgeButton Component:

Poltergeist:
<details><summary><strong>feat(badgebutton): add badge button component</strong></summary><hr/>

This adds the `BadgeButton` component which is a badge as a button it embraces the same
functionality like the Badge component.
</details>

---
Commit:
03fef1c076e048a20f8a022d12991285f68ab6c7
<details><summary><strong>feat(badgebutton): add on click property to badgebutton</strong></summary><hr/>

This adds the on click property to the badgebutton and changes the element from a span to a button.
This also changes the style so that the badge button looks similar to the badge.
</details>

---
Commit:
a014425fc2f104f53aff42ae7fbb7beb134eba9f
<details><summary><strong>feat: implement badge button component</strong></summary><hr/>

Add the right type to onClick and add correct cursor.
Change badge style to also have top and bottom padding.
Use badge styles for the badgeButton.
Remove code duplication.
</details>

---
Commit:
0957d9c43fdec8c9977b91657eef3afdff95ca4d
<details><summary><strong>feat: change directory structure</strong></summary><hr/>

Change directory structure to be more representing the contents this adds a package directory on the
root level. Which contains multiple packages. Each package contains a components directory which
contains 1+n components. These are exported specifically by the index.ts file in the package
directory. Also the shared directory is a package now. Also the component generator has been
adjusted to the new structure.

BREAKING CHANGE: directory structure changed, badge is no more exported as a default.
</details>

---
Commit:
20500c58192fb664d479828a199e182b466f8701


weblancaster:
@Poltergeist I pushed a commit (https://github.com/dcos-labs/ui-kit/pull/23/commits/05d9cf4e7b124fbbf126dff1ea5e59afbae4d176) where removes all the duplicated code since badge button still under the umbrella `badge` (take a look at the danger preview, that's a button).

For more reference please check:
https://bitbucket.org/atlassian/atlaskit-mk-2/src/291d17eb74a93dd9cbeaf76f13fe93864a85328e/packages/core/button/src/index.js?at=master&fileviewer=file-view-default

Poltergeist:
I disagree partially with that, in the case of Atlaskit, `button` is a package by itself. And all components are sorted in the components directory. We decided that we will have one directory per component. We should keep up to that. 

What I am agreeing is that we do not necessarily need the duplicated code. So I propose that we move the `badgeButton` Component again out into its own directory, accompanying some tests. But using the styles from the button. Also adding the other padding definition from the CSS which was the only difference in the style. Which got deleted, unfortunately. Also by changing the export strategy in the `button/index.js` we introduced a breaking change. Let's not introduce a breaking change for something like that. I am going to add those changes to your commit.

PR #79, opened 2018-04-25
DCOS-22331: Add Clickable Component

Poltergeist:
<details><summary><strong>feat(clickable): add clickable component</strong></summary><hr/>

This adds a clickable component which adds click specific attributes to a child element and is using
a11y best practices.

Close DCOS-22331
</details>

---
Commit:
a5ddbf3b3a56fdf411dbf7be1cb7d3bbf0507b33
<details><summary><strong>docs(clickable): add clickable documentation</strong></summary><hr/>

This adds the documentation for the clickable component.
</details>

---
Commit:
d1c49bbc92ea2f0a0de4119bfb55c6a9f91e37e5

### DIFF
--- a/packages/clickable/README.md
+++ b/packages/clickable/README.md
@@ -1,0 +1,3 @@
+# Clickable Name
+
+Clickable documentation
--- a/packages/clickable/README.md
+++ b/packages/clickable/README.md
@@ -1,3 +1,5 @@
-# Clickable Name
+# Clickable
 
-Clickable documentation
+The Clickable is a higher order presentational component, children get the necessary a11y attributes added.
+
+Use this whenever you want to bind an action on click onto a HTML element. The component will add the appropiate role attribute, add a tabIndex, add a focus state and bind the action to `onClick` and `onKeyPress`. The `onKeyPress` has a wrapping function which filters out eveerything except the space key and enter key events. This prevents that the function gets triggered by accident.
--- a/packages/clickable/components/clickable.tsx
+++ b/packages/clickable/components/clickable.tsx
@@ -1,0 +1,50 @@
+import * as React from "react";
+
+import { cx } from "emotion";
+import { outline } from "../style";
+
+export interface IClickableProps {
+  /**
+   * Children should be a HTML element.
+   */
+  children: React.ReactElement<HTMLElement>;
+  /**
+   * Action is a event handler for the onClick and onKeypress events
+   */
+  action: (event?: React.SyntheticEvent<HTMLElement>) => void;
+  /**
+   * The tabIndex is passed down and is the same as the native tabIndex
+   */
+  tabIndex?: number | string;
+}
+
+class Clickable extends React.PureComponent<IClickableProps, {}> {
+  public static defaultProps: Partial<IClickableProps> = {
+    tabIndex: -1
+  };
+
+  constructor(props: IClickableProps) {
+    super(props);
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  public render() {
+    const { children, action, tabIndex } = this.props;
+    const { className = "" } = children.props;
+
+    return React.cloneElement(React.Children.only(children), {
+      onClick: action,
+      className: cx(className, outline),
+      tabIndex,
+      onKeyPress: this.handleKeyPress
+    });
+  }
+
+  public handleKeyPress(event: React.KeyboardEvent<HTMLElement>): void {
+    if (event.key === " " || event.key === "Enter") {
+      this.props.action(event);
+    }
+  }
+}
+
+export default Clickable;
--- a/packages/clickable/index.ts
+++ b/packages/clickable/index.ts
@@ -1,0 +1,1 @@
+export { default as Clickable } from "./clickable";
--- a/packages/clickable/stories/clickable.stories.tsx
+++ b/packages/clickable/stories/clickable.stories.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import Clickable from "../components/clickable";
+import { action } from "@storybook/addon-actions";
+
+const readme = require("../README.md");
+
+storiesOf("Clickable", module)
+  .addDecorator(withReadme([readme]))
+  .addWithInfo("default", () => (
+    <Clickable action={action("action trigger")} tabIndex="0">
+      <span>Click me!</span>
+    </Clickable>
+  ));
--- a/packages/clickable/tests/clickable.test.tsx
+++ b/packages/clickable/tests/clickable.test.tsx
@@ -1,0 +1,92 @@
+import React from "react";
+
+import { shallow } from "enzyme";
+import Clickable from "../components/clickable";
+
+describe("Clickable", () => {
+  it("has onClick function", () => {
+    const action = jest.fn();
+    const wrapper = shallow(
+      <Clickable action={action}>
+        <span>onClick</span>
+      </Clickable>
+    );
+    wrapper.simulate("click");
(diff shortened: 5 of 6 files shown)


packages/clickable/index.ts
```diff
@@ -0,0 +1 @@
+export { default as Clickable } from "./clickable";
```
DanielMSchmidt: Could you explain this level of indirection to me?

packages/clickable/components/clickable.tsx
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  constructor(_) {
+    super(_);
+    this.handleKeyPress = this.handleKeyPress.bind(this);
+  }
+
+  public render() {
+    const { children, action, tabIndex } = this.props;
+    const { className = "" } = children!.props;
```
DanielMSchmidt: TIL about [`!.`](https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)

packages/clickable/stories/clickable.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
```
DanielMSchmidt: Can we use `React from "react";` like we normally do?

packages/clickable/README.md
```diff
@@ -0,0 +1,3 @@
+# Clickable Name
+
+Clickable documentation
```
DanielMSchmidt: There seems to be documentation missing here

packages/clickable/index.ts
Poltergeist: What do you mean by indirection, do you mean that we export something from a different file? or the default as Clickable?

packages/clickable/components/clickable.tsx
Poltergeist: This is a Non-Null Assertation Operator

[read more here, should be the second section on the page](https://basarat.gitbooks.io/typescript/docs/options/strictNullChecks.html)

This is needed when something might be null, it is a typescript thing.

packages/clickable/index.ts
Poltergeist: But since I declared the children not to be optional anymore this is actually not needed anymore.

packages/clickable/components/clickable.tsx
Poltergeist: But since I declared the children not to be optional anymore this is actually not needed anymore.

packages/clickable/stories/clickable.stories.tsx
Poltergeist: Yes, we can at some point I thought we needed to do that.

packages/clickable/README.md
Poltergeist: Yes, I haven't written any documentation yet. :-/ And I should have done that before creating a PR.

packages/clickable/index.ts
DanielMSchmidt: I think now I get it :)

packages/clickable/README.md
DanielMSchmidt: Or just leave it in WIP and setting the label ;)

packages/clickable/stories/clickable.stories.tsx
Poltergeist: I found out why we need it. If we do not import complete react the prop documentation is not created.

packages/clickable/components/clickable.tsx
```diff
@@ -0,0 +1,50 @@
...
+   */
+  tabIndex?: number | string;
+}
+
+class Clickable extends React.PureComponent<IClickableProps, {}> {
+  public static defaultProps: Partial<IClickableProps> = {
+    tabIndex: -1
+  };
+
+  constructor(_) {
```
juliangieseke: can you give some explanation why you decided to use `_` as a parameter name here instead of something better understandable?

packages/clickable/tests/clickable.test.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
+
+import { shallow } from "enzyme";
+import Clickable from "../components/clickable";
+
+describe("Clickable", () => {
+  it("Clickable has onClick function", () => {
```
juliangieseke: shouldnt it be `it("has onClick…`? 🤔  (same for the following tests)

packages/clickable/README.md
```diff
@@ -0,0 +1,5 @@
+# Clickable
+
+The Clickable is a higher order presentational component, children get the necessary a11y attributes added.
+
+Use this whenever you want to bind an action on click onto a HTML element. The component will add the appropiate role attribute, add a tabIndex, add a focus state and bund the action to `onClick` and `onKeyPress`. The `onKeyPress` has a wrapping function which filters for the space key. This prevents that the function gets triggered by accident.
```
Fabs: :write: ...bind the action.

packages/clickable/components/clickable.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
+
+import { cx } from "emotion";
```
Fabs: from emotion 💘

packages/clickable/components/clickable.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return React.cloneElement(React.Children.only(children), {
+      onClick: action,
+      className: cx(className, outline),
+      tabIndex,
+      onKeyPress: this.handleKeyPress
+    });
+  }
+
+  public handleKeyPress(event: React.KeyboardEvent<HTMLElement>): void {
+    if (event.key === " " || event.key === "Enter") {
```
Fabs: the docs said that the space was to avoid trigger by accident, but here it looks like the space and the enter does the same, which would be triggering the action. What am I missing?

packages/clickable/index.ts
Fabs: I did not, can you explain why this file is necessary?

Poltergeist: This file is necessary to create consistency. 
We have multiple packages like `clickable`. And their components are located in `packages/clickable/clickable.tsx` in this case. We would like to import like this: `packages/clickable` which redirects to the index file. Which is basically exporting all the components. Within this package. 

We also do have the packages/index.js file. Which will be also exporting all the packages. So the user now gets multiple ways to import in the application:
 1. `import {Clickable} from "@dcos/ui-kit"; -- Which imports the complete library. Which is not shakeable even if you only use one component.
2. `import {Clickable} from "@dcos/ui-kit/packages/clickable"; -- Which imports the complete package. Which is not shakeable even if you only use one component, but it is only containing one package.
3. `import {Clickable} from "@dcos/ui-kit/packages/clickable/components/clickable"; -- Which imports only that exact component. Which is kind of the recommended way since this will only include in the build what is needed. 

At some point, we will split up the whole lib into multiple packages and then there would also be a 4 way to import everything which is: `import {Clickable} from "@dcos/ui-kit-clickable";

packages/clickable/components/clickable.tsx
Poltergeist: A button triggers on Space and Enter key. So the docs are wrong will clarify.

packages/clickable/README.md
Poltergeist: 👍

packages/clickable/README.md
```diff
@@ -0,0 +1,5 @@
+# Clickable
```
weblancaster: 👍 🏅

packages/clickable/index.ts
Fabs: alright.

PR #69, opened 2018-04-19
chore(deps): update dependency tslint-config-prettier to v1.12.0 - autoclosed

mesosphere-web:
This Pull Request updates dependency [tslint-config-prettier](https://github.com/alexjoverm/tslint-config-prettier) from `v1.10.0` to `v1.12.0`

<details>
<summary>Release Notes</summary>

### [`v1.12.0`](https://github.com/alexjoverm/tslint-config-prettier/releases/v1.12.0)
[Compare Source](https://github.com/alexjoverm/tslint-config-prettier/compare/v1.11.0...v1.12.0)
##### Features

* **checker:** Do not output anything if there are no conflicting rules ([#&#8203;134](`https://github.com/alexjoverm/tslint-config-prettier/issues/134`)) ([ae07ea7](https://github.com/alexjoverm/tslint-config-prettier/commit/ae07ea7))

---

### [`v1.11.0`](https://github.com/alexjoverm/tslint-config-prettier/releases/v1.11.0)
[Compare Source](https://github.com/alexjoverm/tslint-config-prettier/compare/v1.10.0...v1.11.0)
##### Features

* **checker:** add ability to check multiple files at once ([#&#8203;127](`https://github.com/alexjoverm/tslint-config-prettier/issues/127`)) ([e3e8012](https://github.com/alexjoverm/tslint-config-prettier/commit/e3e8012))
##### Reverts

* "Merge pull request [#&#8203;119](`https://github.com/alexjoverm/tslint-config-prettier/issues/119`) from alexjoverm/greenkeeper/semantic-release-15.1.2" ([#&#8203;136](`https://github.com/alexjoverm/tslint-config-prettier/issues/136`)) ([963e690](https://github.com/alexjoverm/tslint-config-prettier/commit/963e690))

---

</details>

---

This PR has been generated by [Renovate Bot](https://renovatebot.com).

PR #93, opened 2018-05-07
Add npm ls to the initialization of jenkins

Poltergeist:
<details><summary><strong>ci: add check for peer dependecies</strong></summary><hr/>

`npm ls` will throw an error if any peer dependecies are unmet. This commit adds npm ls to the
initialization step of the `Jenkinsfile` so this step will fail if not all dependecies are met.
</details>

---
Commit:
2a968d2c3e65b75cf12772b93b8077e7c08ce922

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,8 @@ pipeline {
       steps {
         ansiColor('xterm') {
           retry(2) {
-            sh '''npm install'''
+            // The npm ls ensures that the peer dependecies are met.
+            sh '''npm install && npm ls'''
           }
         }
       }


weblancaster:
@Poltergeist seems like a legit failure.

Poltergeist:
Yes, unfortunately `npm ls` also fails if an optional dependency is not installed. I was trying to get this done because of #69 as the peer dependency was unmet.

PR #134, opened 2018-05-28
Refactor(BadgeButton): use clickable within badgebutton

Poltergeist:
<details><summary><strong>refactor(badge): use clickable component for badge button</strong></summary><hr/>

This introduces clickable to BadgeButton.
</details>

---
Commit:
df83487068b2ff4bed9fcc72ff425d49c90d79c5


packages/badge/components/badgeButton.tsx
```diff
@@ -1,5 +1,6 @@
 import { css } from "emotion";
 import * as React from "react";
+import Clickabel from "../../clickable/components/clickable";
```
juliangieseke: 🔡 Clickable*

orlandohohmeier:
Code looks good, but I'd rename the component to clickable badge as it's not a button.

packages/badge/components/badgeButton.tsx
```diff
@@ -10,7 +11,7 @@ export interface IBadgeButtonProps {
     | "warning"
     | "danger"
     | "outline";
-  onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+  onClick: (event?: React.SyntheticEvent<HTMLElement>) => void;
```
juliangieseke: so `onClick` is required, but its parameter isnt? 🤔 shouldnt both be required? I mean, its `Clickable` … something should happen, after you clicked…

packages/badge/components/badgeButton.tsx
```diff
@@ -29,17 +30,20 @@ export class BadgeButton extends React.PureComponent<IBadgeButtonProps, {}> {
...
     };
 
-    return <span {...props}>{children}</span>;
+    const className = css`
+      outline: none;
+      cursor: pointer;
+      ${badgeButton(appearance)};
+    `;
+    return (
+      <Clickabel {...props}>
```
juliangieseke: 🔡 Clickable

weblancaster:
@orlandohohmeier I was going to rename but thank you for merging.

packages/badge/components/badgeButton.tsx
orlandohohmeier: Something would happen... it just means that you might not get an event object.

Poltergeist: It just means that the parameter is optional, so if you pass in a function without a parameter defined. Let's pretend we do have a kill service button. You click it and the application will know by context what it should do, there is no need to look at the event, right? But in some other cases, it might be necessary to look at the event. That is why it is kept optional.

PR #150, opened 2018-06-14
DCOS-38065, DCOS-37579: feat: implement table component

weblancaster:
Wrap react-virtualized in a new component called Table
exposing only the props necessary to display a barebones
table.

Closes DCOS-38064
<details><summary><strong>chore(deps): add packages</strong></summary><hr/>

This adds the following packages:
 - react virtualized
   This is the library we are using for our virtual list it is providing
   us with the multiGrid component which we use for our table.
 - memoize one
   This is a function which will cache return values based on the
   parameters, so it will only execute if the parameters change
 - enzyme-to-json
   This will generate a json for snapshot testing from enzyme objects.
   Unfortunately testing the table is cumbersome with the react
   renderer.
 - jest-emotion
   This will provide the possibility to render the css created by
   emotion instead of just the css class names. By that we can also test
   the resulting css. Which is important for the table.
These packages are needed for the
Table component.
</details>

---
Commit:
3ed0db9c92ed1bb06dc3822ec9981f27d40fad08
<details><summary><strong>feat(table): add column component</strong></summary><hr/>

This introduces the Column component to the library. The Column will take 3 properties:
  - cellRenderer
    The cellRenderer will expect a function which returns a JSX.Element which will be
    used to render the Datum cell.
  - width
    The width will expect a function to calculate the width it should return a number.
    As a parameter the total count of columns, total width, remaining width and current
    column index are provided in a single object.
  - label
    The label expects a JSX Element or a string which will be used to render the header.

closes DCOS-38064
</details>

---
Commit:
39fadd5b83fd4341916966ca47b6ddab3fd76683
<details><summary><strong>feat(table): this introduces the table wrapping component</strong></summary><hr/>

This introduces a Table component which expects data as props and Columns as children. Data is a
Array of any. It will use the props from the Column Component to render a proper table. This also
introduces the horizontal scrolling for the table if the content is bigger then the total width.
Although the first column always stays fixed.

close DCOS-38065, DCOS-37579
</details>

---
Commit:
1a80d30301071704beda1e23169d4e9b2e4e3a35
<details><summary><strong>test: add initial tests</strong></summary><hr/>

</details>

---
Commit:
62b60b75a4397235d8d7bfc5107cee08b4b91b87


Poltergeist:
This needs more
- [ ] Tests
- [ ] Documentation

nLight:
Flicts!

packages/table/tests/__snapshots__/Table.test.tsx.snap
```diff
@@ -0,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Poltergeist: This snapshot does not look right.

packages/table/components/Column.tsx
```diff
@@ -0,0 +1,18 @@
...
+    args: {
+      width: number;
+      totalColumns: number;
+      remainingWidth: number;
+      currentIndex: number;
+    }
+  ) => number;
+}
+
+export class Column extends React.PureComponent<IColumnProps, {}> {}
```
DanielMSchmidt: This seems pretty unfinished, is a column really an empty component?

package.json
```diff
@@ -35,13 +35,14 @@
   ],
   "typings": "dist/components/index",
   "dependencies": {
-    "emotion": "9.2.4",
+    "memoize-one": "3.1.1",
     "react": "16.4.1",
+    "react-virtualized": "9.19.1",
+    "emotion": "9.2.4",
```
DanielMSchmidt: I think this is incorrectly ordered and will re-order once you run npm install.

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+  data: any;
+  children:
+    | Array<React.ReactElement<IColumnProps>>
+    | React.ReactElement<IColumnProps>;
+}
+
+export class Table extends React.PureComponent<ITableProps, {}> {
+  public multiGridRef: { recomputeGridSize?: any } = {};
+
+  public getData = memoizeOne(
```
DanielMSchmidt: 1) Why is this method public, do we really want to call it from the outside?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    const column = React.Children.toArray(this.props.children)[
+      columnIndex
+    ] as React.ReactElement<IColumnProps>;
+
+    if (!column || column.type !== Column) {
+      return null;
+    }
+
+    let className = css({
```
DanielMSchmidt: Can we extract this into a function? I personally like the style a lot better with early returning and not overwriting and having that capsuled into a function as it makes it easier for me to quickly skim over the code here and still understand what it does

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const {
+      fixedColumnCount,
+      enableFixedColumnScroll,
+      enableFixedRowScroll
+    } = this.props;
+    const mergedData = this.getData(this.props.data);
+    const columnCount = React.Children.count(this.props.children);
+
+    return (
+      <AutoSizer onResize={this.updateGrid}>
```
DanielMSchmidt: Update grid made me for a second a bit worried that we update values in the grid or something the like. Can we use a different name, maybe updateGridSize?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+              ref={this.setRef}
+              fixedColumnCount={fixedColumnCount}
+              fixedRowCount={1}
+              cellRenderer={this.getCell}
+              columnWidth={getColumnSize}
+              columnCount={columnCount}
+              enableFixedColumnScroll={enableFixedColumnScroll}
+              enableFixedRowScroll={enableFixedRowScroll}
+              height={height}
+              rowHeight={35}
```
DanielMSchmidt: Can we move these magic numbers into constants? Top of the file would be ok.

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+              rowCount={mergedData.length}
+              width={width}
+              hideTopRightGridScrollbar={true}
```
DanielMSchmidt: I think you could also leave out the `={true}` part for these boolean properties if I remember JSX correctly

packages/table/tests/__snapshots__/Table.test.tsx.snap
DanielMSchmidt: Agreed, I really wonder why though...

packages/table/components/Column.tsx
Poltergeist: Yes it is. It is not intended to render.

package.json
Poltergeist: I'll check that.

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+    | Array<React.ReactElement<IColumnProps>>
+    | React.ReactElement<IColumnProps>;
+}
+
+export class Table extends React.PureComponent<ITableProps, {}> {
+  public multiGridRef: { recomputeGridSize?: any } = {};
+
+  public getData = memoizeOne(
+    (data: any[]): any[] => {
+      return [{}, ...data];
```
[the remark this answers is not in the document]
Poltergeist: Well, we could concat, but we need to have an item for the header. The memoizeOne will make sure that this is only executed when data has changed.

packages/table/components/Table.tsx
Poltergeist: Yes, that makes complete sense sorry to make you feel that way.

packages/table/components/Table.tsx
Poltergeist: Yes sounds good. They not Magic :-)

packages/table/tests/__snapshots__/Table.test.tsx.snap
Poltergeist: My guess only cells in the viewport get rendered.

packages/table/components/Table.tsx
Poltergeist: could you please provide me with an example?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+  children:
+    | Array<React.ReactElement<IColumnProps>>
+    | React.ReactElement<IColumnProps>;
+}
+
+export class Table extends React.PureComponent<ITableProps, {}> {
+  public multiGridRef: { recomputeGridSize?: any } = {};
+
+  public getData = memoizeOne(
+    (data: any[]): any[] => {
```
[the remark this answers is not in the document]
Poltergeist: We do not have a specific type. We render a table all we know is that data is an array how from the array output is generated is defined by the column component.

packages/table/components/Table.tsx
Poltergeist: Did that.

packages/table/components/Table.tsx
Poltergeist: done

packages/table/components/Table.tsx
Poltergeist: that really defines which defaults are set right? I wanted to be over verbal here not if they change something our stuff breaks.

packages/table/tests/__snapshots__/Table.test.tsx.snap
DanielMSchmidt: yeah, that could be right 🤔

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,169 @@
...
+import { css } from "emotion";
+import { AutoSizer, MultiGrid } from "react-virtualized";
+import { IColumnProps, Column } from "./Column";
+import memoizeOne from "memoize-one";
+import { coreColors as Colors } from "../../shared/styles/color";
+
+export interface ITableProps {
+  fixedColumnCount: number;
+  enableFixedColumnScroll: boolean;
+  enableFixedRowScroll: boolean;
```
Poltergeist: I don't think that we need to expose `fixedColumnCount`, `enableFixedColumnScroll`,  `enableFixedRowScroll` in most cases it will be `1`, `true` and `true`. At least I can't recall any example where we would need these props. Maybe let's not expose them now. to prevent later problems.

packages/table/components/Table.tsx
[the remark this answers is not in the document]
Poltergeist: Not necessarily let's say we do have an array of numbers thus `numbers[]` it would return an array of numbers and one object.

packages/table/components/Table.tsx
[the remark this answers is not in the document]
Poltergeist: The cell renderer access the data based on row indexes. Sure we could use some sort of offsets but that would be even more hacky. And error prone.

packages/table/tests/__snapshots__/Table.test.tsx.snap
Poltergeist: I will try to figure out a way to solve that issue.

packages/table/components/Table.tsx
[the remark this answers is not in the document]
DanielMSchmidt: That's why I proposed `(data: T[]): ({} | T)[]`

packages/table/components/Table.tsx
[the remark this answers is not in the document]
DanielMSchmidt: okay, seems fine to me

packages/table/README.md
```diff
@@ -0,0 +1,3 @@
+# Table Name
```
DanielMSchmidt: The documentation is still missing, isn't it?

packages/table/components/Table.tsx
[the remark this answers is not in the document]
Poltergeist: Okay, that sounds good will add that.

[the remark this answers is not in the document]
DanielMSchmidt: For this we could write `Table<T>` to make it generic

packages/table/README.md
Poltergeist: It is I am writing something tomorrow. My muse already left the office. I think this is pretty important since there some questions about the Column and how everything snaps together.

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+    | Array<React.ReactElement<IColumnProps>>
+    | React.ReactElement<IColumnProps>;
+}
+
+const rowHeight = 35;
+export class Table<T> extends React.PureComponent<ITableProps, {}> {
+  public multiGridRef: { recomputeGridSize?: any } = {};
+
+  public getData = memoizeOne(
+    (data: T[]): Array<{}> | T[] => {
```
DanielMSchmidt: 😍

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  public render() {
+    const mergedData = this.getData(this.props.data);
+
+    const columnCount = React.Children.count(this.props.children);
+
+    return (
+      <AutoSizer
+        defaultWidth={1024}
```
nLight: Shall we extract those defaults as props/constants?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+        defaultHeight={768}
+        onResize={this.updateGridSize}
+      >
+        {({ width, height }) => {
```
nLight: We may wanna avoid arrow functions in JSX as they affect performance. Why not extracting a component?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+              hideTopRightGridScrollbar={true}
+              hideBottomLeftGridScrollbar={true}
+            />
+          );
+        }}
+      </AutoSizer>
+    );
+  }
+
+  private getHeaderCell({
```
nLight: Shall we be consistent with the way the `getCell` is written? Back in a day we've decided not to destruct arguments of a method but rather do `const { columnIndex, rowIndex } = args;` instead

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+    if (!column || column.type !== Column) {
+      return null;
+    }
+
+    return rowIndex === 0
+      ? this.getHeaderCell({ ...args, column, data })
+      : this.getContentCell({ ...args, column, data });
+  }
+
+  private getContentCell({
```
nLight: same, please consider not destructing arguments

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+    data,
+    rowIndex
+  }: {
+    style: any;
+    key: any;
+    column: React.ReactElement<IColumnProps>;
+    data: any[];
+    rowIndex: any;
+  }) {
+    let className = css({
```
nLight: We don't change it, do we? Shall it be `const`?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+      borderBottom: `1px solid ${Colors().greyLight}`,
+      boxSizing: "border-box",
+      ...style
+    });
+    return (
+      <div {...{ className, key }}>
```
nLight: Any reason not going with `<div className={className} key={key}>`?

packages/table/README.md
```diff
@@ -0,0 +1,19 @@
+# Table
+
+## Introduction
+
+The Table component gets used to render Tables with the help of a virtual list. It exposes two properties one of them is the `data` which awaits an Array.
+
+The other one is `children` which has to be either an Array of Columns or one Column.
+
+# Columns
```
nLight: There's no `Columns` (plural) component is there?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+    );
+  }
+
+  private getHeaderCell({
+    key,
+    style,
+    column
+  }: {
+    style: any;
+    key: any;
```
nLight: Let's please avoid `any` esp for a new components we write from the scratch. It nullifies the whole point of using TS 🙏

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+    column: React.ReactElement<IColumnProps>;
+  }) {
+    const className = css({
+      boxSizing: "border-box",
+      borderBottom: `1px solid ${Colors().black}`,
+      ...style
+    });
+    return <div {...{ className, key }}>{column.props.label}</div>;
```
nLight: I'd be willing to discuss this pattern via email with the team

packages/table/README.md
Poltergeist: No you are right.

packages/table/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Introduction
+
+Columns is exposing 3 Properties and does not expect to have children. The Columns component is more of an abstract which ensures that the right properties are available and available since the table is working with these.
+
+The `label` is providing the content for the header.
```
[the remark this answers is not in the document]
Poltergeist: Good argument. I picked label since it is more generic. It becomes the header with the combination of the table.

packages/table/components/Column.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface IColumnProps {
+  /**
+   * label is providing the contents for the header cell for the column.
+   */
+  label: string | JSX.Element;
+  /**
+   * cellRenderer is the function which is creating the cell contents for this column.
+   */
+  cellRenderer: (data: any) => JSX.Element;
```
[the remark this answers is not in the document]
Poltergeist: `JSX.Element` is looser it is including HTML elements afaik.

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+import memoizeOne from "memoize-one";
+import { coreColors as Colors } from "../../shared/styles/color";
+
+export interface ITableProps {
+  /**
+   * data is expecting an array of any values. These values should contain all the data represented in the table.
+   */
+  data: any[];
+  /**
+   * children can only be of type Column
```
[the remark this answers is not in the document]
Poltergeist: Do you think that everyone would be able to comprehend from the types that only column components are valid children for the table?

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+   */
+  children:
+    | Array<React.ReactElement<IColumnProps>>
+    | React.ReactElement<IColumnProps>;
+}
+
+const rowHeight = 35;
```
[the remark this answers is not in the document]
Poltergeist: ✅ Will change.

packages/table/components/Table.tsx
Poltergeist: Yes sounds good to me unfortunately we need those for the tests to work.

packages/table/components/Table.tsx
Poltergeist: Let me try that. You are right. But the performance reason does not fit here. Since the lambda function is not passed as properties but as children. This is also how it is solved in the docs: https://github.com/bvaughn/react-virtualized/blob/master/docs/AutoSizer.md

packages/table/components/Table.tsx
Poltergeist: Will do that.

packages/table/components/Table.tsx
Poltergeist: agree

packages/table/components/Table.tsx
Poltergeist: ✅

packages/table/components/Table.tsx
Poltergeist: Convenience, if that is a good reason.

packages/table/components/Table.tsx
Poltergeist: I'll do. Except if it is absolutely necessary, for example, data as this can basically be an array of any value. 

But in this case, I do agree not necessary.

packages/table/components/Table.tsx
Poltergeist: I put it on the sync for tomorrow.

Poltergeist:
Thanks, @nLight for your insights I will fix the comments as soon as possible. Also, I will facilitate the discussions around some patterns.

packages/table/components/Column.tsx
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * width is the function which will get called to determine which width a column has.
+   */
+  width: (
+    args: {
```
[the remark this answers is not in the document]
Poltergeist: ✅

packages/table/README.md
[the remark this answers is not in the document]
Poltergeist: Not a fan because it becomes the `header` by the table component. But I agree it can be confusing going to rename.

packages/table/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The `cellRenderer` is expecting a function which will get called for each cell and the return value should be a React element.
+
+The last property is `width` this expects also a function which should return a numeric value of a width in pixels. This is getting used to calculate the width of a column.
```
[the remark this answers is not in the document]
Poltergeist: ✅

packages/table/components/Column.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+
+export interface IColumnProps {
+  /**
+   * label is providing the contents for the header cell for the column.
+   */
+  label: string | JSX.Element;
```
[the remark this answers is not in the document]
Poltergeist: ✅

packages/table/components/Column.tsx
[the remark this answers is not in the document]
Poltergeist: ✅

packages/table/components/Table.tsx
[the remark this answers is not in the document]
Poltergeist: ✅

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,172 @@
+import * as React from "react";
+import { css } from "emotion";
+import { AutoSizer, MultiGrid } from "react-virtualized";
+import { IColumnProps, Column } from "./Column";
+import memoizeOne from "memoize-one";
+import { coreColors as Colors } from "../../shared/styles/color";
+
+export interface ITableProps {
+  /**
+   * data is expecting an array of any values. These values should contain all the data represented in the table.
```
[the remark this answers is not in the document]
Poltergeist: ✅

Poltergeist:
@nLight and @DanielMSchmidt I worked on the comments from @nLight maybe.

packages/table/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The Table component gets used to render Tables with the help of a virtual list. It exposes two properties one of them is the `data` which awaits an Array.
+
+The other one is `children` which has to be either an Array of Columns or one Column.
+
+# Column
+
+## Introduction
+
+Column component exposes 3 Properties and does not expect to have children. The Columns component ensures that the right properties are available.
```
nLight: One more plural 👼

packages/table/components/Table.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const columnCount = React.Children.count(this.props.children);
+    const columnSizes = this.getColumnSizes(
+      React.Children.toArray(this.props.children) as Array<
+        React.ReactElement<IColumnProps>
+      >,
+      width
+    );
+
+    const getColumnSize = ({ index }) => columnSizes[index];
```
nLight: We may wanna replace `=>` with regular function
```js
function getColumnSize({ index }) { 
  const columnCount = React.Children.count(this.props.children);
  const columnSizes = this.getColumnSizes(
     React.Children.toArray(this.props.children) as Array<
       React.ReactElement<IColumnProps>
     >,
     width
  );
  return columnSizes[index]
}
```

extract the whole logic it needs for `columnSizes` and bind it in the constructor 
so that we spare some performance. Otherwise `=>` will be bound on every function call and a new function will be created that will lead to re-render of `MultiGrid`

Just replacing the line `const getColumnSize = ({ index }) => columnSizes[index];`
with `function getColumnSize ({ index }) { return columnSizes[index]; }` could work.

Poltergeist: I am not sure about the rerender part. Since the column width is getting cached. However, there is nothing speaking against pulling it out anyway.

nLight:
Merge at will

Poltergeist:
:ship:

PR #162, opened 2018-06-29
DCOS-38496, DCOS-38639, DCOS-38640: Table helper functions for calculating width

Poltergeist:
<details><summary><strong>feat(utilities): add min helper function</strong></summary><hr/>

This will provide a curried function to calculate the min value of two parameters. This should be
replaced by ramda as soon as their types have been fixed.

Close DCOS-38496
</details>

---
Commit:
77beb493cad778db3e6df1f1a584f9069025cea2
<details><summary><strong>feat(utilities): this adds a max helper function</strong></summary><hr/>

This provides a max helper function for the users. The function is curried and should be replaced by
ramda as soon as their types are fixed.

Close DCOS-38639
</details>

---
Commit:
ae22a28a0fd218de47c3bb55c2087e428e5afcdd
<details><summary><strong>feat(utilities): add percentage helper</strong></summary><hr/>

This provides the user wiht a curried helper function. The first parameter is the percentage of 100
which we want to calculate the second is the total value so for example if we would like to calulate
30 percent of 1024 we would write `percentage(30, 1024)` or `percentage(30)(1024)`

Close DCOS-38640
</details>

---
Commit:
aae812b56abd9994444bbe3645f026c1c1c36337
<details><summary><strong>feat(utilities): add compose function</strong></summary><hr/>

This provides the user with a compose function which will pass the result from each function to the
next one the functions are executed from right to left
</details>

---
Commit:
743a90b69e489a1e8b44fad4cc69c04c90ac8f71

### DIFF
--- a/packages/utilities/components/compose.ts
+++ b/packages/utilities/components/compose.ts
@@ -1,0 +1,43 @@
+export function compose<A, B, C>(b: (v: B) => C, a: (v: A) => B): (a: A) => C;
+export function compose<A, B, C, D>(
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => D;
+export function compose<A, B, C, D, E>(
+  d: (v: D) => E,
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => E;
+export function compose<A, B, C, D, E, F>(
+  e: (v: E) => F,
+  d: (v: D) => E,
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => F;
+export function compose<A, B, C, D, E, F, G>(
+  f: (v: F) => G,
+  e: (v: E) => F,
+  d: (v: D) => E,
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => G;
+export function compose<A, B, C, D, E, F, G, H>(
+  g: (v: G) => H,
+  f: (v: F) => G,
+  e: (v: E) => F,
+  d: (v: D) => E,
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => H;
+
+export function compose(...functions) {
+  const fn = functions.reverse();
+  return value => fn.reduce((memo, fn) => fn(memo), value);
+}
+
+export default compose;
--- a/packages/utilities/components/max.ts
+++ b/packages/utilities/components/max.ts
@@ -1,0 +1,17 @@
+function calculate(a, b) {
+  return b > a ? b : a;
+}
+
+function max(a?: number | Date | string, b?: number | Date | string) {
+  if (arguments.length === 0) {
+    return max;
+  }
+  if (arguments.length === 1) {
+    return b => {
+      return calculate(a, b);
+    };
+  }
+  return calculate(a, b);
+}
+
+export default max;
--- a/packages/utilities/components/min.ts
+++ b/packages/utilities/components/min.ts
@@ -1,0 +1,17 @@
+function calculate(a, b) {
+  return b < a ? b : a;
+}
+
+function min(a?: number | Date | string, b?: number | Date | string) {
+  if (arguments.length === 0) {
+    return min;
+  }
+  if (arguments.length === 1) {
+    return b => {
+      return calculate(a, b);
+    };
+  }
+  return calculate(a, b);
+}
+
+export default min;
--- a/packages/utilities/components/percentage.ts
+++ b/packages/utilities/components/percentage.ts
@@ -1,0 +1,17 @@
+function calculate(a, b) {
+  return (b / 100) * a;
+}
+
+function percentage(a?: number, b?: number) {
+  if (arguments.length === 0) {
+    return percentage;
+  }
+  if (arguments.length === 1) {
+    return b => {
+      return calculate(a, b);
+    };
+  }
+  return calculate(a, b);
+}
+
+export default percentage;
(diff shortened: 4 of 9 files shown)


packages/utilities/components/compose.ts
```diff
@@ -0,0 +1,5 @@
+function compose(...f) {
```
orlandohohmeier: 📐 Please type the input parameters.

packages/utilities/tests/min.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  it("is returning the min value", () => {
+    expect(min(1)(10)).toBe(1);
+  });
+
+  it("is returning the lower value value", () => {
+    expect(min(12)(10)).toBe(10);
+  });
+
+  // it("lets compose", () => {
```
orlandohohmeier: ☝️ Please delete these lines.

packages/utilities/components/compose.ts
orlandohohmeier: P.S. You should use overloading to properly type this functions for set number of input functions.

```ts
export function compose<A, B, C>(b: (v: B) => C, a: (v: A) => B): (a: A) => C
export function compose<A, B, C, D>(c: (v: C) => D, b: (v: B) => C, a: (v: A) => B): (a: A) => D
export function compose<A, B, C, D, E>(d: (v: D) => E, c: (v: C) => D, b: (v: B) => C, a: (v: A) => B): (a: A) => E
// ... 

export function compose(...f) {
//...
}
```

packages/utilities/components/max.ts
```diff
@@ -0,0 +1,16 @@
+const compare = (a, b) => {
```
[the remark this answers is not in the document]
Poltergeist: That is not true, the original `Math.max` is converting values to `numbers` and returning the bigger number. So if you provide dates you will get back a timestamp, if you add strings you will get back a NaN. But this function will return the bigger item regardless of type. I could name it max but then again the exported function is called max what should the name be here?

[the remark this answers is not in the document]
orlandohohmeier: I understand that It supports more cases then `Math.max`, but it's still not only comparing but _calculating_ or computing the return value. How about `calculate`?

packages/utilities/components/percentage.ts
```diff
@@ -0,0 +1,16 @@
+const calculate = (a, b) => {
```
nLight: wonder why this is an arrow function but `percentage` is a regular function

nLight: It looks like it's more like a pattern of the helper functions in the PR. Any specific reason?

packages/utilities/components/min.ts
```diff
@@ -0,0 +1,16 @@
+const compare = (a, b) => {
+  return b < a ? b : a;
+};
+function min(a?: number | Date | string, b?: number | Date | string) {
```
nLight: Looks like we can compare string with Date, or Number with a string. Is this intended?

Poltergeist: Yes, well we are able to compare them however in a reasonable manner it only makes sense to compare Date's and numbers if you think that a number could be a timestamp. With the strings, it only makes sense to compare them with strings.

packages/utilities/components/compose.ts
Poltergeist: done

Poltergeist:
We tried Ramda, unfortunately, the types are not up to date and they are failing with the latest version. So at some time in the future, we might want to replace these functions with ramda anyways.

nLight:
The branch is outdated tho

packages/utilities/components/compose.ts
```diff
@@ -0,0 +1,43 @@
...
+  g: (v: G) => H,
+  f: (v: F) => G,
+  e: (v: E) => F,
+  d: (v: D) => E,
+  c: (v: C) => D,
+  b: (v: B) => C,
+  a: (v: A) => B
+): (a: A) => H;
+
+export function compose(...functions) {
```
DanielMSchmidt: Why do you have a named and a default export of the very same thing?

packages/utilities/components/max.ts
```diff
@@ -0,0 +1,17 @@
...
+  }
+  if (arguments.length === 1) {
+    return b => {
+      return calculate(a, b);
+    };
+  }
+  return calculate(a, b);
+}
+
+export default max;
```
DanielMSchmidt: 🎨 Can we inline the default exports in general?

packages/utilities/components/max.ts
```diff
@@ -0,0 +1,17 @@
+function calculate(a, b) {
+  return b > a ? b : a;
+}
+
+function max(a?: number | Date | string, b?: number | Date | string) {
+  if (arguments.length === 0) {
```
DanielMSchmidt: Is there a special reason why you allow this?

Poltergeist: Basically, because that is how it is solved in ramda.

Poltergeist:
@nLight @DanielMSchmidt I picked components since these functions are the components of the utility package. However, I can move the scripts from components to a `src` directory?

packages/utilities/components/compose.ts
Poltergeist: So that I can use both notations for import 
`import compose from "compose.ts"` and
`import {compose} from "compose.ts"`

Poltergeist: Here is a quick example: https://codesandbox.io/s/231wklxny0?expanddevtools=1

nLight:
ping @orlandohohmeier

packages/utilities/components/min.ts
```diff
@@ -0,0 +1,17 @@
+function compare(a, b) {
```
orlandohohmeier: 🎨 please rename this one to `calculate`  as well

packages/utilities/components/max.ts
juliangieseke: I think he wanted to go with double export again 🤔

juliangieseke:
Ping @orlandohohmeier

PR #178, opened 2018-07-13
DCOS-39532: feat: table style

weblancaster:
Add table initial styles.

Closes DCOS-39532

### DIFF
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -1,3 +1,7 @@
+import { injectGlobalCss } from "./shared/styles/global";
+
 export { Badge } from "./badge";
 export { BadgeButton } from "./badge";
 export { Column, Table, Cell, TextCell, HeaderCell } from "./table";
+
+injectGlobalCss();
--- a/packages/shared/styles/global.ts
+++ b/packages/shared/styles/global.ts
@@ -1,0 +1,9 @@
+import { injectGlobal } from "emotion";
+
+export const injectGlobalCss = () => {
+  return injectGlobal`
+    .ReactVirtualized__Grid {
+      outline: none;
+    }
+  `;
+};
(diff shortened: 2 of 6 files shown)


DanielMSchmidt:
@weblancaster Could you take a look at tslint, it seems to fail. Also having a Jira issue linked would help the review 🙈

packages/shared/styles/global.ts
```diff
@@ -0,0 +1,9 @@
+import { injectGlobal } from "emotion";
+
+export const injectGlobalCss = () => {
+  return injectGlobal`
+    .ReactVirtualized__Grid {
```
Poltergeist: I don’t feel like this is a good idea from an a11y standpoint. But I am happy to be proven otherwise

packages/index.ts
```diff
@@ -1,3 +1,7 @@
+import { injectGlobalCss } from "./shared/styles/global";
+
 export { Badge } from "./badge";
 export { BadgeButton } from "./badge";
 export { Column, Table, Cell, TextCell, HeaderCell } from "./table";
+
+injectGlobalCss();
```
Poltergeist: We should discuss this. As this will only work if we import from `@dcos/ui-kit` and not with submodule imports.

packages/table/stories/Table.stories.tsx
```diff
@@ -41,7 +41,7 @@ const cityCellRenderer = ({ city }: { city?: string }) => (
 );
 const roleCellRenderer = ({ role }: { role?: string }) => (
   <Cell>
-    <em>{role}</em>
+    <span>{role}</span>
```
Poltergeist: Why change this? The example was intended to show that the user can define how the contents of a cell looks like.

packages/table/components/Table.tsx
```diff
@@ -61,6 +63,7 @@ export class Table<T> extends React.PureComponent<ITableProps, {}> {
     return (
       <AutoSizer
         __dataThatTriggersARerenderWhenChanged={this.props.data} // passed to notify react-virtualized about updates
+        className={tableCss}
```
Poltergeist: Couldn’t we attach the table css rather to the multi grid instead of the autosizer?

PR #190, opened 2018-08-06
DCOS-39830: Create NumberCell in the UI Kit

GeorgiSTodorov:
This adds a NumberCell component that is right aligned.

This is my first time modifying the ui kit, so please tell me if I missed something.

Related Jira: https://jira.mesosphere.com/browse/DCOS-39830


GeorgiSTodorov:
I've added the NumberCell component to the table story.

screenshot from 2018-08-07 15-33-43

PR #193, opened 2018-08-06
DCOS-39674: feat: add column shadow

weblancaster:
Add shadow to table fixed column

Closes DCOS-39674

**How to test**
1. `npm start`
2. Go to `http://localhost:6006/`
3. Go to table component and note the newly added shadow to the fixed column

screen shot 2018-08-08 at 5 59 32 pm


weblancaster:
@leemunroe can you take a quick look in the shadow and let me know what you think, please? thanks.

lilysamimi:
@weblancaster looks great! A few comments: 

- The table shadow should only appear when the user scrolls horizontally.
- Is there any way to marry the shadow in the header with the shadow below the header to make it look like it's one shadow and not two separate shadows? Seems like there is some very tiny pixel padding (both top and bottom) between the header lines that we could get rid of?

weblancaster:
@lilysamimi 

1. I will take a look and see if there is an API for us to accomplish that.
2. No way (perhaps a VERY hacky way) at the moment because they are separated grids.

AmrAbdelrazik:
Re 2 @lilysamimi It's not ideal but I think it's ok.

GeorgiSTodorov:
I think it looks good, but I think it's worth considering if it looks better with a small margin for the second column.

### with 5px margin:
screenshot from 2018-08-07 11-55-13

### currently:
screenshot from 2018-08-07 11-56-14

packages/shared/styles/global.ts
```diff
@@ -5,5 +5,10 @@ export const injectGlobalCss = () => {
     .ReactVirtualized__Grid {
       outline: none;
     }
+
+    .ReactVirtualized__Grid:nth-child(1),
+    .BottomLeftGrid_ScrollWrapper {
+      box-shadow: 5px 0 5px -2px #d6d4d4;
```
DanielMSchmidt: Can we move this color to a central place for colors? I think having an organisation like this should be one of the core elements of the ui-kit. I personally would also like to have the default shadow width and spread in variables, but I don't have hard feelings around this.

DanielMSchmidt: @mperrotti what would you say?

mperrotti: Definitely agree that we should have a collection of "design tokens" to pull from. For now, the design tokens can live here in DC/OS UI Kit, but in the future it would be cleaner to put them in their own package.

For now, we export a new mapping from `packages/shared/styles/shadow.ts`, that includes something like:
`shadowRight: 5px 0 5px -2px #d6d4d4`

mperrotti: Agreed on adding the color as well. `shadow` is a little confusing as a variable name for a color though. We could benefit from namespacing colors

weblancaster:
Thank you @DanielMSchmidt and @mperrotti I completely forgot we have the re-usable colors `https://github.com/dcos-labs/ui-kit/blob/master/packages/shared/styles/color.ts`.

I have addressed the comments.

@DanielMSchmidt now about the default width and spread I would, there is no way for us to really know because depending on the direction we might want less or more of those.

weblancaster:
@Poltergeist Looks pretty good, I will update the code following your example. Thank you.

Now, what do you mean by not using global styles? they are injected into the page using the specific selector. What is your suggestion?

weblancaster:
@leemunroe @Poltergeist I have addressed based on your comments but I still need to add the scroll only at scroll which hasn't been to straight forward.

Poltergeist:
I have a CSS solution for this as well, unfortunately, chrome isn't playing nice.  I am currently investigating solutions.

Poltergeist:
I have added two commits to this PR @weblancaster, unfortunately, it only really works in Firefox. Maybe you could have a look because I am out of ideas. In Chrome it somehow works but not flawless.

weblancaster:
@Poltergeist I worked on a solution yesterday which it should work but for some reason, the grid doesn't update whenever the react lifecycle is triggered. Here is my solution https://github.com/dcos-labs/ui-kit/commit/0a8911aaeed6841c2abafd52a77e7293236ca5bc

packages/table/components/Table.tsx
```diff
@@ -73,7 +82,22 @@ export class Table<T> extends React.PureComponent<ITableProps, {}> {
     );
   }
 
+  private onScrollbarPresenceChange({ horizontal }) {
```
mperrotti: We're following up to add the shadow on scroll with [DCOS-40413](https://jira.mesosphere.com/browse/DCOS-40413). Is this code still necessary for this PR?

packages/table/style.ts
```diff
@@ -22,3 +22,9 @@ export const tableCss = css`
   font-family: ${fontFamilySansSerif};
   font-weight: normal;
 `;
+
+export const rightGrid = css`
+  background: linear-gradient(to right, rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0));
```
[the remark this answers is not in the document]
weblancaster: Agree, I will adjust the code to use the color. Thank you

packages/table/components/Table.tsx
weblancaster: In this case, we are only displaying the shadow when `onScrollbarPresenceChange` is `true` that way when we have few columns without scroll we don't show the shadow as per request from the UX team.

DCOS-40413 is still needed in the future.

packages/table/style.ts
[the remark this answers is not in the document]
Poltergeist: We could also use the hextorgba helper we introduced and just use black from our colors. But I would suggest to do that later

packages/table/style.ts
```diff
@@ -24,7 +24,11 @@ export const tableCss = css`
 `;
 
 export const rightGrid = css`
-  background: linear-gradient(to right, rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0));
+  background: linear-gradient(
+    to right,
+    ${hexToRgbA(black, 0.2)},
```
mperrotti: This works if we don't want to decide on an "official" shadow color yet, but I meant that we should have something like
```
shadow: hexToRgbA(black, 0.2)
```
in `styles/color.ts` and import that here

PR #194, opened 2018-08-07
DCOS-39675: Hide scrollbars in Table grids

mperrotti:
Jira ticket: https://jira.mesosphere.com/browse/DCOS-39675

### Screenshots:
Before (in Chrome):
scroll--before

After (in Chrome):
hidescrollbars_after

### Description
I chose to do this by hiding the scrollbar with vendor-specific CSS on both grids
    **Pros:**
    - User can scroll when cursor is in either of the right grids
    - Vendor-specific CSS will be ignored by browsers who don't support these rules, so no surprise bugs
    **Cons:**
    - Only works for IE/Edge and webkit browsers

I also looked at using `overflow: hidden` on top-left grid, browser-specific for bottom-left. This would still let the top-left grid scroll, but it would never show a scrollbar, so only the bottom scrollbar would appear in browsers that don't give you a way to style/hide it's scrollbars.
    **Pros:**
    - Firefox and other browsers that are incompatible with the vendor-specific styles will only show the bottom-right grid (table content) scrollbar
    **Cons:**
    - User cannot scroll if cursor is in the top-right grid (header cells)

---

If the design team isn't happy with scrollbars showing in Firefox and other browsers that are incompatible with the vendor-specific styles, I know a somewhat hack-y way we can handle this that will work in all browsers. I'd advocate against this because it's not a great idea to fight browser-native scrollbar behavior.

I didn't add any tests because the existing snapshot tests should be sufficient for these changes

### DIFF
--- a/packages/table/style.ts
+++ b/packages/table/style.ts
@@ -32,3 +32,11 @@ export const rightGrid = css`
   background-repeat: no-repeat;
   background-size: 8px 100%;
 `;
+
+export const hideScrollbarCss = css`
+  -ms-overflow-style: -ms-autohiding-scrollbar;
+
+  ::-webkit-scrollbar {
+    display: none;
+  }
+`;
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -1,8 +1,14 @@
 import * as React from "react";
-import { css } from "emotion";
+import { css, cx } from "emotion";
 import { AutoSizer, MultiGrid, GridCellProps } from "react-virtualized";
 
-import { headerCss, cellCss, tableCss, rightGrid } from "../style";
+import {
+  headerCss,
+  cellCss,
+  tableCss,
+  rightGrid,
+  hideScrollbarCss
+} from "../style";
 
 import { IColumnProps, Column } from "./Column";
 import memoizeOne from "memoize-one";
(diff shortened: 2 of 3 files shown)


packages/table/style.ts
```diff
@@ -22,3 +22,11 @@ export const tableCss = css`
   font-family: ${fontFamilySansSerif};
   font-weight: normal;
 `;
+
+export const hideScrollbarCss = css`
+  -ms-overflow-style: -ms-autohiding-scrollbar;
```
mperrotti: The Jira ticket recommends `-ms-overflow-style: none;`, but when I've tried this in the past with horizontal overflows, IE/Edge users cannot scroll horizontally using a swipes on a trackpad. Can confirm once I have a way to test in IE/Edge

mperrotti:
@GeorgiSTodorov - should we add it to our `.gitignore`?

DanielMSchmidt:
Can confirm looks good. 

@Poltergeist Are we using the angular commit conventions in this repo or are they unenforced?

Poltergeist:
@DanielMSchmidt You are right Mike could you please run `npm run setup:git-hooks` this should add a post commit hook to your local repository which will lint your commit massages. (https://github.com/dcos-labs/ui-kit/blob/master/CONTRIBUTING.md#commit-types)

I'll fix them for you in this PR. Also I will update the snapshot tests.

Poltergeist:
Hi @mperrotti, I just pushed changes to this PR:
 - I rebased to master
 - I updated snapshot tests
 - I changed the commit message to be valid to conventional commits
 - I squashed the commits into the first commit
 - I removed the changes from the `package-lock.json` file.

mperrotti:
Thanks for the feedback, everyone. I set up my git hooks, and I'll be sure to follow our commit message convention going forward.
Is there anything else I need to do before this can be approved and merged?

mperrotti:
@juliangieseke - that was an out-of-date screenshot. Just updated it in the description

Poltergeist:
I rebased to master and resolved conflicts.

PR #198, opened 2018-08-09
ci(releases): replace standard-version with semantic-release

juliangieseke:
This replaces `standard-version` with `semantic-release` (which we already use on `dcos-ui` repo). after this PR got merged (and everything works 🤞 ), we do automatic releases everytime something releaseworthy got merged to master 💃

PR #203, opened 2018-08-10
DCOS-40413: Only show Table column shadow on scroll

mperrotti:
shadowscroll_after


Jenkinsfile
```diff
@@ -80,6 +80,17 @@ pipeline {
       }
```
weblancaster: I believe this should not be in your PR.

.releaserc
```diff
@@ -0,0 +1,3 @@
+{
```
weblancaster: I believe this should not be in your PR

README.md
```diff
@@ -72,9 +72,9 @@ A recommended reading is [Better Specs](http://www.betterspecs.org/), we put
 [real effort](https://github.com/dcos/dcos-ui/pull/2524) in making sure we
```
weblancaster: I believe this should not be in your PR

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dcos/ui-kit",
-  "version": "1.3.0",
+  "version": "0.0.0-dev+semantically-released",
```
weblancaster: why was this changed?

.releaserc
nLight: I agree with Michael, let's please keep those separate

package.json
Poltergeist: yes

nLight:
flicts

package.json
Poltergeist: https://github.com/dcos-labs/ui-kit/pull/198

juliangieseke:
@mperrotti this branch is messed up, you own commit is in here twice 🤔

nLight:
flicts

Poltergeist:
I will rebase this Pr and solve the conflict to move this forward.

Poltergeist:
I was not able to rebase this and fix the conflicts, unfortunately.

Poltergeist:
I thought we only want to show the shadow while scrolling, although I like how the functionality is working.

mperrotti:
@Poltergeist - I've always seen it done this way, where the shadow remains visible as long as the user is scrolled over. @leemunroe or @lilysamimi - what do you think?

lilysamimi:
@mperrotti It's perfect! 🙇🏻‍♀️

@leemunroe check out the latest shadow improvements! https://cl.ly/0i3x1i0S0X1o

One minor thing that's not related to this PR and I'm not sure if it's a bug but I want to call it out just in case ... the "Zip Code" header is not aligned with the rest of the column, there is a column with no header but the table cells are called "empty", and the "City" column is being treated as a primary column it seems as the table cells are in bold.

PR #209, opened 2018-08-20
DCOS-40355: add hover style for Table rows

mperrotti:
Before:
rowhover_before

After (the hover shadow is _very_ hard to see in after the uploaded gif has been compressed, but it's there):
rowhover_after

### DIFF
--- a/packages/table/style.ts
+++ b/packages/table/style.ts
@@ -2,7 +2,7 @@ import { css } from "emotion";
 import { coreColors, hexToRgbA } from "../shared/styles/color";
 import { coreFonts } from "../shared/styles/typography";
 
-const { black, greyLight } = coreColors();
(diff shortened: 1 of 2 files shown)


packages/table/style.ts
```diff
@@ -40,3 +40,9 @@ export const hideScrollbarCss = css`
     display: none;
   }
 `;
+
+export const rowHoverCss = css`
+  background-color: ${greyLightLighten5};
+  mix-blend-mode: multiply;
+  will-change: left;
```
mperrotti: Adding `will-change` is to fix a rendering bug in Chrome and Safari. The value has to be something that does not create stacking context, otherwise it will not work in Safari.

If you're curious for a demo of the bug this fixes, we can do a screenshare

PR #210, opened 2018-08-20
DCOS-40668: Fix create component template in UI kit

weblancaster:
Fix component template paths

Closes DCOS-40668

## How to test

1. `npm run create:component blabla`
2. Go to `packages/index.ts` and export your module as so `export { Blabla } from "./blabla";`
3. `npm start` then visit `http://localhost:6006/`
4. Make sure you can see the component `blabla` on the left menu


scripts/templates/component/stories/Component.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { storiesOf } from "@storybook/react";
 import { withReadme } from "storybook-readme";
-import ${Component} from "../${Component}";
+import { ${Component} }from "../../index";
```
nLight: Missing space before `from` and extra `../`
should be `import { ${Component} } from "../index";`

scripts/templates/component/tests/Component.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import ${Component} from ".";
+import { ${Component} }from "../";
```
nLight: missing space before `from`

nLight:
Would appreciate some testing guides. What I did:

1. `npm run create:component Dima`
2. Check if everything looks ok

GeorgiSTodorov:
I also created a new component to see if it works and checked the paths that were changed.

mperrotti:
I ran the same thing Dima did (found [here](https://github.com/dcos-labs/ui-kit/blob/master/CONTRIBUTING.md#creating-a-new-component)) and saw the same issues he did. Should be good to go after fixing those typos.

scripts/templates/component/stories/Component.stories.tsx
weblancaster: Ah good catch 😞

scripts/templates/component/tests/Component.test.tsx
weblancaster: Ah good catch 😞

scripts/templates/component/stories/Component.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { storiesOf } from "@storybook/react";
 import { withReadme } from "storybook-readme";
-import ${Component} from "../${Component}";
+import { ${Component} } from "../../index";
```
nLight: Should be `"../index"`

weblancaster: Why should be `../index`? the reason we are using `../../index` is to target the root path as the projects will.

nLight: Ha! I get it now, so after scaffolding one should put the component into the main index

weblancaster: Correct. I added that to the description too.

Thanks.

PR #211, opened 2018-08-20
DCOS-40043: toggle component

weblancaster:
Add toggle component used to switch between two
elements (String or JSX).

Closes DCOS-40043

### DIFF
--- a/packages/clickable/style.ts
+++ b/packages/clickable/style.ts
@@ -1,10 +1,7 @@
 import { css } from "emotion";
-import { coreColors } from "../shared/styles/color";
 
-const { white } = coreColors();
 export const outline = css`
   &:focus {
-    box-shadow: 0 0 0 1px ${white}, 0 0 0 3px currentColor;
     outline: 0;
   }
 `;
--- a/packages/badge/tests/__snapshots__/badgeButton.test.tsx.snap
+++ b/packages/badge/tests/__snapshots__/badgeButton.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`BadgeButton accept jsx as children 1`] = `
 <span
-  className="css-1ydbtd2"
+  className="css-1mfjuh7"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -16,7 +16,7 @@ exports[`BadgeButton accept jsx as children 1`] = `
 
 exports[`BadgeButton danger 1`] = `
 <span
-  className="css-16xjcxo"
+  className="css-5yz412"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -28,7 +28,7 @@ exports[`BadgeButton danger 1`] = `
 
 exports[`BadgeButton default 1`] = `
 <span
-  className="css-1ydbtd2"
+  className="css-1mfjuh7"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -40,7 +40,7 @@ exports[`BadgeButton default 1`] = `
 
 exports[`BadgeButton outline 1`] = `
 <span
-  className="css-w3zj9q"
+  className="css-1nvs8l1"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -52,7 +52,7 @@ exports[`BadgeButton outline 1`] = `
 
 exports[`BadgeButton primary 1`] = `
 <span
-  className="css-n0m8w5"
+  className="css-1e4v5oe"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -64,7 +64,7 @@ exports[`BadgeButton primary 1`] = `
 
 exports[`BadgeButton success 1`] = `
 <span
-  className="css-qzf5te"
+  className="css-stidvd"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
@@ -76,7 +76,7 @@ exports[`BadgeButton success 1`] = `
 
 exports[`BadgeButton warning 1`] = `
 <span
-  className="css-dus73p"
+  className="css-ddfcfh"
   onClick={[Function]}
   onKeyPress={[Function]}
   role="button"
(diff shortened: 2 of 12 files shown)


packages/toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { Toggle } from "../";
+
+expect.addSnapshotSerializer(createSerializer(emotion));
+
+const primary = () => <div>primary</div>;
+const secondary = () => <div>secondary</div>;
+
+describe("Toggle", () => {
+  it("renders correctly", () => {
```
DanielMSchmidt: Can you write a title that explains what you are testing here? Correctly is pretty vague.

packages/toggle/README.md
```diff
@@ -0,0 +1,3 @@
+# Toggle Name
+
+Toggle documentation
```
DanielMSchmidt: 🔥The documentation is missing.

packages/toggle/components/toggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+  public render() {
+    const { primaryContent, secondaryContent } = this.props;
+    let content = primaryContent;
+
+    if (!this.state.isPrimary) {
+      content = secondaryContent;
+    }
+
+    return (
+      <div className={style} onClick={this.handleClick}>
```
mperrotti: This isn't keyboard-navigable as-is. We should either do this as a `input[type="checkbox"]` in disguise, or add the correct ARIA roles and a `tabindex` to it, as well as keyboard handlers to toggle the state when Enter or Spacebar are pressed.

packages/toggle/style.ts
```diff
@@ -0,0 +1,6 @@
+import { css } from "emotion";
+
+export const style = css`
+  display: inline-block;
```
mperrotti: I think we should let component consumers decide what the `display` is. Let's _just_ make this component responsible for toggling itself on and off.

packages/toggle/README.md
weblancaster: @DanielMSchmidt We don't have documentation for this component, I will come up with something very simple but I would not block this PR because of the documentation.

packages/toggle/tests/Toggle.test.tsx
weblancaster: Yes

packages/toggle/components/toggle.tsx
weblancaster: Agree, I will add aria.

packages/toggle/style.ts
weblancaster: Sounds good.

packages/clickable/style.ts
```diff
@@ -1,10 +1,7 @@
 import { css } from "emotion";
-import { coreColors } from "../shared/styles/color";
 
-const { white } = coreColors();
 export const outline = css`
   &:focus {
-    box-shadow: 0 0 0 1px ${white}, 0 0 0 3px currentColor;
```
weblancaster: Not sure why we had this bold outline.

packages/toggle/stories/Toggle.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const readme = require("../README.md");
+
+const primary = () => <div>primary component</div>;
+const secondary = () => <div>secondary component</div>;
+
+storiesOf("Toggle", module)
+  .addDecorator(withReadme([readme]))
+  .addWithInfo("string", () => (
+    <Toggle primaryContent="Hello" secondaryContent="Bye" />
```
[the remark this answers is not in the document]
weblancaster: @mperrotti I have addressed the ARIA/role issue which discards the second code suggestion.

I was advocating for us to abstract the logic and component entirely, meaning the developer only need to pass the data and UI Kit takes care of displaying it.

I'm open to leaving a bit more flexible using the render prop interface but I would rather see us not giving that condition flexibility in this component.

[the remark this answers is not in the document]
weblancaster: What would you suggest for the props to replace `primaryContent` and `secondaryContent`

DanielMSchmidt:
@weblancaster Could you combine the commits? I think introducing a feature and fixing it afterwards is a bit strange 😇

packages/toggle/stories/Toggle.stories.tsx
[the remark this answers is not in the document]
mperrotti: If `primary-` and `secondary-` are a familiar naming convention for toggle content/values, we should keep what you have instead of making people learn something new

mperrotti: Pass a `tabIndex` to the story - it defaults to -1

Not part of this PR, but something to consider for the Clickable component:
`:focus { outline: 0; }` hides any visual indication that the element has focus. Where might I file a ticket to improve the keyboard navigation for that component?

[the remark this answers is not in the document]
weblancaster: What do you think about `on` `off`?

weblancaster: I will remove in this PR because it's pretty small overhead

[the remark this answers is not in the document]
mperrotti: What about `contentOn` and `contentOff`? I think adding `content-`makes it clearer that those props accept content to be rendered. `on` and `off` seem a little mysterious to me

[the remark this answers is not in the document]
weblancaster: Sounds good.

mperrotti:
Sorry, one more small thing:
I think I prefer the name `ContentToggle` or `ToggleContent`. It better describes what this component does, and we're less likely to run into a conflict if we ever create a component that we want to name `Toggle`

packages/clickable/style.ts
```diff
@@ -1,10 +1,7 @@
...
     outline: 0;
```
weblancaster: @mperrotti I decided to leave it here and decide later what to  do about the `outline` because of design wise we do not want to show but a11y I would yes

mperrotti: Sounds good. Have you spoken with the design team about how they want to handle this styling? Or should I?

PR #216, opened 2018-08-29
DCOS-41098: ToggleContent no toggle when text highlighted

mperrotti:
The click event was completing when a user highlighted text and released the click

Before:
before

After:
after

### DIFF
--- a/packages/toggleContent/components/toggleContent.tsx
+++ b/packages/toggleContent/components/toggleContent.tsx
@@ -30,9 +30,11 @@ export class ToggleContent extends React.PureComponent<
   }
 
   public handleClick(): void {
-    this.setState({
-      isOn: !this.state.isOn
-    });
+    if (!window.getSelection().toString()) {
+      this.setState({
+        isOn: !this.state.isOn
+      });
+    }
   }
 
   public render() {
@@ -45,7 +47,7 @@ export class ToggleContent extends React.PureComponent<
 
     return (
       <Clickable tabIndex={tabIndex} action={this.handleClick}>
-        <div className={style}>{content}</div>
+        <span className={style}>{content}</span>
       </Clickable>
     );
   }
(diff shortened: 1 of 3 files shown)


packages/toggleContent/tests/toggleContent.test.tsx
```diff
@@ -12,6 +12,14 @@ const primary = () => <div>primary</div>;
 const secondary = () => <div>secondary</div>;
 
 describe("ToggleContent", () => {
+  beforeEach(() => {
+    const mock = jest.fn();
+    mock.mockReturnValue({
+      toString: jest.fn()
+    });
+    window.getSelection = mock;
```
mperrotti: Jest doesn't support `window.getSelection`, so I had to mock it

weblancaster: Makes sense.

weblancaster:
Thanks for the fix.

packages/toggleContent/components/toggleContent.tsx
```diff
@@ -45,7 +47,7 @@ export class ToggleContent extends React.PureComponent<
 
     return (
       <Clickable tabIndex={tabIndex} action={this.handleClick}>
-        <div className={style}>{content}</div>
+        <span className={style}>{content}</span>
```
weblancaster: This will cause the tests in DCOS UI to fail but we can update the snapshot with ease.
https://github.com/dcos/dcos-ui/pull/3218/files#diff-b344e4851c4356add413f1168ead98a5R75

nLight:
When click selects the string it is hardly going away

aug-29-2018 18-00-23

packages/toggleContent/components/toggleContent.tsx
mperrotti: Updated the snaps already

mperrotti:
@nLight - Double-clicking a text node to highlight it is default browser behavior. I think it might feel strange to users if we override that behavior.

What behavior did you have in mind?

mperrotti:
Holding merge until we hear back from @lilysamimi about whether we should give toggle-able text some kind of visual treatment (per the comments in [this PR](https://github.com/dcos/dcos-ui/pull/3218))

weblancaster:
@mperrotti I would suggest opening a new jira for the visual change and merge this one.

PR #220, opened 2018-09-04
DCOS-34197: pass calculated width to cellRenderer

mperrotti:
This PR makes the cell width available as an argument in the function passed to `cellRenderer` so we can make informed decisions about what we render in the cells based on their available width


packages/table/components/Column.tsx
```diff
@@ -5,6 +5,8 @@ export interface IWidthArgs {
   remainingWidth: number;
   currentIndex: number;
 }
+
+export type Width = string | number | undefined;
```
DanielMSchmidt: I think this abstraction has a problem. As a cell renderer I would need to know in which table I am used and if this table uses a string width like `10%` or a numeric pixel value. I think we should only work with numbers here, anything else provides an extra layer of complexity with no real benefit on the user-side

PR #221, opened 2018-09-05
DCOS-41057: customize text alignment for cells

mperrotti:
Now you can pass `textAlign` to any `<Cell>`-based component to align the text


packages/table/style.ts
```diff
@@ -15,12 +16,30 @@ export const headerCss = css`
...
+export const innerCellCss = css`
+  height: 100%;
+  box-sizing: border-box;
+  padding: ${cellPadding}px;
+`;
+
+export const cellAlignmentCss = (textAlign: string) => css`
+  text-align: ${textAlign};
+  margin-right: ${textAlign === "right"
+    ? `${pointerSize + pointerSpace + cellPadding}px`
```
mperrotti: This is here because @lilysamimi specifically requested we treat the down-pointing triangle as a hanging indent. So when right-aligned, the text of the cell content should always be aligned with the text, not the sort icon

PR #231, opened 2018-10-09
DCOS-42621: add base InfoBox component

mperrotti:
Added an InfoBox component to help us render consistently render banner-like messages to users. This component is needed for the "upgrade DC/OS UI as a package" work we're doing.

## Screenshots:
**InfoBoxBanner (used for the message to upgrade DC/OS' UI):**
screen shot 2018-10-08 at 5 14 50 pm

**InfoBoxInline:**
screen shot 2018-10-08 at 5 14 41 pm

Apologies to the reviewers: This PR got a little bigger than what I'd normally submit for a 1-point ticket


packages/clickable/tests/clickable.test.tsx
```diff
@@ -73,12 +73,10 @@ describe("Clickable", () => {
           <span>default tabIndex</span>
         </Clickable>
       );
-      expect(
```
mperrotti: This was breaking tests, even in master. Maybe related to some recent package upgrades?

packages/table/tests/SortableHeaderCell.test.tsx
```diff
@@ -6,7 +6,6 @@ import { createSerializer } from "jest-emotion";
 import { shallow } from "enzyme";
 import toJson from "enzyme-to-json";
 import HeaderCell from "../components/HeaderCell";
-import Clickable from "../../clickable/components/clickable";
```
mperrotti: This was breaking tests, even in master. Maybe related to some recent package upgrades?

mperrotti:
@lilysamimi or @leemunroe, I could use some guidance from the design team

**What I need to get this merged:**
Where can I find the "x" icon SVG?
Is the "x" always intended to be vertically centered? Consider how this might look when the message/actions wrap to ≥2 lines

**What we'll need to consider in the near future:**
I don't think we should put the InfoBox component to be in charge of enforcing the button styles because:
1. We don't have our canonical button styles anywhere in ui-kit yet
2. Some buttons may change in appearance after they've been clicked (e.g.: active state to loading state)
3. It will be a pain to work around that styling if we ever decide we want the Primary action to be more called out than the Secondary action
4. It will be easier to re-color the text when it's on a colored background rendered by the success, danger, and warning variants
I'm in favor of specifying this style in design system documentation, and calling out undesirable styles in PR. However, if the design team feels strongly about the InfoBox component enforcing styles, that's something we'd be able to handle pretty easily once we have a `<Button>` component

We need a way to decide which banner message to show when there are more than 1 possibilities

packages/clickable/tests/clickable.test.tsx
```diff
@@ -73,12 +73,10 @@ describe("Clickable", () => {
...
-        wrapper
-          .find("span")
-          .props()
-          .tabIndex.toString()
```
mperrotti: This was breaking tests, even in master. Maybe related to some recent package upgrades?

packages/infobox/README.md
```diff
@@ -0,0 +1,5 @@
+# InfoBox Name
+
+InfoBox documentation
+
+<!-- TODO: collaborate with Design team to add documentation from Confluence doc -->
```
mperrotti: @lilysamimi - should I just create a follow-up ticket for documentation? We can pair like we did for the Table docs :)

GeorgiSTodorov:
Looks good.
Small question:
screenshot from 2018-10-09 11-01-42

Is there a specific reason why the secondary action is first?

mperrotti:
@GeorgiSTodorov - this is how the design was spec'd. My guess is because it's just slightly easier to click when it's at the end

nLight:
Conflicts! Sorry I've merged the renovate update :)

lilysamimi:
> I don't think we should put the InfoBox component to be in charge of enforcing the button styles because...

@mperrotti I think this is fine as the styling of the buttons on the infobox may change overtime. 

Are you going to include the 4 visual styles of info box in the ui-kit. i.e. warning, danger, success and info

mperrotti:
@lilysamimi - of course :) 

The variants will be done as a separate (much simpler) ticket

lilysamimi:
Looking at the text in the infobox, I'm wondering size you're using for the font?

lilysamimi:
> I think the x should be fixed to the top right.

@leemunroe I'm wondering why top right is better than middle right?

mperrotti:
@lilysamimi - I'm not explicitly setting a font size, it will be whatever the default font size of whatever app is importing it. In this case, it's `16px`

lilysamimi:
> @lilysamimi - I'm not explicitly setting a font size, it will be whatever the default font size of whatever app is importing it. In this case, it's `16px`

So we could potentially have inconsistent sizes in font across our info boxes? 16px seems pretty large. Could we just have the font size be at 14px?

PR #236, opened 2018-10-18
DCOS-43348: create a design token distribution

mperrotti:
I'm using [style-dictionary](https://github.com/amzn/style-dictionary/) to create a distribution of design tokens (design attributes that can be expressed in a key-value pair).

As a proof-of-concept, I've updated `color.stories.ts` to use design tokens.


package.json
```diff
@@ -15,15 +15,17 @@
...
     "commitlint": "commitlint -e",
     "create:component": "./scripts/create-component",
     "deploy:storybook": "storybook-to-ghpages",
     "predist": "npm run clean",
-    "dist": "tsc -p tsconfig.dist.json",
+    "dist": "tsc -p tsconfig.dist.json && npm run build:design-tokens",
     "lint": "tslint -c tslint.json -p tsconfig.json",
     "setup:git-hooks": "./scripts/pre-install",
-    "start": "NODE_ENV=development npm run storybook --root ./ ",
+    "start": "NODE_ENV=development npm run build:design-tokens && npm run storybook --root ./ ",
```
mperrotti: Yes, this is weird, but we can just import design tokens as a package once we use Lerna to manage multiple packages in this repo.
Then we can remove all the design token changes in here

mperrotti:
@GeorgiSTodorov - the tokens should be usable in dcos-ui. The style utilities will require us to set up Emotion in dcos-ui

GeorgiSTodorov:
@mperrotti Please look into file conflict so that we can review.

packages/design-tokens/scripts/formats.js
```diff
@@ -0,0 +1,38 @@
+const SD_lessVariables = require("../../../node_modules/style-dictionary/lib/common/formats")[
```
nLight:
```suggestion
const SD_lessVariables = require("style-dictionary/lib/common/formats")[
```
also worked for me

packages/design-tokens/scripts/formats.js
```diff
@@ -0,0 +1,38 @@
...
+
+const dateHeader = () => `
+  /**
+   * Do not edit directly
+   * Generated on ${new Date()}
+   */
+`;
+
+const lessVariables = {
+  name: "less/variablesCustom",
```
nLight: Is there a specific name we could give it? What is custom about our way of formatting it?

packages/design-tokens/scripts/formats.js
nLight: I would still call it `lessVariables`

packages/design-tokens/scripts/formats.js
mperrotti: I meant to remove this. I was doing a trying a slightly different format than style-dictionary gives you by default

PR #242, opened 2018-10-24
fix: export notification components

mperrotti:
Forgot to export InfoBox and Toast components in `packages/index.ts` 😳


nLight:
😅

PR #244, opened 2018-10-26
fix: do not export global element reset styles

mperrotti:
Injecting global element styles was overriding some CNVS styles that we want in dcos-ui.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -4,6 +4,7 @@ import { configure, setAddon, addDecorator } from "@storybook/react";
 import { withKnobs } from "@storybook/addon-knobs/react";
 import { setOptions } from "@storybook/addon-options";
 import infoAddon, { setDefaults } from "@storybook/addon-info";
+require("../packages/shared/styles/global").injectStorybookResetCss();
 
 addDecorator(withKnobs);
 setAddon(infoAddon);
--- a/packages/shared/styles/global.ts
+++ b/packages/shared/styles/global.ts
@@ -4,27 +4,16 @@ const { fontFamilySansSerif } = coreFonts();
 
 export const injectGlobalCss = () => {
   return injectGlobal`
-    body {
-      font-family: ${fontFamilySansSerif};
-    }
-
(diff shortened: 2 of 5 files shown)


packages/shared/styles/global.ts
```diff
@@ -4,27 +4,16 @@ const { fontFamilySansSerif } = coreFonts();
...
     .ReactVirtualized__Grid {
       outline: none;
     }
   `;
 };
+
+export const injectStorybookResetCss = () => {
+  return injectGlobal`
+    body {
+        font-family: ${fontFamilySansSerif};
```
mperrotti: Components look really weird when they're in the browser's default serif font. I'm comfortable with some minimal styling on anything rendered in Storybook instead of exporting this style with the components.

GeorgiSTodorov:
Is there a way to test this?

mperrotti:
@GeorgiSTodorov, here are steps to test:
1. `npm run dist` on this branch of ui-kit
2. Drag the generated `dist/` directory to `dcos-ui/node_modules/@dcos/ui-kit/dist/`
3. Run dcos-ui and go to a package detail page (I recommend Datadog)

The paragraphs in the pre-install notes should have bottom margins, and the headers on the page should have a bold or medium weight (use the [screenshot comparison in Lee's comment](https://github.com/dcos/dcos-ui/pull/3381#issuecomment-433548700) as a guide)

.storybook/config.js
```diff
@@ -4,9 +4,11 @@ import { configure, setAddon, addDecorator } from "@storybook/react";
 import { withKnobs } from "@storybook/addon-knobs/react";
 import { setOptions } from "@storybook/addon-options";
 import infoAddon, { setDefaults } from "@storybook/addon-info";
+import { injectStorybookResetCss } from "../packages/shared/styles/global";
```
Poltergeist: Maybe 
```js
require("../packages/shared/styles/global").injectStorybookResetCss(); 
```
is better becasue we are going to call this one anyways only once.

PR #249, opened 2018-11-14
fix: add dev build config for style dictionary

mperrotti:
Before these changes, design tokens were built in the `postinstall` step. However, the script to build the tokens was not in the npm package, so install would always fail. To get around this, a second distribution of design tokens is built in the `predist` step.

Once we're using Lerna to manage `design-tokens` as it's own package, we shouldn't need a second distribution of design tokens because they will be imported as a package.


Poltergeist:
Looks good to me.

PR #253, opened 2018-11-20
Infobox style update: hover state and vertical center for dismiss

mperrotti:
Before:
nohover-nocenter

After:
withhover-withcenter


lsamimi:
Looks great @mperrotti! 🎉

PR #254, opened 2018-11-26
DCOS-45443: Allow AppChrome-related components to be themed

GeorgiSTodorov:
## General
Allow AppChrome-related components to be themed

Closes https://jira.mesosphere.com/browse/DCOS-45443

## Testing
Edit: See `HeaderBar Customizable w/ Knobs`, `Sidebar Customizable w/ Knobs` and `SidebarItem Customizable w/ Knobs` in AppChrome storybook.

## Notes and questions
Since I am not an ui-kit guru, please review critically.

I wanted to create an enum for this - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-41fb508168f47f98220876ec033f23d1R14 , but I am not sure if I should and where to put it.

I had to remove this - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-3768d7bd77fce72475adfd42705edcfcL89 for custom icon width to work. I don't think I broke anything, but a second look would be valuable.

What would be the preferred way to set the icon size here - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R163 , in the `PlaceholderIcon` component like I did it, or somewhere else?

I assumed that the padding should only have our predetermined values here (for example "xs", "xl", etc.) - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R372 , is this correct?

## Screenshots
Edit:
### Knobs
screenshot from 2018-11-30 14-08-30

screenshot from 2018-11-30 14-08-54

screenshot from 2018-11-30 14-09-05

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "typings": "dist/packages/index",
   "dependencies": {
     "emotion": "9.2.12",
-    "emotion-theming": "^9.2.9",
+    "emotion-theming": "9.2.9",
     "memoize-one": "4.0.2",
     "react": "16.5.2",
     "react-dom": "16.5.2",
@@ -58,7 +58,7 @@
     "@storybook/addon-a11y": "3.4.11",
     "@storybook/addon-actions": "3.4.11",
     "@storybook/addon-info": "3.4.11",
-    "@storybook/addon-knobs": "^3.4.11",
+    "@storybook/addon-knobs": "3.4.11",
     "@storybook/addon-links": "3.4.11",
     "@storybook/addon-options": "3.4.11",
     "@storybook/addons": "3.4.11",
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "typings": "dist/packages/index",
   "dependencies": {
     "emotion": "9.2.12",
+    "emotion-theming": "^9.2.9",
     "memoize-one": "4.0.2",
     "react": "16.5.2",
     "react-dom": "16.5.2",
@@ -57,7 +58,7 @@
     "@storybook/addon-a11y": "3.4.11",
     "@storybook/addon-actions": "3.4.11",
     "@storybook/addon-info": "3.4.11",
-    "@storybook/addon-knobs": "3.4.11",
+    "@storybook/addon-knobs": "^3.4.11",
     "@storybook/addon-links": "3.4.11",
     "@storybook/addon-options": "3.4.11",
     "@storybook/addons": "3.4.11",
--- a/packages/appChrome/components/HeaderBar.tsx
+++ b/packages/appChrome/components/HeaderBar.tsx
@@ -4,6 +4,7 @@ import styled from "react-emotion";
 import { darkMode, flex } from "../../shared/styles/styleUtils";
 import { spaceSizes } from "../../../packages/shared/styles/styleUtils/modifiers/modifierUtils";
 import { purpleDarken4 } from "../../design-tokens/build/js/designTokens";
+import { isHexDark } from "../../shared/styles/color";
 
 export interface HeaderProps {
   children: React.ReactElement<HTMLElement> | string;
@@ -15,20 +16,26 @@ class Header extends React.PureComponent<HeaderProps, {}> {
     /* tslint:disable:no-string-literal */
     const HeaderBar = styled("div")`
       background-color: ${props =>
-        props.theme.backgroundColor || purpleDarken4};
+        props.theme.headerBackgroundColor || purpleDarken4};
+      ${props =>
+        !props.theme.headerBackgroundColor ||
+        (props.theme.headerBackgroundColor &&
+          isHexDark(props.theme.headerBackgroundColor))
+          ? darkMode
+          : null};
       padding-left: ${props =>
-        spaceSizes[props.theme.paddingHor] || spaceSizes["l"]};
+        spaceSizes[props.theme.headerPaddingHor] || spaceSizes["l"]};
       padding-right: ${props =>
-        spaceSizes[props.theme.paddingHor] || spaceSizes["l"]};
+        spaceSizes[props.theme.headerPaddingHor] || spaceSizes["l"]};
       padding-bottom: ${props =>
-        spaceSizes[props.theme.paddingVert] || spaceSizes["xs"]};
+        spaceSizes[props.theme.headerPaddingVert] || spaceSizes["xs"]};
       padding-top: ${props =>
-        spaceSizes[props.theme.paddingVert] || spaceSizes["xs"]};
+        spaceSizes[props.theme.headerPaddingVert] || spaceSizes["xs"]};
     `;
     /* tslint:enable:no-string-literal */
 
     return (
-      <HeaderBar className={cx(darkMode, flex({ align: "center" }))}>
+      <HeaderBar className={cx(flex({ align: "center" }))}>
         {children}
       </HeaderBar>
     );
--- a/packages/appChrome/components/HeaderBar.tsx
+++ b/packages/appChrome/components/HeaderBar.tsx
@@ -1,7 +1,9 @@
 import * as React from "react";
 import { cx } from "emotion";
-import { darkMode, flex, padding } from "../../shared/styles/styleUtils";
-import { headerBar, appChromeInsetContent } from "../style";
+import styled from "react-emotion";
+import { darkMode, flex } from "../../shared/styles/styleUtils";
+import { spaceSizes } from "../../../packages/shared/styles/styleUtils/modifiers/modifierUtils";
+import { purpleDarken4 } from "../../design-tokens/build/js/designTokens";
 
 export interface HeaderProps {
   children: React.ReactElement<HTMLElement> | string;
@@ -10,20 +12,25 @@ export interface HeaderProps {
 class Header extends React.PureComponent<HeaderProps, {}> {
   public render() {
     const { children } = this.props;
+    /* tslint:disable:no-string-literal */
+    const HeaderBar = styled("div")`
+      background-color: ${props =>
+        props.theme.backgroundColor || purpleDarken4};
+      padding-left: ${props =>
+        spaceSizes[props.theme.paddingHor] || spaceSizes["l"]};
+      padding-right: ${props =>
+        spaceSizes[props.theme.paddingHor] || spaceSizes["l"]};
+      padding-bottom: ${props =>
+        spaceSizes[props.theme.paddingVert] || spaceSizes["xs"]};
+      padding-top: ${props =>
+        spaceSizes[props.theme.paddingVert] || spaceSizes["xs"]};
+    `;
+    /* tslint:enable:no-string-literal */
 
     return (
-      <div
-        className={cx(
-          headerBar,
-          darkMode,
-          appChromeInsetContent,
-          flex({ align: "center" }),
-          padding("top", "xs"),
-          padding("bottom", "xs")
-        )}
-      >
+      <HeaderBar className={cx(darkMode, flex({ align: "center" }))}>
         {children}
-      </div>
+      </HeaderBar>
     );
   }
 }
--- a/packages/appChrome/components/Sidebar.tsx
+++ b/packages/appChrome/components/Sidebar.tsx
@@ -1,7 +1,10 @@
 import * as React from "react";
-import { cx } from "emotion";
+import { cx, css } from "emotion";
+import styled from "react-emotion";
 import { darkMode } from "../../shared/styles/styleUtils";
 import { sidebar, sidebarAnimator } from "../style";
+import { greyDark } from "../../design-tokens/build/js/designTokens";
+import { atMediaUp } from "../../shared/styles/breakpoints";
 
 export interface SidebarProps {
   children: React.ReactElement<HTMLElement> | string;
@@ -9,6 +12,28 @@ export interface SidebarProps {
   onOpen?: () => void;
   onClose?: () => void;
 }
+
+const sidebarWidths = {
+  default: "240px",
+  large: "280px"
+};
+const layoutBreakpoint = "large";
+
+const sidebarWidth = css`
+  width: ${sidebarWidths.default};
+
+  ${atMediaUp[layoutBreakpoint](css`
+    width: ${sidebarWidths[layoutBreakpoint]};
+  `)};
+`;
+
+const sidebarAnimatorWidth = (isOpen: boolean) => css`
+  width: ${isOpen ? sidebarWidths.default : 0};
+
+  ${atMediaUp[layoutBreakpoint](css`
+    width: ${isOpen ? sidebarWidths[layoutBreakpoint] : 0};
+  `)};
+`;
 
 class Sidebar extends React.PureComponent<SidebarProps, {}> {
   public componentWillReceiveProps(nextProps: SidebarProps) {
@@ -23,12 +48,26 @@ class Sidebar extends React.PureComponent<SidebarProps, {}> {
 
   public render() {
     const { children, isOpen } = this.props;
-    const classNames = cx(sidebar, darkMode);
+    const navClassNames = cx(sidebar, sidebarWidth, darkMode);
+    const divClassNames = cx(sidebarAnimator);
+
+    const Sidebar = styled("div")`
+      background-color: ${props => props.theme.backgroundColor || greyDark};
+      ${props =>
+        props.theme.width
+          ? "width: " + props.theme.width
+          : sidebarAnimatorWidth(isOpen)};
+    `;
+
+    const Nav = styled("nav")`
+      ${props =>
+        props.theme.width ? "width: " + props.theme.width : sidebarWidth};
+    `;
 
     return (
-      <div className={sidebarAnimator(isOpen)}>
-        <nav className={classNames}>{children}</nav>
-      </div>
+      <Sidebar className={divClassNames}>
+        <Nav className={navClassNames}>{children}</Nav>
+      </Sidebar>
     );
   }
 }
--- a/packages/appChrome/components/Sidebar.tsx
+++ b/packages/appChrome/components/Sidebar.tsx
@@ -5,6 +5,7 @@ import { darkMode } from "../../shared/styles/styleUtils";
 import { sidebar, sidebarAnimator } from "../style";
 import { greyDark } from "../../design-tokens/build/js/designTokens";
 import { atMediaUp } from "../../shared/styles/breakpoints";
+import { isHexDark } from "../../shared/styles/color";
 
 export interface SidebarProps {
   children: React.ReactElement<HTMLElement> | string;
@@ -48,20 +49,29 @@ class Sidebar extends React.PureComponent<SidebarProps, {}> {
 
   public render() {
     const { children, isOpen } = this.props;
-    const navClassNames = cx(sidebar, sidebarWidth, darkMode);
+    const navClassNames = cx(sidebar, sidebarWidth);
     const divClassNames = cx(sidebarAnimator);
 
     const Sidebar = styled("div")`
-      background-color: ${props => props.theme.backgroundColor || greyDark};
+      background-color: ${props =>
+        props.theme.sidebarBackgroundColor || greyDark};
       ${props =>
-        props.theme.width
-          ? "width: " + props.theme.width
+        props.theme.sidebarWidth
+          ? "width: " + props.theme.sidebarWidth
           : sidebarAnimatorWidth(isOpen)};
     `;
 
     const Nav = styled("nav")`
       ${props =>
-        props.theme.width ? "width: " + props.theme.width : sidebarWidth};
+        props.theme.sidebarWidth
+          ? "width: " + props.theme.sidebarWidth
+          : sidebarWidth};
+      ${props =>
+        !props.theme.sidebarBackgroundColor ||
+        (props.theme.sidebarBackgroundColor &&
+          isHexDark(props.theme.sidebarBackgroundColor))
+          ? darkMode
+          : null};
     `;
 
     return (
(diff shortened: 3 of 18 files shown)


packages/appChrome/components/HeaderBar.tsx
```diff
@@ -1,15 +1,68 @@
...
+    | undefined;
 }
 
 class Header extends React.PureComponent<HeaderProps, {}> {
   public render() {
-    const { children } = this.props;
+    const {
+      children,
+      backgroundColor,
+      paddingBottom,
```
DanielMSchmidt: Deconstructing might make the defaults easier to visually grasp

```suggestion
      paddingBottom = "xs",
```

packages/appChrome/components/HeaderBar.tsx
```diff
@@ -1,15 +1,68 @@
...
+  darkMode,
+  flex,
+  padding
+} from "../../shared/styles/styleUtils";
 import { headerBar, appChromeInsetContent } from "../style";
 
 export interface HeaderProps {
+  backgroundColor?: string;
   children: React.ReactElement<HTMLElement> | string;
+  paddingBottom?:
```
DanielMSchmidt: Types are like variables for structure, we could easily make this shorter and more readable.

```suggestion
  paddingBottom?: SizingOption
```


Somewhere else (ideally in a place for base types):

```ts
export type SizingOption = 
  | "xxs"
  | "xs"
...
```

packages/appChrome/stories/AppChrome.stories.tsx
```diff
@@ -42,6 +43,21 @@ storiesOf("AppChrome", module)
       <SidebarBareContent />
     </Sidebar>
   ))
+  .addWithInfo("Sidebar bare w/ custom background", () => (
+    <Sidebar backgroundColor={greenDarken1} isOpen={true}>
+      <SidebarBareContent />
+    </Sidebar>
+  ))
+  .addWithInfo("Sidebar bare w/ custom padding", () => (
+    <Sidebar paddingLeft="xl" paddingRight="xl" isOpen={true}>
```
DanielMSchmidt: I would really really love to see the usage of [storybook-knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs#array) here. It would allow us to play with it and get some more insights on how it behaves. It would also allow us to do most of these examples in one bigger, more configurable one, which would be awesome.

If you like we can pair on this 👍

packages/shared/styles/styleUtils/modifiers/backgroundColor.ts
```diff
@@ -0,0 +1,7 @@
+import { css } from "react-emotion";
+
+export const background = color => {
+  return css`
+    background-color: ${color} !important;
```
DanielMSchmidt: Do we really need the !important? It always sends shivers down my spine.

packages/appChrome/components/HeaderBar.tsx
GeorgiSTodorov: I don't know where the place for base types is, which is why I asked.

GeorgiSTodorov: For example, `display` here doesn't have a place for base types - https://github.com/dcos-labs/ui-kit/blob/90280672f59c9af244a8c2e5838294c3bb5969a7/packages/shared/styles/styleUtils/modifiers/display.ts#L3 .

packages/shared/styles/styleUtils/modifiers/backgroundColor.ts
GeorgiSTodorov: I wanted to be consistent with existing modifiers, for example - https://github.com/dcos-labs/ui-kit/blob/90280672f59c9af244a8c2e5838294c3bb5969a7/packages/shared/styles/styleUtils/modifiers/display.ts#L19 .

packages/appChrome/components/HeaderBar.tsx
DanielMSchmidt: There is none yet, for example `packages/types`

packages/appChrome/stories/AppChrome.stories.tsx
DanielMSchmidt: I would say yes, as you are adding a lot of examples which may be cluttered and hard to oversee (hurting usability), but you judge :)

packages/shared/styles/styleUtils/modifiers/backgroundColor.ts
DanielMSchmidt: @Poltergeist @mperrotti Could you give me some background on this?

packages/appChrome/components/HeaderBar.tsx
GeorgiSTodorov: I would prefer to wait for @mperrotti to discuss the best place to have these, since he is the one who does most of the work here.

nLight:
Hey @GeorgiSTodorov, curious what was the outcome of the emotion ThemeProvider investigation?

GeorgiSTodorov:
@nLight I decided that a new dependency isn't worth it, if it's something we can easily do with props. If you disagree, please challenge me.

nLight:
@GeorgiSTodorov I see value in the separation of concerns the `ThemeProvider` provides. So that you don't need changing your app to pass props around. You just wrap it with the provider and it just works.

packages/appChrome/components/HeaderBar.tsx
mperrotti: @DanielMSchmidt - I like the idea of having a place for types like `packages/types`

packages/appChrome/stories/AppChrome.stories.tsx
mperrotti: +1 for knobs usage. Also happy to pair with you on that

packages/shared/styles/styleUtils/modifiers/backgroundColor.ts
mperrotti: I think it's safe to remove the `!important`s on any modifiers.
`display` modifiers were some of the first ones I wrote when I was new to using emotion, so I was using the brute-force approach I'd use in regular CSS to make sure modifier styles can never be beaten.

packages/shared/styles/styleUtils/modifiers/width.ts
```diff
@@ -0,0 +1,7 @@
+import { css } from "react-emotion";
+
+export const width = width => {
```
mperrotti: Not sure if `width` really needs to be a modifier. We don't have a predetermined collection of values to pick from like we do with modifiers like `padding`.

packages/appChrome/components/HeaderBar.tsx
```diff
@@ -1,15 +1,68 @@
...
+    | "xxs"
+    | "xs"
+    | "s"
+    | "m"
+    | "l"
+    | "xl"
+    | "xxl"
+    | "none"
+    | undefined;
+  paddingLeft?:
```
mperrotti: The design team always wants the left and right padding to be the same value, and the top and bottom padding to be the same value.

So we could use the value from a prop like `paddingHoriz="16px"` to set 16px left _and_ right padding styles in the component, and `paddingVert="16px"` to set 16px top _and_ bottom padding styles in the component.

packages/appChrome/components/SidebarItem.tsx
```diff
@@ -2,21 +2,53 @@ import * as React from "react";
...
 import Clickable from "../../clickable/components/clickable";
-import { tintContentPrimary } from "../../shared/styles/styleUtils";
+import { tintContentPrimary, padding } from "../../shared/styles/styleUtils";
 
 export interface SidebarItemProps {
   children?: React.ReactElement<HTMLElement> | string;
   icon?: React.ReactElement<HTMLElement> | string;
   isActive?: boolean;
   onClick: (event?: React.SyntheticEvent<HTMLElement>) => void;
+  paddingBottom?:
```
mperrotti: Left and right padding should also be customizable

mperrotti:
@tnguyen0 should have a look too since Learning Platform designs are where these came from :)

packages/appChrome/components/Sidebar.tsx
```diff
@@ -22,12 +49,31 @@ class Sidebar extends React.PureComponent<SidebarProps, {}> {
...
+      backgroundColor,
+      children,
+      isOpen,
+      paddingLeft,
+      paddingRight,
+      sidebarWidth
+    } = this.props;
+    const navClassNames = cx(
+      sidebar,
+      darkMode,
```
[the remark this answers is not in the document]
mperrotti: I ran into a similar issue while working on the buttons, so I wrote a helper function to check color brightness: https://github.com/dcos-labs/ui-kit/compare/master...8d3c6bf6f5a81b87f3eadcbb1d691580a1f3d995#diff-c7218e3329e428d70b7790b006ce556bR161

mperrotti:
> I wanted to create an enum for this - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-41fb508168f47f98220876ec033f23d1R14 , but I am not sure if I should and where to put it.

I think the `packages/types` directory Daniel suggested is a good spot


> I had to remove this - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-3768d7bd77fce72475adfd42705edcfcL89 for custom icon width to work. I don't think I broke anything, but a second look would be valuable.

I'd keep it in. Setting the width of that container div to the size of the icon protects the alignment of the SidebarItemLabel text with the submenu items.
screen shot 2018-11-26 at 8 01 04 pm


> What would be the preferred way to set the icon size here - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R163 , in the PlaceholderIcon component like I did it, or somewhere else?

This is fine, but we should also pass the size of the icon into the component for styling purposes. Something like `<SidebarItemLabel iconSize={24} icon={<PlaceholderIcon size={24}} />`

We're using the icon size px value to align the submenu items to the SidebarItemLabel's text [using padding](https://github.com/dcos-labs/ui-kit/blob/master/packages/appChrome/style.ts#L106). This isn't an ideal situation, but I don't think there's a way to achieve that alignment without having the icon width :(
screen shot 2018-11-26 at 8 04 28 pm


> I assumed that the padding should only have our predetermined values here (for example "xs", "xl", etc.) - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R372 , is this correct?

Great, that's what I would have done too :)

packages/appChrome/components/SidebarItem.tsx
GeorgiSTodorov: Okay, updated the jira to reflect this.

packages/appChrome/components/HeaderBar.tsx
GeorgiSTodorov: Great suggestion.

packages/appChrome/components/Sidebar.tsx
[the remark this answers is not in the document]
GeorgiSTodorov: Should we also pass darkMode/lightMode?

packages/appChrome/stories/helpers/StorybookSidebarHelpers.tsx
```diff
@@ -9,10 +9,10 @@ import {
 } from "../../index";
 import { padding } from "../../../shared/styles/styleUtils";
 
-export const PlaceholderIcon = () => (
+export const PlaceholderIcon = props => (
```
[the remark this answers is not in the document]
GeorgiSTodorov: Good point.

packages/appChrome/components/HeaderBar.tsx
```diff
@@ -1,15 +1,68 @@
...
+    | "xxs"
+    | "xs"
+    | "s"
+    | "m"
+    | "l"
+    | "xl"
+    | "xxl"
+    | "none"
+    | undefined;
+  paddingTop?:
```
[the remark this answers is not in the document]
GeorgiSTodorov: Header bar height shouldn't be customizable according to the jira.

packages/appChrome/components/Sidebar.tsx
```diff
@@ -1,13 +1,40 @@
...
+} from "../../shared/styles/styleUtils";
 import { sidebar, sidebarAnimator } from "../style";
 
 export interface SidebarProps {
+  backgroundColor?: string;
   children: React.ReactElement<HTMLElement> | string;
   isOpen: boolean;
   onOpen?: () => void;
   onClose?: () => void;
+  paddingLeft?:
```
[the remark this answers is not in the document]
GeorgiSTodorov: Okay, will update the jira so that it remains the single source of truth.

packages/appChrome/components/SidebarItem.tsx
```diff
@@ -2,21 +2,53 @@ import * as React from "react";
 import { cx } from "emotion";
 import { appChromeInsetContent, sidebarNavItem } from "../style";
 import Clickable from "../../clickable/components/clickable";
-import { tintContentPrimary } from "../../shared/styles/styleUtils";
+import { tintContentPrimary, padding } from "../../shared/styles/styleUtils";
 
 export interface SidebarItemProps {
```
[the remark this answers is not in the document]
GeorgiSTodorov: Okay.

GeorgiSTodorov:
@Poltergeist , I agree with @nLight . Making it too specific would limit its usefulness.

packages/shared/styles/styleUtils/modifiers/backgroundColor.ts
GeorgiSTodorov: I can remove it here as well - https://github.com/dcos-labs/ui-kit/blob/90280672f59c9af244a8c2e5838294c3bb5969a7/packages/shared/styles/styleUtils/modifiers/modifierUtils.ts#L35 .

Poltergeist:
@nLight and @GeorgiSTodorov I was using products as a synonym for different variations. The question is do we need the flexibility to define paddings in the applications or wouldn't we rather prefer 
 to just use `<HeaderBarVersion1 />` instead of `<HeaderBar paddingHorizontal="xs" paddingVertical="xl" />`. I would argue that we probably would have one style of `headerBar` in a ui. Why aren't we using the `themeProvider` on the top level of the application and just use withTheme on this component? This could also be a viable solution which also forces consistency within one application plus we do not always need to provide the paddings for each instance of the `headerBar`.

GeorgiSTodorov:
@mperrotti , can we use https://github.com/dcos-labs/ui-kit/blob/90280672f59c9af244a8c2e5838294c3bb5969a7/packages/shared/styles/styleUtils/modifiers/modifierUtils.ts#L14 instead of `packages/types`?

mperrotti:
@Poltergeist - I agree that making a new component for any stylistic change feels a bit limiting. I could see us making a new component if the variant was functionally different, but these are minor stylistic changes.

That being said, I like the idea of `withTheme` and letting the host application use `ThemeProvider`

packages/appChrome/components/HeaderBar.tsx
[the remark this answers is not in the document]
mperrotti: I'm not suggesting letting the height be customizable, but if we have don't remove `height: 32px`, the top and bottom padding won't be accurate because it starts to take up the height. Here's two screenshots to demo what I'm talking about:
paddingandheight

screen shot 2018-11-27 at 9 53 33 am

packages/appChrome/components/Sidebar.tsx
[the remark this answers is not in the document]
mperrotti: "Light mode" is the default. I would just do something like
```
    const navClassNames = cx(
      sidebar,
      {
            darkMode: isHexDark(backgroundColor),
            [background(backgroundColor)]: Boolean(backgroundColor),
            [width(sidebarWidth)]: Boolean(width)
      }
    );
```

`cx` takes an object where the key is the classname, and the value is a boolean. If the boolean is true, the classname gets added

Poltergeist:
It is not that I am proposing to do a completely new component we would just not expose the options, we just have to generate a component which is setting these options. However, it feels like my opinion is so far off of everyone else's one and this kinda feels to disagree and commit for me. I will dismiss my review for now.

mperrotti:
Loving the knob functionality. Do they need to be in their own item in Storybook's sidebar? I'd prefer to keep one item per component, and right now AppChrome will have 2.

I've answered your questions (except for 5), but if anything is still unclear we can pair when I come online tomorrow

> Should we have the knobs stories in the same file - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R49

Yes, that's how've normally seen it done

> Should we still keep the non-knobs stories - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-d6fa03d51056b7fa98215af2d66344c3R525

Yes, unless you can think of a reason we shouldn't

> Is this important okay here - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-2888d54dab937f82b549310c963d7b4dR31 ? I couldn't find a better way to potentially override the other styles.

You don't need the important. You just have to pull the width styles out of `sidebar` like you did with all of the other themable styles.
You will also need to do this for `sidebarAnimator`

Also keep in mind that the sidebar width changes at a larger breakpoint, so the width theme actually needs to pass an object with the small screen width and the large screen width.

> You said to add this back, but with it the icon cannot be more than 24px - https://github.com/dcos-labs/ui-kit/pull/254/files#diff-3768d7bd77fce72475adfd42705edcfcL89 , how to overcome this?

The icon can be whatever size we want, we just have to make sure that value matches the icon size.

> How to implement the icon size there - #254 ?
This could be tricky...I need to take a little time to figure this out.

> I can't find your helper function - #254 .

Because it doesn't exist in master yet :)
I pointed you to a compare view using a commit from my Button branch. Sorry for the confusion

GeorgiSTodorov:
@tnguyen0 , do you mean that we should add a static padding or make the padding customizable?

GeorgiSTodorov:
@mperrotti when I try to access your helper function link I get:
screenshot from 2018-11-30 14-01-38

I changed the separate story item, once you start your work day, we can take a look at the sidebar width and the icon sizes together.

mperrotti:
Sorry about that, @GeorgiSTodorov. Here's a new link: https://github.com/dcos-labs/ui-kit/compare/master...134cc09f557f03b082fd526abc552e8531e30b5d#diff-c7218e3329e428d70b7790b006ce556bR161

GeorgiSTodorov:
@mperrotti Okay, will remove them.

GeorgiSTodorov:
Should be ready for review now.

package.json
```diff
@@ -41,6 +41,7 @@
   "typings": "dist/packages/index",
   "dependencies": {
     "emotion": "9.2.12",
+    "emotion-theming": "^9.2.9",
```
DanielMSchmidt: I think we use pinned dependencies here, so without the `^`

packages/appChrome/components/HeaderBar.tsx
```diff
@@ -10,20 +13,30 @@ export interface HeaderProps {
...
     const { children } = this.props;
+    /* tslint:disable:no-string-literal */
+    const HeaderBar = styled("div")`
+      background-color: ${props =>
+        props.theme.backgroundColor || purpleDarken4};
+      ${props =>
+        !props.theme.backgroundColor ||
+        (props.theme.backgroundColor && isHexDark(props.theme.backgroundColor))
+          ? darkMode
+          : null};
```
natmegs: A couple newbie emotion questions: Would `null;` be an issue here? Is there a benefit to having all these properties as functions instead of destructuring `theme` from props above with `children` and using them directly?

GeorgiSTodorov: These `props` aren't the same as the `this.props` above. Took me some time figuring out why `this.props.theme` didn't work.

GeorgiSTodorov:
@DanielMSchmidt what is the bug?

DanielMSchmidt:
@GeorgiSTodorov Sorry for the delay. If you see the attached screenshot, if I provide a very long section name it just gets cut somewhere in the middle instead of having either `...` or a line break in place. This seems broken to me.

nLight:
@DanielMSchmidt what's the behaviour on master? Was this bug introduced in this PR? Does it have anything to do with the AppChrome or rather menu item?

GeorgiSTodorov:
@nLight I think the scope changed a few times already. :D

DanielMSchmidt:
@nLight might be out of scope, lacking the knobs it was just harder to discover. I'm fine with the current state, let's see if this is even a problem out in production.

nLight:
@GeorgiSTodorov @DanielMSchmidt please capture this in a `type:ux` task

GeorgiSTodorov:
https://jira.mesosphere.com/browse/DCOS-45875

packages/appChrome/components/SidebarSection.tsx
```diff
@@ -27,14 +24,25 @@ class SidebarSection extends React.PureComponent<SidebarSectionProps, {}> {
...
+      padding-top: ${props =>
+        spaceSizes[props.theme.sidebarHeaderPaddingVert] || spaceSizes["s"]};
+    `;
+    /* tslint:enable:no-string-literal */
+
     return (
       <div>
         {label && (
-          <h3
+          <H3
```
mperrotti: Just curious - why a capital "H"?
It doesn't matter to the browser, so I'm not requesting a change, just curious if there was a reason for changing this :)

mperrotti:
The wrapping was intentional to support long translations strings, but I'm ok with truncation if the design team is ok with it.

mperrotti:
Pass the icon width to the submenu in stories that have submenus with icons:
screen shot 2018-12-04 at 9 46 05 am

The stories affected are:
- "AppChrome DCOS clone"
- "Sidebar w/ submenus (w/ icons)"

packages/appChrome/components/SidebarSection.tsx
GeorgiSTodorov: `H3` is my custom component - https://github.com/dcos-labs/ui-kit/pull/254/files/aa1bd95030f56ba0d105343df1d0d8b9da2c254a..97e3e540494c44b0a83eaceec0627653522285c1#diff-c449621d73b5c5204961ca0ba944428aR28 .

GeorgiSTodorov:
For the truncation please update the jira if you have any updates.

PR #255, opened 2018-11-28
DCOS-44052: add button components

mperrotti:
Created components to support all button types and states as spec'd by the design team.

**Screenshot from their Sketch library:**
screen shot 2018-11-27 at 12 52 22 pm

Note: the `:focus` styles only show up when using keyboard navigation (per Lily's request)

### DIFF
--- a/packages/button/README.md
+++ b/packages/button/README.md
@@ -1,0 +1,56 @@
+# Buttons
+
+Buttons are used for action items. Some examples include Create, Save, Delete. Buttons make it obvious to the user that they can take an action.
+
+Do not use buttons as navigational elements. Instead, use anchors because it takes the user to a new page and is not associated with an action.
+
+Each page will likely have one primary button. Any remaining calls-to-action are represented as secondary buttons.
+
+## Button states and types
+
+### Types
+
+**Primary**: Primary call to action button. Should stand out above all other actions. Do not put two primary buttons next to each other. Examples include Save, Deploy.
+
+**Secondary**: Most buttons on a page are likely this style.
+
+**Danger**: Destructive buttons are used to highlight something dangerous that can't be undone e.g. "Delete".
+
+**Icon only**: Use when an icon is enough to represent the action e.g. "Create a Service". Should be accompanied by a tooltip.
+
+**Icon with label**: If there is space and we feel an icon by itself is not obvious enough e.g. "Kill Task"
+
+### States
+
+**Disabled**: Disable a button when the action is not applicable but we want to show the user that it may be possible. Disabled buttons should have a tooltip.
+
+**Processing**: When user clicks or taps a button and an API request is initiated we should provide feedback so they know their action was received.
+
+## Button content
+
+Button labels tell users what will happen when they click the button. Use verbs that describe the action. Acceptable labels include:
+
+- New $object
+- Create $object
+- Edit $object
+- Save $object
+- Delete $object
+- Deploy, Scale, Restart
+
+Use sentence-style capitalization (only the first word in a phrase and any proper nouns capitalized) and no more than three words for button labels.
+
+For Sets of Buttons, use specific labels, such as Save or Cancel, instead of using Yes and No. This is particularly helpful when the user is confirming an action.
+
+## Feedback
+
+Buttons should have hover, active, focus, disabled states.
+
+When a user interacts with a button, either via click or tap, and an action is being completed, e.g. an API call, the button should change to disabled and provide feedback via the label to the user that it received the interaction and is busy performing the action e.g. “Saving…”
+
+## Tooltips
+
+Buttons may or may not require tooltips to help communicate their intent.
+
+Icon buttons that do not have labels should always have a tooltip so the user knows what will happen when they click it.
+
+Disabled buttons should always have a tooltip to explain why it is disabled so the user is aware.
--- a/packages/button/components/ButtonBase.tsx
+++ b/packages/button/components/ButtonBase.tsx
@@ -1,0 +1,187 @@
+import * as React from "react";
+import { cx } from "emotion";
+import FocusStyleManager from "../../shared/components/FocusStyleManager";
+import {
+  buttonReset,
+  textWeight,
+  flex,
+  flexItem,
+  display,
+  padding
+} from "../../shared/styles/styleUtils";
+import {
+  button,
+  buttonBase,
+  buttonInverse,
+  fullWidthButton,
+  focusStyleByAppearance,
+  getInverseMutedButtonStyles,
+  getMutedButtonStyles,
+  processingTextStyle
+} from "../style";
+
+export enum ButtonAppearances {
+  Primary = "primary",
+  Secondary = "secondary",
+  Standard = "standard",
+  Danger = "danger",
+  Success = "success"
+}
+
+export interface ButtonProps {
+  /**
+   * if the button triggers new content to appear (e.g.: modals and dropdowns)
+   */
+  ariaHaspopup?: boolean;
+  children?: React.ReactNode | string;
+  /**
+   * whether or not the button is enabled
+   */
+  disabled?: boolean;
+  /**
+   * the icon that appears before the button text
+   */
+  iconStart?: React.ReactElement<HTMLElement>;
+  /**
+   * the icon that appears after the button text
+   */
+  iconEnd?: React.ReactElement<HTMLElement>;
+  /**
+   * whether the action the button was meant to do has completed
+   */
+  isProcessing?: boolean;
+  /**
+   * if the button should fill the width of it's container
+   */
+  isFullWidth?: boolean;
+  /**
+   * if the button is on a dark background
+   */
+  isInverse?: boolean;
+  /**
+   * the function the button calls
+   */
+  onClick?: (e?: React.SyntheticEvent<HTMLElement>) => void;
+  /**
+   * the type of button - same as HTML "type" attribute on the <button> tag
+   */
+  type?: "button" | "reset" | "submit";
+}
+
+export interface ButtonBaseProps extends ButtonProps {
+  appearance: ButtonAppearances;
+  className?: string;
+}
+
+class ButtonBase extends React.PureComponent<ButtonBaseProps, {}> {
+  constructor(props) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+    this.buttonContentWithIcons = this.buttonContentWithIcons.bind(this);
+  }
+
+  public render() {
+    const {
+      appearance,
+      children,
+      className,
+      disabled,
+      iconStart,
+      iconEnd,
+      isInverse,
+      isProcessing,
+      isFullWidth,
+      onClick,
+      ...other
+    } = this.props;
+
+    return (
+      <FocusStyleManager
+        focusEnabledClass={focusStyleByAppearance(appearance, isInverse)}
+      >
+        <button
+          className={cx(
+            buttonReset,
+            button[appearance],
+            buttonBase,
+            textWeight("medium"),
+            display("block"),
+            className,
+            {
+              [fullWidthButton]: isFullWidth,
+              [buttonInverse[appearance]]: isInverse,
+              [getMutedButtonStyles(appearance)]: disabled || isProcessing,
+              [getInverseMutedButtonStyles(appearance)]:
+                (disabled || isProcessing) && isInverse
+            }
+          )}
+          disabled={disabled || isProcessing}
+          onClick={this.onClick}
+          tabIndex={0}
+          {...other}
+        >
+          {Boolean(iconStart) || Boolean(iconEnd) ? (
+            this.buttonContentWithIcons()
+          ) : (
+            <span className={isProcessing ? processingTextStyle : ""}>
+              {children}
+            </span>
+          )}
+        </button>
+      </FocusStyleManager>
+    );
+  }
+
+  private onClick(e: React.SyntheticEvent<HTMLElement>) {
+    if (!this.props.disabled && this.props.onClick) {
+      this.props.onClick(e);
+    }
+  }
+
+  private getIconStart(icon?: React.ReactElement<HTMLElement>) {
+    return (
+      Boolean(icon) && (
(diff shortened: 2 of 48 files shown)


packages/shared/styles/styleUtils/modifiers/modifierUtils.ts
```diff
@@ -22,18 +29,30 @@ const spaceSizes = {
...
   side: BoxSides,
   spaceSize?: SpaceSizes
 ) => {
   const size = spaceSize || "m";
 
-  if (side === "all") {
-    return `${property}: ${spaceSizes[size]} !important;`;
-  } else {
-    return `${property}-${side}: ${spaceSizes[size]} !important;`;
+  switch (side) {
```
mperrotti: I originally made this update to apply padding from standard spaces to the top and bottom of the button. When I reviewed with Lily, she told me the design team is intentionally deviating from standard spacing for button padding (currently 10px vertical, 18px horizontal).

Instead of removing this code I already wrote, I kept it in for future work

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+      <button
+        className={cx(buttonReset, className, {
+          [fullWidthButton]: isFullWidth
+        })}
+        disabled={disabled || isProcessing}
+        onClick={this.onClick}
+        tabIndex={0}
+        {...other}
+      >
+        <span
```
mperrotti: This span wrapper might seem like an odd choice, but it's necessary for only showing the `:focus` style to keyboard users.

If anybody is super upset with this markup, I guess I could handle that in JS instead :/

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+} from "../../shared/styles/styleUtils";
+import {
+  button,
+  buttonBase,
+  buttonContent,
+  buttonInverse,
+  fullWidthButton,
+  getInverseMutedButtonStyles,
+  getMutedButtonStyles,
+  processingText
```
Poltergeist: maybe we want to import some of these classes with a different name like `processingText as processingTextStyle` it gets confusing to parse that processingText is a style class and not a variable containing processing text.

packages/shared/icons.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+
+// TODO: once we have an icon library, use that instead
+export const DownTriangle = () => (
+  <svg
```
Poltergeist: Can we think about how we could achieve this with svg sprites instead? I would like to save on the size of the dom tree.

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+      children,
+      className,
+      disabled,
+      iconStart,
+      iconEnd,
+      isInverse,
+      isProcessing,
+      isFullWidth,
+      onClick,
+      ...other
```
Poltergeist: This long props list makes me feel like this component might be to configurable instead of being composeable. For example `iconStart` and `iconEnd` why do we need these as props couldn't we just provide the icons with the Children like this:
```JS
<Button>
  <Icon name="xyz"/>
  Loading
  <Icon name="end" />
</Button>
```
And then check the Button children for Icons component if we actually need different styling for these. 

Why do we need the `isInverse` prop? isn't that just another appearance?

packages/shared/styles/color.ts
```diff
@@ -124,21 +124,38 @@ export const customColors = (): Colors => {
...
+};
+
+export const hexToRgbA = (hex: string, alpha: string | number = 1): string => {
+  if (isValidHex(hex)) {
+    return `rgba(${hexToRgbArr(hex).join(",")},${alpha})`;
+  }
   return "rgba(0,0,0,0)";
 };
+
+// TODO: when/if we start using rgb/rgba colors in
```
DanielMSchmidt: Can we attach a JIRA to this todo?

packages/shared/styles/color.ts
```diff
@@ -124,21 +124,38 @@ export const customColors = (): Colors => {
...
+// design-tokens, write color functions that can get
+// rgb/rgba brightness as well
+export const getHexBrightness = (hex: string) => {
+  const rgbArr = hexToRgbArr(hex);
+  return (rgbArr[0] * 299 + rgbArr[1] * 587 + rgbArr[2] * 114) / 1000;
```
DanielMSchmidt: Can we move these magic numbers to constants that give them a meaning?

packages/shared/styles/styleUtils/modifiers/modifierUtils.ts
```diff
@@ -8,7 +8,14 @@ import {
...
 } from "../../spacing";
 
-export type BoxSides = "all" | "top" | "right" | "bottom" | "left";
+export type BoxSides =
+  | "all"
+  | "top"
+  | "right"
+  | "bottom"
+  | "left"
+  | "horiz"
```
DanielMSchmidt: Why the abbreviation?

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  iconStart?: React.ReactElement<HTMLElement>;
+  /**
+   * the icon that appears after the button text
+   */
+  iconEnd?: React.ReactElement<HTMLElement>;
+  /**
+   * whether the action the button was meant to do has completed
+   */
+  isProcessing?: boolean;
```
Poltergeist: Isn't `isActive` a better name here?

packages/button/components/StandardButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+  ButtonAppearances
+} from "./ButtonBase";
+import {
+  greyLightDarken1,
+  white
+} from "../../design-tokens/build/js/designTokens";
+import { focusStyles } from "../style";
+import { hexToRgbA } from "../../shared/styles/color";
+
+const StandardButton = (props: ButtonProps) => (
```
Poltergeist: How do you feel about omitting the `Standard`?  Shouldn't this just be ButtonBase?

packages/button/components/ButtonBase.tsx
brandonc: I prefer this-- active could be confused for what happens while the mouse is down

packages/shared/styles/styleUtils/modifiers/display.ts
```diff
@@ -16,6 +17,6 @@ export const display = (
     | "table-cell"
 ) => {
   return css`
-    display: ${displayType} !important;
+    display: ${displayType};
```
brandonc: what will break without this?

packages/button/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Danger**: Destructive buttons are used to highlight something dangerous that can't be undone e.g. "Delete".
+
+**Icon only**: Use when an icon is enough to represent the action e.g. "Create a Service". Should be accompanied by a tooltip.
+
+**Icon with label**: If there is space and we feel an icon by itself is not obvious enough e.g. "Kill Task"
+
+### States
+
+**Disabled**: Disable a button when the action is not applicable but we want to show the user that it may be possible. Disabled buttons should have a tooltip.
```
brandonc: Should ButtonBase implement this recommendation through a disabledHelpText prop or something?

packages/button/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+When a user interacts with a button, either via click or tap, and an action is being completed, e.g. an API call, the button should change to disabled and provide feedback via the label to the user that it received the interaction and is busy performing the action e.g. “Saving…”
+
+## Tooltips
+
+Buttons may or may not require tooltips to help communicate their intent.
+
+Icon buttons that do not have labels should always have a tooltip so the user knows what will happen when they click it.
+
+Disabled buttons should always have a tooltip to explain why it is disabled so the user is aware.
```
brandonc: Repeated

packages/button/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Icon with label**: If there is space and we feel an icon by itself is not obvious enough e.g. "Kill Task"
+
+### States
+
+**Disabled**: Disable a button when the action is not applicable but we want to show the user that it may be possible. Disabled buttons should have a tooltip.
+
+**Processing**: When user clicks or taps a button and an API request is initiated we should provide feedback so they know their action was received.
+
+## Button content
```
brandonc: Useful README

packages/shared/styles/color.ts
mperrotti: It's just the standard formula for calculating luminance on RGB: https://en.wikipedia.org/wiki/Relative_luminance

The numbers are based on how much "brightness" our eyes perceive from each color channel, so I'm not sure how much less magic they'll seem if we convert them to constants

packages/shared/styles/styleUtils/modifiers/modifierUtils.ts
mperrotti: Less typing ¯\_(ツ)_/¯

packages/shared/icons.tsx
mperrotti: When we build out the icon distribution, sure :)

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+            {
+              [buttonInverse[appearance]]: isInverse,
+              [getMutedButtonStyles(appearance)]: disabled || isProcessing,
+              [getInverseMutedButtonStyles(appearance)]:
+                (disabled || isProcessing) && isInverse
+            }
+          )} ${buttonContent}`}
+          tabIndex={-1}
+        >
+          {Boolean(iconStart) || Boolean(iconEnd) ? (
```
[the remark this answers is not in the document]
mperrotti: Yes, that seems cleaner

packages/button/components/ButtonBase.tsx
mperrotti: We _could_ just require they get passed in as children, but there's some layout and spacing styles attached to the divs that wrap icons, and I don't want engineers to have to worry about applying those styles every time they need to use a button with an icon

packages/button/components/ButtonBase.tsx
mperrotti: Agreed with Brandon, and "processing" is how the design team refers to that state, and it would be nice if we could use shared language

packages/button/components/StandardButton.tsx
mperrotti: The design team calls it "Standard", so I'd rather we adopt their language.

packages/button/components/ButtonBase.tsx
mperrotti: You're right, that's more understandable

packages/button/components/ButtonBase.tsx
mperrotti: The appearances have meaning attached to them (see the README). `isInverse` is just a way to adapt any button appearance to a dark background.

packages/button/components/ButtonBase.tsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  isProcessing?: boolean;
+  /**
+   * if the button should fill the width of it's container
+   */
+  isFullWidth?: boolean;
+  /**
+   * if the button is on a dark background
+   */
+  isInverse?: boolean;
```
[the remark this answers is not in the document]
mperrotti: The README didn't mention it because the design team hadn't considered buttons on dark backgrounds until I started working on this ticket. I'll collaborate with Lily and make an update :)

isInverse is just a way to adapt any button styles for a dark background. `StandardButton` and the disabled button style are good examples

DanielMSchmidt:
@Poltergeist Could you take another look?

packages/button/components/ButtonBase.tsx
Poltergeist: Okay sounds good to me.

packages/button/components/StandardButton.tsx
Poltergeist: ✅

packages/shared/icons.tsx
Poltergeist: So this is not here to stay great. :-)

packages/utilities/components/interactionMode.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.docEl.removeEventListener("keydown", this.handleKeyDown);
+    this.docEl.removeEventListener("mousedown", this.handleMouseDown);
+  }
+
+  private handleKeyDown = e => {
+    if (e.which === TAB_KEY_CODE) {
+      this.reset();
+      this.docEl.addEventListener("mousedown", this.handleMouseDown);
+    }
+  };
```
TattdCodeMonkey: The linter in CI is complaining about this semicolon.
> ERROR: /jenkins/workspace/tend_UI-Kit_Pipeline_PR-255-LNJE7LCH5OIWLFWM5ZD2BNVWJSL3Q6C22KYJBBKQ6KMVE3NZUWSA/packages/utilities/components/interactionMode.ts[31, 4]: Unnecessary semicolon

packages/utilities/components/interactionMode.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  private handleMouseDown = () => {
+    this.reset();
+    this.container.classList.remove(this.className);
+    this.docEl.addEventListener("keydown", this.handleKeyDown);
+  };
```
TattdCodeMonkey: This one too.
> ERROR: /jenkins/workspace/tend_UI-Kit_Pipeline_PR-255-LNJE7LCH5OIWLFWM5ZD2BNVWJSL3Q6C22KYJBBKQ6KMVE3NZUWSA/packages/utilities/components/interactionMode.ts[37, 4]: Unnecessary semicolon

packages/utilities/components/interactionMode.ts
mperrotti: Yea, I know. I've gotta figure out a way to shut it up

PR #256, opened 2018-11-30
DCOS-45314: Dropdownable

brandonc:
### Overview

Dropdownable is a utility component that facilitates a simple dropdownable interaction: if the open prop is true, a dropdown component prop is rendered relative to the position of the Dropdownable's children according to a list of preferences and available space. When the user clicks outside of the component, it is hidden. The dropdown is rendered using an overlay that is appended to the document body so it should appear on top of everything else.

~You can also specify through a matchWidth prop that the dropdown should match the width of the thing it appears to drop from.~

One key decision I made was to eagerly render the contents of the dropdown onto the dom, hidden, so that it can be measured for viewport fit when it is shown. Without this eager rendering, we would have to rely on callers to specify the render direction (probably without respecting the available viewport area)

Don't hold back on the feedback.

### Screenshots

dec-05-2018 16-25-47

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,13 +44,13 @@
     "emotion-theming": "9.2.9",
     "memoize-one": "4.0.2",
     "react": "16.5.2",
-    "react-click-outside": "^3.0.1",
+    "react-click-outside": "3.0.1",
     "react-dom": "16.5.2",
     "react-emotion": "9.2.12",
     "react-toggled": "1.2.7",
     "react-transition-group": "2.5.0",
     "react-virtualized": "9.20.1",
-    "semantic-release": "^15.9.6"
+    "semantic-release": "15.9.6"
   },
   "devDependencies": {
     "@commitlint/cli": "7.2.0",
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "emotion-theming": "9.2.9",
     "memoize-one": "4.0.2",
     "react": "16.5.2",
+    "react-click-outside": "^3.0.1",
     "react-dom": "16.5.2",
     "react-emotion": "9.2.12",
     "react-toggled": "1.2.7",
--- a/packages/dropdownable/components/Dropdownable.tsx
+++ b/packages/dropdownable/components/Dropdownable.tsx
@@ -39,6 +39,7 @@ interface State {
 const METHODS_TO_BIND = [
   "getDropdown",
   "setPosition",
+  "setPositionFromCurrentProps",
   "calculateBestDirection",
   "dropdownStyles",
   "positionForDirection",
@@ -73,12 +74,12 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
   }
 
   componentDidMount() {
-    window.addEventListener("resize", () => this.setPosition());
+    window.addEventListener("resize", this.setPositionFromCurrentProps);
     this.setPosition();
   }
 
   componentWillUnmount() {
-    window.removeEventListener("resize", () => this.setPosition());
+    window.removeEventListener("resize", this.setPositionFromCurrentProps);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -129,6 +130,10 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
     );
   }
 
+  setPositionFromCurrentProps() {
+    this.setPosition();
+  }
+
   setPosition(props: DropdownableProps = this.props) {
     const dropdownDimensions = this.dropdownDimensions();
     const windowDimensions = this.windowDimensions();
@@ -142,7 +147,7 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
 
     this.setState({
       direction,
-      position: this.positionForDirection(direction, childBounds),
+      position: this.positionForDirection(direction, childBounds)
     });
   }
 
--- a/packages/dropdownable/components/Dropdownable.tsx
+++ b/packages/dropdownable/components/Dropdownable.tsx
@@ -164,21 +164,18 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
     windowDimensions
   ): Direction {
     // Determine if there is enough space in each direction to fit dropdown
+    // There is always enough room to be aligned to the right or left
     const possibleDirections = {
       top: dropdownDimensions.height <= childBounds.top,
-      right:
-        dropdownDimensions.width + childBounds.right <= windowDimensions.width,
       bottom:
         childBounds.bottom + dropdownDimensions.height <=
-        windowDimensions.height,
-      left:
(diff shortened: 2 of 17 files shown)


packages/dropdownable/stories/Dropdownable.stories.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import { withInfo } from "@storybook/addon-info";
+
```
DanielMSchmidt: I would love to see some knobs especially for preferred directions and open in place to make the demo more interactive.

package.json
```diff
@@ -43,6 +43,7 @@
     "emotion": "9.2.12",
     "memoize-one": "4.0.2",
     "react": "16.5.2",
+    "react-click-outside": "^3.0.1",
```
nLight: We try to avoid any version ranges. Let's please do "3.0.1"

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,235 @@
...
+    });
+  }
+
+  componentDidMount() {
+    window.addEventListener("resize", () => this.setPosition());
+    this.setPosition();
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("resize", () => this.setPosition());
```
nLight: I didn't know it works that way. Arrow function creates a new function every time thus #L77 and #L82 will produce different pointers, thus, I thought, it won't remove listener. Will it?

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.open !== nextProps.open) {
+      this.setState({ dropdownWidth: this.childBounds().width });
+      this.setPosition(nextProps);
```
nLight: Is this for optimization sake? I believe react advises to recalculate what can be calculated on every `render`

packages/shared/components/Overlay.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  getContents() {
+    const { className, children } = this.props;
+
+    return <div className={className}>{children}</div>;
+  }
+
+  render() {
+    return ReactDOM.createPortal(this.getContents(), this.el);
```
nLight: that wouldn't work with react@15 we're using in the dcos-ui, right?

brandonc: It will not. The decision was made between mike and rodney to wait until we upgrade react to use these components in dcos-ui. Interestingly, ui-kit package.json already stipulates react 16 so how does any of this work?? 🙃

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,235 @@
+import React from "react";
+import styled, { css } from "react-emotion";
+
+import Overlay from "../../shared/components/Overlay";
+import DropdownContents from "./DropdownContents";
+
+export enum Direction {
```
mperrotti: We'll probably need to support a dropdown that's centered relative to it's parent's width

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,235 @@
...
+      }
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    window.addEventListener("resize", () => this.setPosition());
```
mperrotti: This might be a good place to use `requestAnimationFrame` to throttle how often we need to set the position.

I'm not seeing much of a difference in frame rate when the dropdown is open vs when it's closed, so I don't think this is a blocker.

brandonc:
@DanielMSchmidt What you may be seeing with the flipping is that there is not enough room for either position as you change size and so it reverts to the first choice. I'll try to make the interactive example better.

brandonc:
@mperrotti _droparoundable_ ? :trollface:

natmegs:
What is the desired resize/positioning of the dropdown when it is cut off? Recording of what I'm seeing (stays in current position when screen is resized to cut off, appears in preferred direction even if cut off):
resize

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,246 @@
...
+        dropdownDimensions
+      )
+    });
+  }
+
+  calculateBestDirection(
+    preferredDirections,
+    childBounds,
+    dropdownDimensions,
+    windowDimensions
```
natmegs: It would be helpful to type these params as well as `positionForDirection` params, and `childBounds`, `windowBounds`, and `dropdownDimensions` return types

GeorgiSTodorov:
@natmegs Not sure about the cut off, but I added some types. I am not that familiar with this PR, so please check if I missed or mistyped anything.

brandonc:
@natmegs taking a look

brandonc:
@natmegs There were two problems: When there is enough scroll space in the document, as in your example, I consider that enough space to drop down. I didn't fix that. I also didn't update the fit logic to match the revised positioning scheme, so it was totally broken anyway. I fixed it and added a new storybook variant to demonstrate auto fit top/bottom. There is no auto fit necessary for left/right because now it just aligns with the child.

brandonc:
dec-06-2018 09-53-33

brandonc:
@GeorgiSTodorov I rebased out the merge commit and force pushed the branch

mperrotti:
It may actually just be a bug that happens with the story after I interact with the "knobs" on the first story. That seems to be the only way I can _sometimes_ get it to happen.

Let me know if you can reproduce

brandonc:
@mperrotti I think it  has to do with the width of the dropdown not being fixed. Sometimes it's more narrow and the content takes up more height

mperrotti:
You're right - when I change `minWidth` to `width`, it's fine :)

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -0,0 +1,251 @@
...
+  Direction.BottomRight,
+  Direction.TopRight,
+  Direction.TopLeft
+];
+
+export interface DropdownableProps {
+  open: boolean;
+  dropdown: React.ReactNode;
+  preferredDirections?: Direction[];
+  onClose?: () => void;
```
nLight: it's optional here with no default value, but is required in `DropdownContents`, potential 💥 ?

brandonc: https://github.com/dcos-labs/ui-kit/pull/256/commits/538a723b168be2c487f940b991535c0d0b594cb2

PR #257, opened 2018-11-30
[DCOS-41079] Input components

TattdCodeMonkey:
Created `TextInput` & `TextInputWithIcon` components to be used in forms.

## Screenshots
image
image


mperrotti:
I noticed you were using style utilities in `style.ts`. There's nothing "wrong" about this at all, but I like to use them as class names so when an engineer opens up the component, they can get a sense of what's going on just by looking at the markup.
An added "bonus" (in my opinion) is that `style.ts` contains mostly styles that are specific to the needs of that component

In the world of regular CSS, this kind of approach is called "Functional CSS" if you want to read more

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+        type={this.props.type}
+        onFocus={this.inputOnFocus}
+        onBlur={this.inputOnBlur}
+        {...inputElementProps}
+      />
+    );
+  }
+
+  protected inputOnFocus(e: React.FocusEvent<HTMLInputElement>) {
+    this.setState({ focus: true });
```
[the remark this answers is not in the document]
TattdCodeMonkey: I originally had it using [`:focus-within`](https://css-tricks.com/almanac/selectors/f/focus-within/) since I'm styling the div/span that contains the input, but I was concerned about browser support for that. But I can go back to that, as I'd prefer to not have state here if I can avoid it.

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { cx } from "emotion";
+import { inputContainer, inputLabel, inputElement } from "../style";
+
+export enum TextInputAppearance {
+  Default = "default",
+  Error = "error",
+  Success = "success"
+}
+
+export interface TextInputProps extends React.HTMLProps<HTMLInputElement> {
```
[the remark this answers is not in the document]
TattdCodeMonkey: 👍  I do this if `id` is provided, but currently id is optional. I'll add that to be a required prop.

packages/forms/style.ts
```diff
@@ -0,0 +1,91 @@
...
+  ${flex()};
+  ${inputAppearances[appearance]};
+`;
+
+export const inputElement = css`
+  ${flexItem("grow")};
+  background: none;
+  border: 0;
+  box-shadow: none;
+  order: 1;
```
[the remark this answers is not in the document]
TattdCodeMonkey: I think I added these in preparation of doing the icons. but ended up being able to put them in the correct source order. I removed order, thanks.

packages/forms/style.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export const inputLabel = css`
+  margin-top: 0;
+  margin-bottom: 4px;
+  font-weight: 600;
+  display: block;
+`;
+
+export const inputContainer = appearance => css`
```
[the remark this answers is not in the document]
TattdCodeMonkey: when we apply the style to the container vs the input directly focus has to be handled differently doesn't it? on the `input` we can do `:focus` but on the container we would have to use `:focus-within`. should I be concerned about browser support for `:focus-within`?

packages/forms/style.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export const inputContainer = css`
+  border: 1px solid;
+  border-radius: 4px;
+  height: 36px;
+`;
+
+export const iconEnd = css`
+  order: 2;
```
TattdCodeMonkey: @mperrotti I was able to remove all of the `order` properties except this one. Not sure why it wasn't adhering to source order, but without this the `endIcon` is rendered directly next to the `startIcon`.

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+  id: string;
+  type: "text" | "number" | "search" | "email" | "password" | "tel" | "url";
+  appearance: TextInputAppearance;
+  inputLabel: string | React.ReactNode;
+  showInputLabel: boolean;
+  validationContent?: string | React.ReactNode;
```
TattdCodeMonkey: I don't love this prop name, but `validationMessage` and `errorMessage` didn't seem right to me either. But only slightly. Open to renaming this anything better.

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+  textWeight
+} from "../../shared/styles/styleUtils";
+
+export enum TextInputAppearance {
+  Standard = "standard",
+  Error = "error",
+  Success = "success"
+}
+
+export interface TextInputProps extends React.HTMLProps<HTMLInputElement> {
```
TattdCodeMonkey: is `extends React.HTMLProps<HTMLInputElement>` the best way to ensure we support all the same props that `<input>` does?

packages/forms/components/TextInput.tsx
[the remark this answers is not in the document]
DanielMSchmidt: [Seems to be problematic on edge](https://caniuse.com/#search=focus-with). I am missing context here, so I don't feel confident to do a judgement if that is an ok trade-off.

packages/forms/style.ts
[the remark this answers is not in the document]
mperrotti: You're right, `:focus-within` is not supported in Microsoft browsers: https://caniuse.com/#search=focus-within

We could add/remove classes on blur/focus. I'm not sure if visually indicating `:focus` is something we can consider an enhancement for "good" browsers

packages/forms/style.ts
mperrotti: That's odd. I'll take a look

packages/forms/components/TextInput.tsx
mperrotti: I'm comfortable with this name 👍

packages/forms/components/TextInput.tsx
mperrotti: Yea, I should actually update my Button PR to do something similar :)

packages/forms/style.ts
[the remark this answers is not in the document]
TattdCodeMonkey: ok, I'll update `TextInputWithIcons` to add/remove classes with `onFocus` & `onBlur` but leave `TextInput` with the simpler implementation of just using CSS :)

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  protected getLabelContent() {
+    return (
+      <label
+        className={cx(
+          flush("top"),
+          margin("bottom", "xxs"),
+          textWeight("medium"),
+          display(this.props.showInputLabel ? "block" : "none")
```
mperrotti: `display: none` and `visibility: hidden` will cause assistive technology to ignore the label (e.g.: causes screen readers not to read the label aloud)

We should add something to style utils to handle this. Maybe something like...

packages/shared/styles/styleUtils/modifiers/visuallyHidden.ts:
````
export const visuallyHidden = css`
		border: 0;
		clip: rect(0 0 0 0);
		position: absolute;
		overflow: hidden;
		margin: -1px;
		padding: 0;
		width: 1px;
		height: 1px;
`
````

File: packages/forms/stories/TextInput.stories.tsx [hunk 9]
```diff
@@ -0,0 +1,72 @@
...
+
+const defaultDisplayItem = css`
+  width: 160px;
+`;
+
+storiesOf("Forms/TextInput", module)
+  .addDecorator(withReadme([readme]))
+  .addWithInfo("default", () => (
+    <div>
+      <TextInput
```
mperrotti: These don't have an `id` prop passed, but it looks like it's required in your types. Why aren't we getting an error? 🤔

mperrotti:
Font size of text input should be the same as body content. The browser is currently applying it's user-agent styles for font-size, but if you set `font-size: inherit;` on the input, it should take the same font size as it's parent

screen shot 2018-12-04 at 11 32 25 am

mperrotti:
Error styling is just slightly off:
- Label should also be red
- Error message font size should be `fontSizeS` (12px)
- Error message is too close to the input box. Apply 4px top padding

screen shot 2018-12-04 at 10 26 46 am

mperrotti:
Maybe instead we should have each form component be a package. So instead of `packages/forms/components/[every form component]`, we should do `packages/forms/components/textInput`, `packages/forms/components/radioInput`, `packages/forms/components/selectInput`, etc...

@Poltergeist, what do you think?

mperrotti:
Maybe we need a separate ticket for rendering informational text below the input. Here are some issues I came across:
1. I just realized there was something I forgot in the ticket - a way to add a little helper description below an input. Current behavior is that this description goes away when there's validation content to show
screen shot 2018-12-04 at 12 00 41 pm
screen shot 2018-12-04 at 12 00 49 pm

2. a11y for showing errors could be improved
- Set `ariaInvalid={true}` to the input when there's an error
- Associate error messages with the input. For example:
````
<input aria-describedby="errorMsg" />
<span id="errorMsg" />
````

3. We should probably support multiple error messages

File: packages/forms/README.md [hunk 10]
```diff
@@ -0,0 +1,32 @@
+# Forms
+
+## TextInput
+`TextInput` is a component used to enter text for forms.
+
+### Required Props
+
+ - `id`: String unique identifier used for the form input element.
```
mperrotti: If you move these to comments above the type definitions, they'll render in the prop table.
e.g.: https://github.com/dcos-labs/ui-kit/blob/55d37c6a751457ff627f3120856af4d6898cd3d7/packages/button/components/ButtonBase.tsx#L31

packages/forms/README.md
```diff
@@ -0,0 +1,32 @@
+# Forms
```
mperrotti: Some of this information seems redundant with what we render in Storybook. I'd recommend we task @lsamimi with writing text input documentation

mperrotti:
I'm going to investigate whether we can simplify this prop table to only include props we're setting up - not EVERY html attribute an <input /> can have
screen shot 2018-12-04 at 11 30 42 am

packages/forms/stories/TextInput.stories.tsx
TattdCodeMonkey: it is showing me a warning in my IDE, but not an error. I'll update the stories.

mperrotti:
One more thing we forgot to handle: marking fields as "required". It's missing from the design team's Sketch file, so I need to confirm that this is the right styling:
screen shot 2018-12-04 at 12 50 02 pm

TattdCodeMonkey:
I wondered about that too. Its in the screenshot on the ticket though. I can easily add it, my only concern would then be should we add built-in support for the tooltip? To ensure it is always rendered after the asterisk if set?

TattdCodeMonkey:
@mperrotti for `3. We should probably support multiple error messages` Should I change `validationContent` to `errors?: string[]` (or `errorMessages?: string[]`)? not as flexible but then we can ensure they render correctly and `a11y` attributes are set correctly.

packages/forms/style.ts
```diff
@@ -0,0 +1,101 @@
...
+      svg {
+        fill: ${purple};
+      }
+    }
+  `,
+  disabled: css`
+    background-color: ${greyLightLighten3};
+    border-color: ${greyLightLighten3};
+    svg {
+      fill: ${greyDarkLighten3};
```
TattdCodeMonkey: @mperrotti or @lilysamimi this was a total guess on my part so that icons are still displayed when the input is disabled, but are lighter then normal. Is that acceptable or desired?

mperrotti: We should try and normalize the text and icon colors. Set `color: ${whateverColor}`, use the [::placeholder selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder), use the same color for all three: icon, placeholder text, and value text

TattdCodeMonkey: Shouldn't placeholder text be lighter then value text? 

image

TattdCodeMonkey:
Thanks @GeorgiSTodorov, I needed to rebase, then update snapshots. should ✅  this time 🤞

packages/shared/styles/styleUtils/resets/inputReset.ts
```diff
@@ -0,0 +1,12 @@
+import { css } from "emotion";
+
+export const inputReset = css`
+  background: none;
+  border: 0;
+  box-shadow: none;
+  order: 1;
```
mperrotti: Remove this

packages/forms/style.ts
mperrotti: It's because you had `order: 1` on your input reset :)

packages/forms/components/TextInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+    return inputElementProps;
+  }
+
+  protected getInputElement(additionalClasses: string[] = []) {
+    const inputElementProps = this.getInputElementProps();
+
+    return (
+      <input
```
mperrotti: You're getting React props in your HTML:
screen shot 2018-12-05 at 12 55 08 pm

packages/forms/style.ts
```diff
@@ -0,0 +1,111 @@
...
+    border-color: ${green};
+    svg {
+      fill: ${green};
+    }
+  `
+};
+
+export const inputContainer = css`
+  border: 1px solid;
+  border-radius: 4px;
```
mperrotti: There was something in the button PR that added border-radius design tokens, so now you can do `border-radius: ${borderRadiusSmall};`

packages/forms/style.ts
```diff
@@ -0,0 +1,111 @@
...
+  height: 36px;
+  font-size: inherit;
```
mperrotti: You don't need this :)

packages/forms/components/TextInputWithIcon.tsx
```diff
@@ -0,0 +1,132 @@
...
+      return null;
+    }
+    return (
+      <span
+        className={cx(
+          padding("left", "xs"),
+          flex({ align: "center", justify: "center" }),
+          flexItem("shrink"),
+          iconEnd
+        )}
```
mperrotti: Add `flush("left")`.
I thought I was being "helpful" by adding gutters for flex items, but it seems to be hurting more than helping in a lot of cases...

packages/forms/components/TextInput.tsx
TattdCodeMonkey: Doh! I need to omit the icon pros in `TextInputWithIcon`

packages/forms/style.ts
TattdCodeMonkey: without this the font-size is incorrect for `TextInput`. I have both this and the `input {` here to handle when we apply this style directly to the input in `TextInput`, as well as when we apply it to the div container for `TextInputWithIcon`.

mperrotti:
>for 3. We should probably support multiple error messages Should I change validationContent to errors?: string[] (or errorMessages?: string[])? not as flexible but then we can ensure they render correctly and a11y attributes are set correctly.

I think `errors?: string` makes sense. We can map that array to `<li>`s instead of spitting them all out in a span

mperrotti:
Re: label tooltip
> I wondered about that too. Its in the screenshot on the ticket though. I can easily add it, my only concern would then be should we add built-in support for the tooltip? To ensure it is always rendered after the asterisk if set?

I think this is a separate task: https://jira.mesosphere.com/browse/DCOS-45978

packages/textInput/README.md
```diff
@@ -0,0 +1,32 @@
+# Forms
+
+## TextInput
+`TextInput` is a component used to enter text for forms.
+
+### Required Props
```
mperrotti: This information is captured from the comments above the props in `TextInput`

packages/textInput/README.md
```diff
@@ -0,0 +1,32 @@
+# Forms
```
mperrotti: @lsamimi - how do you feel about this README?

GeorgiSTodorov:
Leaving this unmerged in case Rodney wants to update the README.

nLight:
Oh, merged then saw `Leaving this unmerged in case Rodney wants to update the README.`

nLight:
@TattdCodeMonkey feel free following up with a PR

PR #261, opened 2018-12-06
[spike] Replace values duplicated by design tokens

mperrotti:
https://jira.mesosphere.com/browse/DCOS-46024


mperrotti:
This shouldn't merge without https://github.com/dcos-labs/ui-kit/pull/260 - it will break flexItem gutters, which have been fixed in that PR

GeorgiSTodorov:
Conflicts, because I merged https://github.com/dcos-labs/ui-kit/pull/260 .

mperrotti:
I think this is unrelated to these changes, but definitely something I'll fix. Thanks for catching!

mperrotti:
That error appears to be coming from `action` (from `@storybook/addon-actions`), and it only happens in Firefox. Also, the function is still called successfully.

I don't think this is something we have to worry about fixing - it doesn't happen if I pass in my own function

GeorgiSTodorov:
File conflicts.

PR #263, opened 2018-12-06
[spike] Let Button component accept html button props

mperrotti:
https://jira.mesosphere.com/browse/DCOS-46033


packages/button/components/ButtonBase.tsx
```diff
@@ -66,6 +62,7 @@ export interface ButtonProps {
    * the type of button - same as HTML "type" attribute on the <button> tag
    */
   type?: "button" | "reset" | "submit";
+  ref?: any;
```
nLight: is this really any?

TattdCodeMonkey: Wouldn't `ref` be stored on the class and not the props anyway? you need to create the ref when rendering. I think the type would be `React.RefObject<HTMLButtonElement>`?

[Good article on React refs & TS](https://medium.com/@martin_hotell/react-refs-with-typescript-a32d56c4d315)

mperrotti: I tried that type and it didn't work. The ref is actually created in `FocusStyleManager`. I'll see what I can do to get rid of the `any`. Thanks for the article!

GeorgiSTodorov:
Is `PrimaryButton` having `undefined` class expected? It's the same on master, but still.
screenshot from 2018-12-13 12-03-38

mperrotti:
I agree, `undefined` classname is weird. Will fix

GeorgiSTodorov:
A file conflict, because we merged https://github.com/dcos-labs/ui-kit/pull/264 . Since you have a lot of stuff to do, I hope you don't mind if I fix it.

mperrotti:
@GeorgiSTodorov - we already have our own validation:
screen shot 2018-12-14 at 3 35 31 pm

`no-validate` is just telling the browser not to do it's own validation

And the value that gets sent is formatted with a period - you can see this by changing the number (while using a comma separated decimal) and looking at the `value` attribute of the `<input />`

GeorgiSTodorov:
`console.log` is still there. :D

packages/button/components/ButtonBase.tsx
```diff
@@ -93,7 +89,7 @@ class ButtonBase extends React.PureComponent<ButtonBaseProps, {}> {
       isFullWidth,
       onClick,
       ...other
-    } = this.props;
+    } = this.props as ButtonBaseProps;
```
pierrebeitz: Event without that explicit typecase everything seems fine. Let's remove it!

packages/button/style.ts
```diff
@@ -187,21 +190,31 @@ export const button = {
...
+      return css`
+        ${tintContent(purpleLighten3)};
+        &:hover {
+          ${tintContent(purpleLighten2)};
+        }
+        &:active {
+          ${tintContent(purpleLighten1)};
+        }
+      `;
+      break;
```
pierrebeitz: no need to `break` if we're already returning before.

packages/button/style.ts
```diff
@@ -187,21 +190,31 @@ export const button = {
...
+    case "standard":
+      return css`
+        ${filledButton(
+          hexToRgbA(white, 0.3),
+          hexToRgbA(white, 0.4),
+          hexToRgbA(white, 0.45)
+        )};
+      `;
+      break;
```
pierrebeitz: no need to `break` if we're already returning before.

packages/button/style.ts
```diff
@@ -187,21 +190,31 @@ export const button = {
...
+    default:
+      return "";
```
pierrebeitz: is this intentional or do we might want to add a `error("Hey, you used a wrong appearance here!")`?

packages/button/tests/ButtonBase.test.tsx
```diff
@@ -17,7 +17,7 @@ describe("ButtonBase", () => {
           appearance={ButtonAppearances[appearance]}
           isFullWidth={true}
           type="submit"
-          ariaHaspopup={true}
+          aria-haspopup={true}
```
pierrebeitz: I'm new, so i ask dumb questions. What is changing (or not changing) due to this change?

My best bet would be that before `ButtonBase` was in charge of rendering this prop and now we directly let `button` do this? If so, where can i see that this is working? With a brief look at the snapshots i don't find that property. help!

mperrotti: That's odd - I'm not sure why this wasn't showing up in previous snapshots.

You're currect though, it changes because now we let `button` do this directly, so instead of using `ariaHaspopup`, we have to use `aria-haspopup`

DanielMSchmidt:
@pierrebeitz are your concerns still valid?

pierrebeitz:
only that one: https://github.com/dcos-labs/ui-kit/pull/263#discussion_r256789674

it does not justify a block, but i'll only approve once we addressed that :)

packages/button/style.ts
mperrotti: It's not actually "wrong", it's just there's only inverse (dark mode) styles for the "secondary" and "standard" button appearances. So if `ButtonAppearances.Primary` gets passed, there's no styles to add

GeorgiSTodorov:
We don't have the button props here anymore.
Screenshot from 2019-05-10 12-19-20

GeorgiSTodorov:
@mperrotti , file conflicts.

GeorgiSTodorov:
Rebased.

mperrotti:
Adding "hold merge" to this PR because a button can now be a `<a>` or a `<button>`, which accept different props.

@pierrebeitz - You're a Typescript wizard. Do you have any ideas for how we can deal with this? :)

GeorgiSTodorov:
@mperrotti , I resolved a file conflict.

mperrotti:
I resolved the conflicts locally, but now that a button can also render an anchor tag I need to figure out someway to accept props for either a button _or_ an anchor tag 🤔

mperrotti:
I'm closing this PR because I don't think this is the right approach.

We need to either:
- Have a way to pick the right props based on which element gets rendered to the DOM
- Pick specific props that the Button components can accept, and don't pass incompatible props to `<a>` or `<button>`. E.g.: `<button>` should never be passed `href`

PR #264, opened 2018-12-13
DCOS-46141: remove padding from secondary button

mperrotti:
Lily and I discussed potential issues with aligning the text of a secondary button with other text on the page or in a modal.

We decided to change the secondary button to have no padding and a different `:focus` style.

### DIFF
--- a/packages/button/style.ts
+++ b/packages/button/style.ts
@@ -36,6 +36,8 @@ const filledButton = (
   return css`
     ${tintContent(contentColor)};
     background-color: ${baseColor};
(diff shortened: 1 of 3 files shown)


mperrotti:
This work came out of a design review session I had with Lily where I pointed out that the secondary button (used for dismissing) in the footer of a modal would not align with the content of the modal unless we manually set a negative margin.

We realized this problem will show up in more places and decided on modifying the `SecondaryButton` styles instead of creating a new component

PR #265, opened 2018-12-14
DCOS-44048: Lerna introduction

Poltergeist:
Unfortunately, this is only a WIP, I was running into a bunch of typescript compiler errors. Although I don't know how they slipped onto master.

- [ ] I haven't yet fixed all packages to actually build yet.
- [ ] There is one test failing I didn't investigate it yet.
- [ ] And I haven't checked storybook yet.


lerna.json
```diff
@@ -0,0 +1,5 @@
+{
+  "packages": ["packages/*"],
+  "npmClient": "npm",
+  "version": "0.0.0"
```
DanielMSchmidt: Doesn't this version refer to the currently published version? We have releases of the ui-kit, so this needs to be higher right? (see [wix/detox/lerna.json](https://github.com/wix/Detox/blob/master/lerna.json#L13) for reference)

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@dcos/ui-kit",
+  "name": "@dcos/ui-kit-mono-repo",
```
DanielMSchmidt: Does it need a name if it's private?

package.json
```diff
@@ -13,11 +13,11 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
+  "private": true,
   "scripts": {
     "build:storybook": "build-storybook",
     "prebuild:storybook": "npm run build:design-tokens",
-    "prebuild:design-tokens": "npm run clean:design-tokens",
-    "build:design-tokens": "node ./packages/design-tokens/scripts/build.js",
+    "build:design-tokens": "lerna run dist --scope=@dcos/ui-kit-design-tokens",
```
DanielMSchmidt: Why do we need this in the main package?

packages/appChrome/components/AppChrome.tsx
```diff
@@ -6,7 +6,7 @@ import {
   flexItem,
   textSize,
   flush
-} from "../../shared/styles/styleUtils";
+} from "@dcos/ui-kit-shared/dist/styles/styleUtils";
```
DanielMSchmidt: 💥 I thought we wanted to get rid of the sublevel imports for the styleUtils, did I misunderstand our intentions here?

packages/appChrome/package.json
```diff
@@ -0,0 +1,40 @@
+{
```
DanielMSchmidt: Is this file copy & paste or is it generated?

packages/appChrome/package.json
```diff
@@ -0,0 +1,40 @@
...
+  ],
+  "main": "dist/index",
+  "homepage": "https://github.com/dcos-labs/ui-kit#readme",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/dcos-labs/ui-kit.git"
+  },
+  "dependencies": {
+    "@dcos/ui-kit-clickable": "^0.0.0-dev+semantically-released",
```
DanielMSchmidt: 💥  I'm unsure if installs really work this way. I mean they are not really in that version, but in another one, aren't they? Why can't we have the right numbers here and have them updated by lerna?

packages/appChrome/stories/AppChrome.stories.tsx
```diff
@@ -30,7 +30,7 @@ import {
   blue,
   pink,
   purple
-} from "../../design-tokens/build/js/designTokens";
+} from "@dcos/ui-kit-shared/dist/build/js/designTokens";
```
DanielMSchmidt: 💥 Same as for the styleUtils, but didn't we want to get rid of the sublevel imports?

packages/shared/package.json
```diff
@@ -0,0 +1,38 @@
+{
```
DanielMSchmidt: Is the shared package really something we would want to publish? Has it worth on its own?

packages/shared/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/dcos-labs/ui-kit.git"
+  },
+  "dependencies": {
+    "@dcos/ui-kit-design-tokens": "^0.0.0-dev+semantically-released",
+    "@dcos/ui-kit-utilities": "^0.0.0-dev+semantically-released"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1",
```
DanielMSchmidt: Why only allow test runs from the root? Couldn't they all just run with jest?

packages/ui-kit-core/README.md
```diff
@@ -0,0 +1,11 @@
+# `ui-kit-core`
```
DanielMSchmidt: I would try to find another name, core seems a little unspecific, like a kitchen sink

packages/shared/package.json
Poltergeist: Yes, we need it in all the other packages. It is a shared central library. This has more value internally than externally but there is a huge benefit for this. We would need to include the shared files in all other packages which would result in code duplication.

packages/shared/package.json
Poltergeist: Would you rather run tests once or for each package? If we run it from root it is executing all tests, with the right scope

packages/ui-kit-core/README.md
Poltergeist: I am open for suggestions and happy to change this. This is basically the package which contains all the packages.
`ui-kit-general`. I did not want to go with `ui-kit` since it basically is a different package then `ui-kit`. I just picked something which kinda made sense and I choose not to be blocked on something which ain't important right now.

Poltergeist:
We decided to not go forward on this right now.

PR #267, opened 2018-12-20
[spike] Add a11y linting

mperrotti:
Added a subset of a11y linting rules from [Microsoft's tslint rules](https://github.com/Microsoft/tslint-microsoft-contrib), and enabled Storybook's a11y panel

### DIFF
--- a/tslint-a11y.js
+++ b/tslint-a11y.js
@@ -1,0 +1,21 @@
+const path = require("path");
+
+module.exports = {
+  rulesDirectory: path.dirname(require.resolve("tslint-microsoft-contrib")),
+  rules: {
+    "react-a11y-anchors": true,
+    "react-a11y-aria-unsupported-elements": true,
+    "react-a11y-event-has-role": true,
+    "react-a11y-image-button-has-alt": true,
+    "react-a11y-img-has-alt": true,
+    "react-a11y-lang": true,
+    "react-a11y-props": true,
+    "react-a11y-proptypes": true,
+    "react-a11y-role-has-required-aria-props": true,
+    "react-a11y-role-supports-aria-props": true,
+    "react-a11y-role": true,
+    "react-a11y-tabindex-no-positive": true,
+    "react-a11y-titles": false,
+    "react-a11y-meta": true,
+  },
+};
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -2,11 +2,13 @@ const req = require.context("./../packages", true, /.stories.tsx$/);
 
 import { configure, setAddon, addDecorator } from "@storybook/react";
 import { withKnobs } from "@storybook/addon-knobs/react";
+import { checkA11y } from "@storybook/addon-a11y";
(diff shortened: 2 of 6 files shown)


tslint-a11y.js
```diff
@@ -0,0 +1,21 @@
+const path = require("path");
+
+module.exports = {
+  rulesDirectory: path.dirname(require.resolve("tslint-microsoft-contrib")),
+  rules: {
```
mperrotti: I tried adding these to [@dcos/tslint-config](https://github.com/dcos-labs/tslint-config/blob/master/index.js) instead of keeping them only in ui-kit, but when I used `npm link` to test it, I got this error:

```
{ Error: Cannot find module 'typescript'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/mikep/tslint-config/node_modules/tslint-microsoft-contrib/reactA11yAnchorsRule.js:20:10)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3) code: 'MODULE_NOT_FOUND' }
```

tslint-a11y.js
```diff
@@ -0,0 +1,21 @@
...
+    "react-a11y-image-button-has-alt": true,
+    "react-a11y-img-has-alt": true,
+    "react-a11y-lang": true,
+    "react-a11y-props": true,
+    "react-a11y-proptypes": true,
+    "react-a11y-role-has-required-aria-props": true,
+    "react-a11y-role-supports-aria-props": true,
+    "react-a11y-role": true,
+    "react-a11y-tabindex-no-positive": true,
+    "react-a11y-titles": false,
```
mperrotti: I turned this off because it fails for the `<title>` in `<svg>`, which I think should be allowed to accept single-word titles. Also, this will only be useful for the title of an HTML document, which we aren't going to be using in any of ui-kit's packages

GeorgiSTodorov:
On this branch, when I open a story in AppChrome, it is automatically scrolled to the bottom. This doesn't happen on master.
peek 2019-01-03 11-00

mperrotti:
Thanks for catching, @GeorgiSTodorov. This is a known issue that was fixed in Storybook 4: https://github.com/storybooks/storybook/issues/4274

Do you think this is a big enough blocker that I should remove the changes that populate the Accessibility info panel from this PR?

GeorgiSTodorov:
@mperrotti , no, it is not a blocker.

PR #268, opened 2018-12-20
[spike] Clickable default focus style

mperrotti:
Instead of setting `outline: 0` for Clickable when it's focused, let the engineer using the component decide whether it's appropriate to hide the focus style using the prop `disableFocusOutline`

`outline: 0` should really only be set if you have custom focus styles on your component, or if there's some extenuating circumstance


GeorgiSTodorov:
Conflicting files.

GeorgiSTodorov:
Snapshot conflicts.

PR #269, opened 2018-12-20
[spike] `fontSizes` keys match similar utils

mperrotti:
`fontSize` style util now takes a size param that matches `margin` and `padding`

Before:
`fontSize("default")` vs `padding("all", "m")`

After:
`fontSize("m")` vs `padding("all", "m")`


GeorgiSTodorov:
Snapshot conflicts.

PR #270, opened 2019-01-03
[spike] Improve screenreader UX for Sidebar

mperrotti:
Tested before and after using VoiceOver and ChromeVox extension


GeorgiSTodorov:
Tested with ChromeVox, but couldn't hear any difference.

mperrotti:
@GeorgiSTodorov

Before:
- Screenreader would not consider the expandable menu items as list items because they were a `<div>`, not a `<li>`. So when a user focused on a link in the "Resources" section of the sidebar, it would read that there were only 2 list items when there are actually 3
- Screenreader would read each sidebar item as a "button", not a "link"

PR #271, opened 2019-01-07
DCOS-46484: Table columns can size themselves based on column content

mperrotti:
The `width` prop is now optional. When the `width` prop is excluded, the Column width will be as wide as the widest Cell in the Column.

I've also added `minWidth` and `maxWidth` props to Column.

**Screenshot:**
screen shot 2019-01-07 at 1 02 01 pm
Note: the "Very Long" column has a `maxWidth` of 300 applied


packages/table/components/Column.tsx
```diff
@@ -18,7 +18,15 @@ export interface ColumnProps {
   /**
    * width should return a column width in pixels.
    */
-  width: (args: WidthArgs) => number;
+  width?: (args: WidthArgs) => number;
```
juliangieseke: regarding `minWidth` and `maxWidth`, what do you think about passing in `caluclatedWith` with `args` here instead and let the function do it? It feels a bit more streamlined if one function does all instead of having a function and 2 config thingies 🤔

mperrotti: That's possible. I felt like it was more straightforward to have them as separate props, but I'm curious to hear what other reviewers think.

Happy to make that change if even one other person thinks your suggestion is a better developer experience :)

packages/table/components/Table.tsx
```diff
@@ -152,16 +171,35 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
     const { key, style, column } = args;
-    const className = css(headerCss, {
-      ...style
-    });
     return (
-      <div className={className} key={key}>
+      <div
+        className={headerCss}
+        style={{
+          ...style
```
nLight: 💅 No need in spreading `style={style}`, am I missing something?

packages/table/components/Table.tsx
```diff
@@ -186,19 +224,19 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
 
     return (
       /* tslint:disable:react-a11y-event-has-role */
       <div
-        className={cx(className, {
+        className={cx(cellCss, {
           [rowHoverCss]: rowIndex === this.state.hoveredRowIndex
         })}
+        style={{
+          ...style
```
nLight: same

packages/table/components/Table.tsx
```diff
@@ -45,6 +51,12 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
     }
   );
 
+  public _cache = new CellMeasurerCache({
```
nLight: should `cache` be public?

nLight: maybe also more specific name since it's not a general purpose cache, is it?

packages/table/components/Table.tsx
nLight: Why is it better than `emotion.css`?

packages/table/components/Table.tsx
```diff
@@ -45,6 +51,12 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
+    defaultHeight: ROW_HEIGHT,
+    defaultWidth: 150,
+    fixedHeight: true
+  });
+
   public getColumnSizes = memoizeOne(
```
nLight: It looks like we cache it twice with `memoizeOne` and with `this._cache`. Do they serve different purposes?

packages/table/components/Table.tsx
mperrotti: @nLight - if we don't put it in the style attribute, react-virtualized sets `width: auto` in the style attribute which overrides whatever we add using `emotion`

packages/table/components/Table.tsx
mperrotti: What specifically do we cache twice?

nLight: So the function itself is memoized with `memoizeOne` and it uses `this._cache` internally so it looks like we're memoizing the cache, am I missing something?

mperrotti: Ahh I understand. `this._cache` (changed to `this.cellMeasureCache`) is what `react-virtualized` uses to calculate the width dynamically, but we need to get the value it returns in `getColumnSizes`

PR #272, opened 2019-01-09
DCOS-45656: generate icon sprite from folder of SVGs

mperrotti:
Running `npm run build:svg` should build two SVG sprites: one for "system" icons and one for "product" icons.

screen shot 2019-01-09 at 9 04 30 pm

Next step is to render these icons from the sprite with an Icon component: [DCOS-45657](https://jira.mesosphere.com/browse/DCOS-45657)


mperrotti:
@GeorgiSTodorov - yes

mperrotti:
`svgstore` moves `<defs>` outside of each `<symbol>`. This is a problem for product icons, which use defs for gradients. I'm trying to work around this, but may need to generate the sprite some other way

.gitignore
```diff
@@ -13,3 +13,4 @@ tmp/
 .env
 coverage
 storybook-static/
+packages/icons/dist/
```
mperrotti: For some reason, the files in this particular `dist/` dir weren't being ignored.

Reviewer: try removing this line on your machine, run `npm run build:svg`, and see if git ignores what gets created

GeorgiSTodorov: It doesn't for me.
screenshot from 2019-01-11 09-16-02

nLight:
SVG seems to be valid according to W3C validator, but I can't see anything in Inkscape nor in photoshop.
Regarding the `dist` folder. I wonder if it should actually go inside the top level `/dist` instead?

mperrotti:
@GeorgiSTodorov and @nLight - SVG sprites aren't going to be visible in vector graphics GUIs or the browser. However, they _do_ show when I render them an icon like this: `<svg><use xlink:href="#icon-id" /></svg>`

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,100 @@
...
+
+        prev[nameToPascal] = curr;
+        return prev;
+      }, {});
+
+  const enumString = Object.keys(svgNamesObj).reduce((prev, curr) =>
+    prev += `${curr} = "${idPrefix}-${svgNamesObj[curr]}",`,
+  "")
+    .split(",")
+    .filter(entry => entry != "")
```
nLight: why would entry be `""`?

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,100 @@
...
+    .join(",\n  ");
```
nLight: Can't we reduce with `",\n  "`?

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,100 @@
...
+      throw err;
+    }
+
+    svgo.optimize(data, { path: spritePath }).then(result => {    
+      writeFileSync(spritePath, result.data);
+    });
+  });
+};
+
+const writeEnum = (srcDir, iconSetName, idPrefix) => {
```
nLight: I wonder if @DanielMSchmidt would have any idea on how we could improve code generator?

mperrotti:
> Regarding the dist folder. I wonder if it should actually go inside the top level /dist instead?

@nLight - since we're not going with Lerna, that could work. I'd also update design tokens to export to top-level `dist/` for consistency

mperrotti:
@nLight - I made it export to the top-level dist, but I realized we would want to create something in `packages/icons/dist/` as well for importing the icon sprite in Storybook.

GeorgiSTodorov:
@mperrotti , I can't see them with `<svg><use xlink:href="#icon-id" /></svg>`, but I am not familiar with svg, so I could be doing something wrong. Do we need to import them. I think adding a story to show how these icons look like is a good idea.

packages/icons/scripts/build.js
DanielMSchmidt: I think we could go the other route and get rid of the quotes problem if we did something like this:


```js
const t = require("babel-types");
const generate = require("@babel/generator");

const ast = t.tsEnumDeclaration(
  t.identifier(`${iconSetName.replace(/^\w/, c => c.toUpperCase())}Icons`),
  Object.keys(svgNamesObj).map(curr =>
    t.tsEnumMember(`${idPrefix}-${svgNamesObj[curr]}"`)
  )
);
const string = generate(ast);
```

mperrotti:
@GeorgiSTodorov - the sprite needs to be rendered into Storybook. You can see everything working in Storybook in this branch: `wip/DCOS-45657-add-icon-component`

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,102 @@
...
+    t.identifier(`${iconSetName.replace(/^\w/, c => c.toUpperCase())}Icons`),
+    Object.keys(svgNamesObj).map(svgName =>
+      t.tSEnumMember(t.identifier(svgName), t.stringLiteral(`${idPrefix}-${svgNamesObj[svgName]}`))
+    )
+  );
+  const { code } = generate(ast);
+
+  writeFileSync(
+    path.join(buildDirPath, `${iconSetName}-icons-enum.ts`),
+    // when generate parses the AST, it adds a trailing comma that needs
```
nLight: I tried to find an option on the `generator` to disable that but couldn't

PR #275, opened 2019-01-11
fix: Convert React to peer dependency

brandonc:
It's unclear whether this will fix an underlying issue or not, but it would be good not to bundle two versions of react nonetheless.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,15 +44,17 @@
     "emotion-theming": "9.2.9",
     "focus-trap-react": "4.0.1",
     "memoize-one": "4.0.2",
-    "react": "16.5.2",
     "react-click-outside": "3.0.1",
     "react-delegate-component": "1.0.0",
-    "react-dom": "16.5.2",
     "react-emotion": "9.2.12",
     "react-toggled": "1.2.7",
     "react-transition-group": "2.5.0",
     "react-virtualized": "9.20.1",
     "semantic-release": "15.9.6"
+  },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,16 +44,18 @@
...
     "react-delegate-component": "1.0.0",
-    "react-dom": "16.5.2",
     "react-emotion": "9.2.12",
     "react-toggled": "1.2.7",
     "react-transition-group": "2.5.0",
     "react-virtualized": "9.20.1",
     "semantic-release": "15.9.6"
   },
+  "peerDependencies": {
+    "react": ">=15.4.1 <=16.7.0",
```
mperrotti: `16.7.0` will still work for `16.7.0-alpha.0`, right?

natmegs: Could we just say <= 16.7?

brandonc: I think 16.7.0 is released anyway

natmegs: I'm not sure Hooks are available outside of non-alpha versions?

natmegs: They're not listed in the 16.7.0 changelog

mperrotti: I assume npm considers `16.7.0-alpha.0` as matching `>=15.4.1 <=16.7.0`. If not, we could do `>=15.4.1 <=16.7.0 || 16.7.0-alpha.0`

brandonc: I'll change it

PR #279, opened 2019-01-18
[spike] Replaced SVGs with Icon component

mperrotti:
No exciting screenshot to show.

While I was doing the replacement, I also updated components that take icons as props to accept the the product and system icon enums. I made sure this change is backwards-compatible with how we did it before we had an icon component.

For example, if you want to have a button with an icon you can do it two ways:
```
new way:
<Button iconStart={SystemIcons.ArrowDown} />

old way:
<Button iconStart={<Icon shape={SystemIcons.ArrowDown} size={iconSizeXs} color="inherit"} />
```

Eventually, I'd like to stop supporting the "old way", but I didn't want this spike PR to break consumer apps.


packages/appChrome/components/SidebarSubMenu.tsx
```diff
@@ -25,13 +26,13 @@ class SidebarSubMenu extends React.PureComponent<SidebarSubMenuProps, {}> {
   public getSubItemList(
     items: Array<React.ReactElement<SidebarSubMenuItemProps>>
   ) {
-    const iconWidth = this.props.iconWidth ? this.props.iconWidth : 0;
+    const iconWidth = this.props.iconWidth ? this.props.iconWidth : iconSizeS;
```
GeorgiSTodorov: How about `const iconWidth = this.props.iconWidth || iconSizeS;`?

packages/icons/src/icon-system/triangle-down.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
+    <path d="M2.9 6.1L8 11.2l5.1-5.1H2.9z"/>
+</svg>
```
GeorgiSTodorov: Are we adding this file manually?

GeorgiSTodorov:
File conflicts.

PR #282, opened 2019-01-18
DCOS_OSS-4364: allow column width to fill available space

mperrotti:
Adding the `growToFill` prop to a column allows that column to grow until the table fills it's parent. `maxWidth`, `minWidth`, and `width` props will still work in conjunction with this even though I can't think of a realistic use case for combining those with `growToFill` ¯\_(ツ)_/¯

In this example, the "Role" and "City" columns have `growToFill`:
screen shot 2019-01-18 at 5 13 54 pm

### DIFF
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -42,6 +42,47 @@ const ROW_HEIGHT = 35;
 
 const DEFAULT_WIDTH = 1024;
 const DEFAULT_HEIGHT = 768;
+
+export const clampWidth = (
+  baseWidth: number,
+  minWidth: number,
+  maxWidth: number
+): number => Math.min(Math.max(baseWidth, minWidth), maxWidth);
+
+export const fillColumns = (
+  columns: Array<React.ReactElement<ColumnProps>>,
+  colWidths: number[],
+  width: number
+): number[] => {
+  let copiedColWidths = [...colWidths];
+  const widthToFill =
+    width - copiedColWidths.reduce((acc, curr) => acc + curr, 0);
+  const fillWidthColsIdxArr: number[] = columns
+    .filter(child => child.props.growToFill)
+    .map(col => columns.indexOf(col));
+
+  fillWidthColsIdxArr.forEach(colIdx => {
+    const oldWidth = copiedColWidths[colIdx];
+    const newWidth = widthToFill / fillWidthColsIdxArr.length + oldWidth;
+
+    if (newWidth <= oldWidth) {
+      return;
+    }
+
+    copiedColWidths[colIdx] = clampWidth(
+      newWidth,
+      columns[colIdx].props.minWidth || 0,
+      columns[colIdx].props.maxWidth || width
+    );
+  });
+
+  if (copiedColWidths.reduce((acc, curr) => acc + curr, 0) < width) {
+    return fillColumns(columns, copiedColWidths, width);
+  } else {
+    return copiedColWidths;
+  }
+};
+
 export class Table<T> extends React.PureComponent<TableProps, TableState> {
   public multiGridRef: { recomputeGridSize?: any } = {};
 
@@ -56,9 +97,10 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
       children: Array<React.ReactElement<ColumnProps>>,
       width: number
     ): number[] => {
-      const totalColumns: number = children.length;
       let remainingWidth: number = width;
-      return children.map((child, currentIndex) => {
+
+      const totalColumns: number = children.length;
+      const colWidths = children.map((child, currentIndex) => {
         const calculatedWidth = child.props.width
           ? child.props.width({
               width,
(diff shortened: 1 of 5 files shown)


packages/table/components/Table.tsx
```diff
@@ -56,9 +56,17 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
       children: Array<React.ReactElement<ColumnProps>>,
       width: number
     ): number[] => {
-      const totalColumns: number = children.length;
       let remainingWidth: number = width;
-      return children.map((child, currentIndex) => {
+      const clampWidth = (
```
nLight: `getColumnSizes` has to be refactored, it grows out of bounds. There are several functions inside that can be extracted and the name is misleading, we then pass the result as `<MultiGrid columnWidth={getColumnSize} />`.
While name change is out of scope I would like to ask to consider refactoring the implementation.

DanielMSchmidt:
I had this error:

```
[tsl] ERROR in /Users/danielschmidt/projects/ui-kit/packages/icon/components/Icon.tsx(13,29)
      TS2307: Cannot find module '../../icons/dist/system-icons-enum'.
```

What am I doing wrong?

GeorgiSTodorov:
@DanielMSchmidt , have you ran `npm run build:svg` before?
https://github.com/dcos-labs/ui-kit/pull/272

DanielMSchmidt:
No, probably not. Maybe we can add them as `pre` hooks, it's really hard to remember what to do in which order

DanielMSchmidt:
I always run `npm run storybook`

mperrotti:
That's still strange. The icon build runs in `prebuild:storybook` too 🤔

packages/table/components/Table.tsx
```diff
@@ -69,13 +77,21 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
+          calculatedWidth,
+          child.props.minWidth || 0,
           child.props.maxWidth || width
         );
         remainingWidth -= clampedWidth;
         return clampedWidth;
       });
+
+      if (children.filter(child => child.props.growToFill).length) {
+        this.fillTableWidth(colWidths, remainingWidth, width);
```
nLight: Let's please not mutate `colWidths`

packages/table/components/Table.tsx
```diff
@@ -122,6 +138,47 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
 
+  private fillTableWidth(
+    colWidths: number[],
+    remainingWidth: number,
+    width: number
+  ) {
+    let filledWidth;
+    const totalWidth = colWidths.reduce((acc, curr) => acc + curr, 0);
+    const widthToFill = width - totalWidth;
+    const columns = React.Children.toArray(this.props.children) as Array<
```
nLight: Let's pass `children` https://github.com/dcos-labs/ui-kit/pull/282/files#diff-f82f764378c02f64dad7e8c8b1910116R63 as an argument to the function, so that we could extract the function from the class and unit-test it.

packages/table/components/Table.tsx
```diff
@@ -122,6 +138,47 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
+      React.ReactElement<ColumnProps>
+    >;
+    const fillWidthColsIdxArr: number[] = columns
+      .filter(child => child.props.growToFill)
+      .map(col => columns.indexOf(col));
+    const fillColumns = (colsIdxArr, widthToFill) => {
```
nLight: let's please extract the inline function and unit-test separately

nLight: so that imaginary utils module would consist of two functions `fillTableWidth` and `fillColumns`

packages/table/components/Table.tsx
```diff
@@ -122,6 +138,47 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
+      }
+      colsIdxArr.forEach(colIdx => {
+        const oldWidth = colWidths[colIdx];
+        const newWidth = widthToFill / fillWidthColsIdxArr.length + oldWidth;
+
+        if (newWidth <= oldWidth) {
+          return;
+        }
+
+        colWidths[colIdx] = clampWidth(
```
nLight: we might want to move the function to the same utility module as well. That may not be a separate file. We could totally have those function in the Table.tsx and exporting them so that they could be tested.

GeorgiSTodorov:
Slavena noticed a problem in the nodes table in `dcos-ui`, and the same thing is happening here. When the window is small, the table is moving when hovering some rows.
peek 2019-01-29 10-39

GeorgiSTodorov:
@mperrotti , I already created a jira - https://jira.mesosphere.com/browse/DCOS-47795 .

GeorgiSTodorov:
It's happening in the nodes table, but it's also happening in this table, so if it's a quick fix, I would vote for fixing it here before we merge it.

nLight:
@GeorgiSTodorov @mperrotti let's please not scope creep 🙏  and fix that separately. Doesn't seem to be a critical issue, does it?

mperrotti:
@GeorgiSTodorov and @nLight:
Upon brief investigation, this _looks_ like it could be as simple as doing this check in `componentDidMount` will fix the issue:

```
  public componentDidUpdate(prevProps) {
    if (this.props === prevProps) {
      return;
    }

    this.cellMeasureCache.clearAll();
    this.updateGridSize();
  }
```

The table is "jumping" because we change state to update which row is hovered, so `componentDidUpdate` gets triggered and recalculates the table layout.

I haven't tested in dcos-ui, so I'm not sure if this will truly fix the issue

mperrotti:
Either way, I agree with Dima that we should resolve DCOS-47795 in a separate PR

PR #283, opened 2019-01-24
DCOS-47014: Select input component

mperrotti:
Added a component for a `<select>` input

screen shot 2019-01-23 at 5 49 43 pm


packages/selectInput/README.md
```diff
@@ -0,0 +1,3 @@
+# SelectInput Name
```
mperrotti: @lsamimi - I couldn't find any existing documentation on select inputs. Maybe we should collaborate on some?

packages/selectInput/components/SelectInput.tsx
```diff
@@ -0,0 +1,181 @@
...
+            </span>
+            {getHintContent}
+            {appearance !== SelectInputAppearance.Error && getValidationErrors}
+          </div>
+        )}
+      </FormFieldWrapper>
+    );
+  }
+
+  private handleFocus(e) {
```
mperrotti: I really don't like that this component has to have state just for focus styles...

I may look into detecting `:focus-within` support, and if the browser doesn't have it, don't set `outline: 0` on the input. I won't lose any sleep if somebody using Edge/IE doesn't get a pretty focus state :x

mperrotti:
There's some work here that could be used in `TextInput`:
- Use `FormFieldWrapper` component instead of making the component handle everything
- Share`formStyles.ts` styles between `SelectInput` component _and_ `TextInput`

I'd like the opinion of reviewers: should I add that work in this PR, or open a new one on the next spike day? Or do you disagree with sharing this code altogether?

GeorgiSTodorov:
Doing these two tasks in this PR, or as a follow-up PR in this sprint sounds like a good idea to me.

mperrotti:
Hold merge until I get the TextInput changes in too (see this comment: https://github.com/dcos-labs/ui-kit/pull/283#issuecomment-457016322)

packages/selectInput/components/SelectInput.tsx
```diff
@@ -0,0 +1,181 @@
...
+  padding,
+  visuallyHidden,
+  display
+} from "../../shared/styles/styleUtils";
+import Icon from "../../icon/components/Icon";
+import { SystemIcons } from "../../icons/dist/system-icons-enum";
+import { iconSizeXs } from "../../design-tokens/build/js/designTokens";
+import FormFieldWrapper from "../../shared/components/FormFieldWrapper";
+
+export enum SelectInputAppearance {
```
natmegs: Can we also share this with `TextInput`?

mperrotti: Good idea. I just have to figure out the right place to put it 🤔

GeorgiSTodorov:
File conflicts.

packages/selectInput/components/SelectInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+        {({ getValidationErrors, isValid, getHintContent, describedByIds }) => (
+          <div>
+            <label
+              className={cx(getLabelStyle(hasError), {
+                [visuallyHidden]: !showInputLabel
+              })}
+              htmlFor={id}
+            >
+              {inputLabel}
+              {this.props.required ? (
```
nLight: we could add `required` to the destructing `const {} = this.props`;

packages/selectInput/README.md
lsamimi: @mperrotti yes, I'll set up a meeting with you and we can write some doc on this.

packages/selectInput/components/SelectInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+    if (this.props.disabled) {
+      return "disabled";
+    }
+    if (this.state.hasFocus) {
+      return `${this.props.appearance}-focus`;
+    }
+    return this.props.appearance;
+  }
+
+  // private getDescribedBy(hintContent, errors) {
```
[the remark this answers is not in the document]
mperrotti: 😅 how embarassing

mperrotti:
@GeorgiSTodorov - I'm going based off of the design team's mocks for rendering hint and error text. This is also consistent with how we do it for the `TextInput`, so I think we should keep it as-is

PR #284, opened 2019-01-26
DCOS-47380: Checkbox component

mperrotti:
Added a checkbox component and a component for rendering a fieldset with a list of checkbox components.

**Checkbox component**
screen shot 2019-01-25 at 8 08 51 pm

**ToggleInputList component**
screen shot 2019-01-25 at 8 09 21 pm

### DIFF
--- a/packages/checkboxInput/stories/CheckboxInput.stories.tsx
+++ b/packages/checkboxInput/stories/CheckboxInput.stories.tsx
@@ -1,0 +1,110 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import CheckboxStoryHelper from "./helpers/CheckboxStoryHelper";
+import { ToggleInputAppearance } from "../../toggleInput/components/ToggleInput";
+import { css } from "emotion";
+
+const readme = require("../README.md");
+
+const checkboxGrid = css`
+  display: grid;
+  grid-gap: 1em;
+  grid-auto-flow: column;
+
+  > div:nth-child(n + 5) {
+    grid-column: 2 / 3;
+  }
+`;
+
+storiesOf("Forms/CheckboxInput", module)
+  .addDecorator(withReadme([readme]))
+  .addWithInfo("default", () => (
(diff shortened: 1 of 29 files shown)


packages/toggleInput/components/ToggleInput.tsx
```diff
@@ -0,0 +1,84 @@
+import * as React from "react";
```
mperrotti: This component would be excessive if we were just using it for the Checkbox, but this will make the radio input and toggle switch input components go much faster.

mperrotti:
Still need to write docs, but everything else is ready for review 👍

packages/checkboxInput/stories/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import CheckboxStoryHelper from "./helpers/CheckboxStoryHelper";
+import { ToggleInputAppearance } from "../../toggleInput/components/ToggleInput";
+import { css } from "emotion";
```
GeorgiSTodorov: This import should be above the local imports.

mperrotti:
@GeorgiSTodorov - I don't think that behavior is intentional. I'm comfortable moving forward without it

packages/toggleInput/components/ToggleInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+  padding,
+  visuallyHidden,
+  display,
+  tintContent,
+  tintContentSecondary
+} from "../../shared/styles/styleUtils";
+import { error, success } from "../../design-tokens/build/js/designTokens";
+import { CheckboxInputProps } from "../../checkboxInput/components/CheckboxInput";
+
+export enum ToggleInputAppearance {
```
natmegs: Will this use the shared `InputAppearance` enum from the select component PR once merged?

mperrotti: of course :)

PR #286, opened 2019-01-29
CheckboxInput followup: added hint text to toggle input

mperrotti:
I asked the design team about this pattern:
screen shot 2019-01-29 at 6 43 38 pm

They said clicking the hint text should not toggle the input. I made it possible to pass `hintContent` to an individual `CheckboxInput`, and automatically render the correct styling and functionality.
This might have been annoying to do manually in a consumer application, and prone to layout bugs if we changed the size of the toggle input.

### DIFF
--- a/packages/checkboxInput/components/CheckboxInput.tsx
+++ b/packages/checkboxInput/components/CheckboxInput.tsx
@@ -12,6 +12,7 @@ import { SystemIcons } from "../../icons/dist/system-icons-enum";
 import { iconSizeXs, white } from "../../design-tokens/build/js/designTokens";
 import { ToggleInputProps } from "../../toggleInput/components/ToggleInput";
 import { InputAppearance } from "../../shared/types/inputAppearance";
+import FormFieldWrapper from "../../shared/components/FormFieldWrapper";
 
 export interface CheckboxInputState {
   hasFocus: boolean;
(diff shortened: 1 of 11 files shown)


GeorgiSTodorov:
File conflicts.

GeorgiSTodorov:
The hint for CheckboxInput looks different than other hints (for example SelectInput) because of the bigger padding. Is this on purpose?
screenshot from 2019-02-04 17-58-30
screenshot from 2019-02-04 17-58-53

mperrotti:
Good catch!
The top padding is actually the same (4px). The extra space appears to be caused by the line-height of the div that wraps the hint content.

GeorgiSTodorov:
@mperrotti , it is still the same for me.

mperrotti:
Resolved conflicts.

@GeorgiSTodorov - the extra space between the text and the checkbox is caused by the height of the checkbox. Here are screenshots showing the hint text with a top border to illustrate:
Screen Shot 2019-03-28 at 11 48 48 AM
Screen Shot 2019-03-28 at 11 48 39 AM

GeorgiSTodorov:
@mperrotti , I think I meant the left padding of the text.

mperrotti:
Ohhh. That's intentional to align with the text instead of the checkbox :)

GeorgiSTodorov:
Clicking on the invisible label triggers a click on the checkbox. Is this intentional?
Peek 2019-08-16 08-47

mperrotti:
Good catch on the invisible label. I'll fix that

PR #287, opened 2019-01-30
AppChrome fix: content/sidebar height

mperrotti:
Before, the `height: 100%` on the sidebar and the main content area was setting the elements to be the full height of their parent.

Now, they only fill the available space after accounting for headerbar height because the headerbar and the div that wraps the sidebar and main content area are in a flex container.


AmrAbdelrazik:
@mperrotti screenshots please?

PR #291, opened 2019-02-04
DCOS-41571: added a table with checkbox-selectable rows

mperrotti:
This PR introduces the `CheckboxTable` component, and updates the Table documentation with relevant information about taking action on a set of rows.

This PR also introduces an indeterminate state to the `CheckboxInput` component for the checkbox that appears in the header of the checkbox column

checkboxtabledemo

### DIFF
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -31,6 +31,14 @@ export interface TableProps {
   children:
     | Array<React.ReactElement<ColumnProps>>
     | React.ReactElement<ColumnProps>;
+  /**
+   * How many columns do not scroll
+   */
+  fixedColumnCount?: number;
+  /**
+   * Selected row data
+   */
+  selectedRows?: object[];
 }
 
 export interface TableState {
@@ -55,15 +63,18 @@ export const fillColumns = (
   width: number
 ): number[] => {
   let copiedColWidths = [...colWidths];
-  const widthToFill =
-    width - copiedColWidths.reduce((acc, curr) => acc + curr, 0);
+  const filledArea = columns
+    .filter(child => !child.props.growToFill)
+    .map(col => copiedColWidths[columns.indexOf(col)])
+    .reduce((acc, curr) => acc + curr, 0);
+  const widthToFill = width - filledArea;
   const fillWidthColsIdxArr: number[] = columns
     .filter(child => child.props.growToFill)
     .map(col => columns.indexOf(col));
 
   fillWidthColsIdxArr.forEach(colIdx => {
     const oldWidth = copiedColWidths[colIdx];
-    const newWidth = widthToFill / fillWidthColsIdxArr.length + oldWidth;
+    const newWidth = widthToFill / fillWidthColsIdxArr.length;
 
     if (newWidth <= oldWidth) {
       return;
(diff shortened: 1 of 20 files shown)


AmrAbdelrazik:
@mperrotti nice work, I am not seeing a big difference in the state between the partial selection and full selection (for the table checkbox above), anyway we can try to tweak that. (for example, simpler shade of purple or something)

GeorgiSTodorov:
Looks good.
Some observations:
1. I think "Delete selected rows" is too verbose, to me "Delete selected" or just "Delete" looks better.
2. Should Group action button disappear when no row is selected the same way the delete button disappears?
3. When I delete all rows, I can't un-check the header checkbox anymore, I think this is a bug.
peek 2019-02-05 10-44

4. Would this work if we could add more rows?

mperrotti:
@AmrAbdelrazik - the GIF just has much lower contrast

mperrotti:
Could you share an example for how you'd prefer to render a table with checkbox-selectable rows?

I completely agree that we don't want to end up with components that are too specific, but I don't think the CheckboxTable component is too specific. Having a CheckboxTable feels like a simpler API to me. Engineers don't have to remember to set `fixedColumnCount` to 2, and the Table doesn't have an `onChange` prop that rarely gets used.

I'm being careful with when we create a new component vs. when we should just use props. I promise we won't end up with annoyingly specific components 😇

DanielMSchmidt:
I would have thought of sth like this, but I might be oversimplifying somehting:


```js
import { Table, CheckBoxColumn } from "ui-kit";


function MyTable({data}){
  return (
    <Table data={data}>
      <CheckBoxColumn value={item => item.checked} onToggle={item => item.checked = true} />
      <MyOtherColumns />
    </Table>
  );
}
```

mperrotti:
Ah ok, I understand. I like the idea of having a new column component instead. Let me give that a try...

GeorgiSTodorov:
I think it looks good. Some more observations:
1. In dcos-ui the checkbox tables that we use allow us to select and unselect the header checkbox even when there are no rows (see for example secrets table). Here it doesn't allow us to select it if there are no rows. Which is the preferred behavior?
2. If we select the header checkbox when there are no rows, the header cells move
peek 2019-02-06 10-59
3. I think the global action button would look better if it was on the right so it doesn't jump around as we select and unselect rows. What does the design document say?
4. The focus for the header cells in the selectable table is unintended I think
Firefox
screenshot from 2019-02-06 10-55-31
Chrome
screenshot from 2019-02-06 10-56-54

mperrotti:
> In dcos-ui the checkbox tables that we use allow us to select and deselect the header checkbox even when there are no rows (see for example secrets table). Here it doesn't allow us to select it if there are no rows. Which is the preferred behavior?

I think it feels weird to be able to select a checkbox that does nothing. I think when there are no selectable rows, I'll disable the checkbox

> If we select the header checkbox when there are no rows, the header cells move

That's no good, I'll investigate

> I think the global action button would look better if it was on the right so it doesn't jump around as we select and deselect rows. What does the design document say?

The design documentation says "When group actions appear, the global actions will sit left of the group action."

> The focus for the header cells in the sortable table is unintended I think

No, that's not intentional. It's not technically part of this ticket, but it should be an easy enough fix to get into this PR

packages/table/components/Table.tsx
```diff
@@ -55,20 +63,24 @@ export const fillColumns = (
...
+    .reduce((acc, curr) => acc + curr, 0);
+  const widthToFill = width - filledArea;
   const fillWidthColsIdxArr: number[] = columns
     .filter(child => child.props.growToFill)
     .map(col => columns.indexOf(col));
 
   fillWidthColsIdxArr.forEach(colIdx => {
     const oldWidth = copiedColWidths[colIdx];
-    const newWidth = widthToFill / fillWidthColsIdxArr.length + oldWidth;
+    const newWidth = widthToFill / fillWidthColsIdxArr.length;
```
mperrotti: I made some changes to `fillColumnWidth` to fix a bug where after deleting a row, if you hovered over the table or triggered a re-render, the column widths would "jump". I found this while testing Georgi's requested changes.

Before:
coljump_before

After:
coljump_after

mperrotti:
@DanielMSchmidt - I still can't figure out a good way to do this and maintain the functionality of the checkbox in the header that toggles all selections.

AmrAbdelrazik:
Nice work everyone, let's get this rolled out in the UI!

PR #293, opened 2019-02-06
chore: update peerDeps to include React 16.8

mperrotti:


nLight:
Do we want to switch to 16.8 in devDeps as well?

PR #294, opened 2019-02-07
DCOS-46403: allow table row heights other than default

mperrotti:
Before this, table rows could only be the default height, 35px

Table rows that are 60px high:
screen shot 2019-02-07 at 3 22 08 pm

### DIFF
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -19,6 +19,7 @@ import {
 
 import { ColumnProps, Column } from "./Column";
 import memoizeOne from "memoize-one";
+import { vAlignChildren } from "../../shared/styles/styleUtils";
 
 export interface TableProps {
   /**
@@ -39,6 +40,10 @@ export interface TableProps {
    * Selected row data
    */
   selectedRows?: object[];
+  /**
+   * Set the row height
+   */
+  rowHeight?: number;
 }
 
 export interface TableState {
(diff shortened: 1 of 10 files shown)


jaxesn:
Where did the need for this come from?

mperrotti:
@jaxesn - when you filter the service table, the "name" column has 2 lines:
image

I suggested an alternative here: https://github.com/dcos/dcos-ui/pull/3583#issuecomment-461585692

AmrAbdelrazik:
@mperrotti , @jaxesn. I would rather not show additional column and show the full length of the service URL in the name

FOr example instead of showing 
Kibanaproxy
Services/Logs

I would rather show
Services/logs/Kibanaproxy

image.

@izharbuendia what do you think?

AmrAbdelrazik:
@jaxesn @mperrotti , chatted with @izharbuendia and we will keep the current behavior of expanding the row to include the parent link.

packages/table/components/Table.tsx
```diff
@@ -39,6 +40,10 @@ export interface TableProps {
    * Selected row data
    */
   selectedRows?: object[];
+  /**
+   * Set the row height
+   */
+  rowHeight?: number;
```
DanielMSchmidt: Random thought, but we could make bad practices more screaming 🤷‍♂️ 

```suggestion
  dangerouslyOverwriteRowHeight?: number;
```

GeorgiSTodorov: Is it really that dangerous? :D

mperrotti: I wouldn't say it's "dangerous" as in it will cause a bug or some unexpected behavior. I just worry about the design tradeoffs of using taller row heights to cram too much information into a single cell

PR #296, opened 2019-02-08
CheckboxTable fix: don't compare row data using objects

mperrotti:


packages/table/components/CheckboxTable.tsx
```diff
@@ -7,13 +7,21 @@ export interface CheckboxTableProps extends TableProps {
...
+   */
+  uniqueKey: string;
 }
 
 class CheckboxTable extends React.PureComponent<CheckboxTableProps> {
   constructor(props) {
     super(props);
 
+    this.state = {
+      selected: props.selectedRows || []
```
nLight: I don't see we're using this piece of state

mperrotti: We're not, I forgot to delete it after testing something

packages/table/components/CheckboxTable.tsx
```diff
@@ -7,7 +7,11 @@ export interface CheckboxTableProps extends TableProps {
   /**
    * Callback for when a user makes a selection. Passes selected row data as a parameter
    */
-  onChange?: (selectedRows: object[]) => void;
+  onChange?: (selectedRows: string[]) => void;
+  /**
+   * The unique key in each row's data object
```
nLight: Is there a strong reason to have `selectedRows` on the `TableProps` and not on `CheckboxTableProps`?

nLight: Why `Table` should be concerned about `selectedRows`?

packages/table/components/CheckboxTable.tsx
```diff
@@ -52,7 +58,7 @@ class CheckboxTable extends React.PureComponent<CheckboxTableProps> {
           id={rowIndex.toString()}
           inputLabel={`Toggle row ${rowIndex}`}
           showInputLabel={false}
-          checked={selectedRows.includes(rowData)}
+          checked={selectedRows.includes(rowId)}
```
nLight: Would an object also work so we could replace `includes` with just `selectedRows[rowId]` and `filter` with `delete selectedRows[rowId]`. Is there any specific reason we would like having them as array?

packages/table/components/CheckboxTable.tsx
```diff
@@ -62,11 +68,14 @@ class CheckboxTable extends React.PureComponent<CheckboxTableProps> {
   private getHeaderCheckbox() {
     const selectedRows = this.props.selectedRows || [];
     let isSelected = false;
```
nLight: I know I'm late to the party 😅 could this `isSelected` be a part of `state` instead of being local mutable variable?

mperrotti: Definitely. I was stubbornly trying to avoid state

packages/table/components/CheckboxTable.tsx
mperrotti: I don't see a reason why not. An array just felt more appropriate for a list of things.

So you're imagining an object that would look something like:
```
{
    row1: true,
    row2: true,
    ...etc
}
```

packages/table/components/CheckboxTable.tsx
mperrotti: `Table` needs to know which rows are selected for styling purposes. When a row is selected, it has a light grey background

packages/table/components/CheckboxTable.tsx
nLight: exactly, if it's just a helper data structure anyway

nLight: Though it won't be possible for _all_ types that `const rowId = rowData[uniqueKey];` can return. 
`rowData[uniqueKey]` could be an `object` right? But then again we can't compare objects anyway.

packages/table/components/CheckboxTable.tsx
nLight: Can this info be transferred differently? Like a method `CheckboxTable` would override?

packages/table/components/CheckboxTable.tsx
mperrotti: Yes, it _could_ return an object, but I'm trusting consumers would only use a piece of row data that's a `number` or `string`. I should add that detail in the prop documentation...

packages/table/components/CheckboxTable.tsx
mperrotti: Sure, I can give that a try.

mperrotti: I tried writing a method in Table that gets overridden in CheckboxTable to handle this, but I couldn't get it to behave as expected.
If this is important to fix before merging, maybe I could show you what I tried

packages/table/components/Table.tsx
```diff
@@ -308,15 +319,16 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
     const updateHoveredRowIndex = () => {
       this.setState({ hoveredRowIndex: rowIndex });
     };
-    const selectedRows = this.props.selectedRows || [];
+    const selectedRows = this.props.selectedRows || {};
```
Poltergeist: Maybe we should add default props for `selectedRows` and `uniqueKey`?

https://github.com/sw-yx/react-typescript-cheatsheet#typing-defaultprops

packages/table/components/Table.tsx
```diff
@@ -38,7 +38,18 @@ export interface TableProps {
   /**
    * Selected row data
    */
-  selectedRows?: object[];
+  selectedRows?: {};
+  /**
+   * Set the row height
+   */
+  rowHeight?: number;
```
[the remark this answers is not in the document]
mperrotti: I'm not sure how this got in here. This is from this PR: https://github.com/dcos-labs/ui-kit/pull/294

I have no problem with specifying a set of options instead of letting it be any number, but I would do it in the PR I linked above and remove it entirely from this one

packages/table/components/CheckboxTable.tsx
```diff
@@ -7,24 +7,52 @@ export interface CheckboxTableProps extends TableProps {
...
+
     this.getSelectedRows = this.getSelectedRows.bind(this);
     this.checkboxCellRenderer = this.checkboxCellRenderer.bind(this);
     this.getHeaderCheckbox = this.getHeaderCheckbox.bind(this);
   }
 
+  public componentWillReceiveProps(nextProps) {
+    const selectedLength = Object.keys(nextProps.selectedRows).length;
+    if (Boolean(selectedLength) && selectedLength === this.props.data.length) {
+      this.setState({ headerChecked: true });
```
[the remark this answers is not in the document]
mperrotti: Originally, I wasn't using state. I was using a mutable variable in `getHeaderCheckbox`. Dima requested I change it to state in his review.

If I just use `checked={this.state.headerChecked}`, it will not be checked on mount if a user passes all rows to `selectedRows`.

packages/table/components/Table.tsx
mperrotti: I get a Typescript error if I change these to default props

packages/table/components/CheckboxTable.tsx
mperrotti: I got around this with using theming, per Philipp's suggestion 🎉

GeorgiSTodorov:
File conflicts.

mperrotti:
@GeorgiSTodorov - a class name got removed from table cells when I resolved conflicts. Fixed now 👍

nLight:
Hey Mike, curious if you had a look into https://github.com/dcos/dcos-ui/blob/master/src/js/components/CheckboxTable.js#L105 
I wonder if we could completely avoid the state here.
Another question is around the `indeterminate` state. Does the propose implementation support it?

mperrotti:
@nLight
Yes, we could avoid having state in this component. I introduced state to address your feedback here: https://github.com/dcos-labs/ui-kit/pull/296#discussion_r256092946
Unless you're not talking about component state?

And yes, the header checkbox is shown in an indeterminate state when some (but not all) checkboxes are checked

packages/table/components/CheckboxTable.tsx
```diff
@@ -7,43 +8,74 @@ export interface CheckboxTableProps extends TableProps {
...
+    this.state = {
+      headerChecked: false
+    };
+
     this.getSelectedRows = this.getSelectedRows.bind(this);
     this.checkboxCellRenderer = this.checkboxCellRenderer.bind(this);
     this.getHeaderCheckbox = this.getHeaderCheckbox.bind(this);
   }
 
+  public componentWillReceiveProps(nextProps) {
```
brandonc: This method is deprecating and will disappear in React 17. It looks like you can replace it with getDerivedStateFromProps fairly easily

mperrotti: Great point, I'll swap it out.

GeorgiSTodorov:
@brandonc @Poltergeist

PR #300, opened 2019-02-16
[Spike] Speed up webpack

pierrebeitz:
This PR contains 3 commits that change behaviour but speed up stuff  
initial: 80s -> 8s
recompile: 8s -> 1.4s

1. introduce cache-loader to cache the output of `ts-loader` and especially `react-docgen-typescript-loader`. **This may introduce caching-bugs** (although a similar setup without `react-docgen-typescript-loader` a lot in the past in other teams without trouble) but speeds up the non-initial startup a lot. 

2. Move type checking into the background. **Type errors are only reported in the console now** and not in the overlay anymore. This speeds up startup and recompilation a little.

3. Switch to **cheaper source maps**. If anyone uses source maps a lot, we should remove that commit again. If not, we shave a whopping 4s per recompile. If nobody actually cares about source maps, we could remove them altogether to speed things up further.

## Before

80s - 1st startup 
77s - 2nd startup
8s - recompilation of removing the `main`-block within the appChrome-component

## After 
75s - 1st startup
8.3s - 2nd startup
1.4s - recompilation of removing the `main`-block within the appChrome-component

closes DCOS_OSS-4793


package.json
```diff
@@ -82,12 +82,14 @@
     "@types/storybook__addon-actions": "3.4.1",
     "@types/storybook__addon-info": "3.4.2",
     "@types/storybook__addon-knobs": "3.4.1",
+    "cache-loader": "^2.0.1",
```
nLight: let's pin version for both packages please

pierrebeitz: ✅

jaxesn:
@brandonc @pierrebeitz maybe something we could look at for the protoss repo too?

AmrAbdelrazik:
hey @pierrebeitz is that compile/engineering workflow performance improvement only or do customers (DC/OS UI users) see benefits as well?

pierrebeitz:
@jaxesn i have an unfinished spike for dcos-ui (it's a little complicated as it reintroduces type checking as well). 
will look into the protoss-setup afterwards.

@AmrAbdelrazik these changes are for engineers only.

PR #301, opened 2019-02-21
DCOS-48147: add tooltip component

mperrotti:
Adds a tooltip component to ui-kit

screen shot 2019-02-20 at 8 50 01 pm

### DIFF
--- a/packages/tooltip/README.md
+++ b/packages/tooltip/README.md
@@ -1,0 +1,21 @@
+# Tooltip
+
+Tooltips provide an additional explanation about a UI element when hovered or has focus.
+
+## Best practices:
+
+### Do:
+
+- Use Tooltips to show additional information when hovering or focusing a button that is just an icon with no text.
+- Use Tooltips to explan why a UI element is disabled when possible.
+- Use brief and concise language for Tooltip content.
+
+### Don't:
+
+- Use Tooltips to display critical information. Ideally, tooltips are only used for "nice-to-know" information.
+- Design interfaces that rely heavily on Tooltips to help the user complete their task. This is a sign that the design needs to be more clear.
+
+### Usage with a form field
+
+If you want to provide information about what input the field expects, use hint text below the field.
+If you want to provide an additional explanation about the field, use a Tooltip. Usually, Tooltips are triggered in form fields by an icon in the field's label.
--- a/packages/tooltip/components/TooltipContent.tsx
+++ b/packages/tooltip/components/TooltipContent.tsx
@@ -1,0 +1,47 @@
+import * as React from "react";
+import { cx } from "emotion";
+import { tooltip, getTooltipArrow } from "../style";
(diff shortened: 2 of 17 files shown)


packages/tooltip/components/TooltipContent.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
+import { cx } from "emotion";
+import { tooltip, getTooltipArrow } from "../style";
+import { darkMode, padding } from "../../shared/styles/styleUtils";
+import { BaseTooltipProps } from "./Tooltip";
+import { Direction } from "../../dropdownable/components/Dropdownable";
+
+export interface TooltipContentProps extends BaseTooltipProps {
+  direction?: Direction;
```
mperrotti: This is not actually optional, but it's passed via `React.cloneElement` from Dropdownable.

Anybody have any ideas how have proper type safety in this situation? I haven't been able to find a good solution

packages/tooltip/README.md
```diff
@@ -0,0 +1,3 @@
+# Tooltip Name
+
+Tooltip documentation
```
mperrotti: Hold merge until I have time to add documentation and review them with @izharbuendia

GeorgiSTodorov:
Why is `package-lock.json` changed, but `package.json` isn't?

mperrotti:
@GeorgiSTodorov 
> Why is `package-lock.json` changed, but `package.json` isn't?

It looks like the only changes in package-lock.json are the urls going form `http` to `https`. My guess is that this is caused by upgrading the node dev dependency to 10.15.1 in the icon sprite branch that merged earlier this week (https://github.com/dcos-labs/ui-kit/pull/292). I don't think it's anything to be concerned about

> The tooltip with min-width 100 has less than 100

That's a **max**-width, not a min-width. I updated the story to be clearer

> the toolitp with min-width 200 has much more that 200

That was from the padding. I updated it to use `border-box` box sizing.

GeorgiSTodorov:
Looks good. But I find it a little confusing that the text is called trigger. And that preferred direction is called "preferred", but is followed even when there is not space in that direction. Shouldn't it be called just direction?
screenshot from 2019-02-25 12-03-00

mperrotti:
@GeorgiSTodorov - I called that prop "preferredDirections" to be consistent with Dropdownable: https://github.com/dcos-labs/ui-kit/blob/master/packages/dropdownable/components/Dropdownable.tsx#L25

I'm having a hard time thinking of something more clear than "trigger" - do you have any suggestions? Or could you tell me why you find "trigger" confusing?

GeorgiSTodorov:
@mperrotti , I think something like "text" will be better, but if the others are okay with trigger, they I don't mind.

mperrotti:
Often times it's an icon, not text, so I think "text" would be more confusing

PR #302, opened 2019-02-23
DCOS-49026: fix icon export issues and add missing icons

mperrotti:
In order to replace the Secrets table, we need to be able to render the checkbox icon from the CheckboxTable being used in dcos-ui to replace the Secrets table.

Instead of adding an additional icon sprite to the one we already have, I've went and replaced the icon sprite and dcos-ui's Icon component with the sprite and Icon component from ui-kit.

While doing this, I noticed there were a few things missing from the ui-kit icons, and that https://github.com/dcos-labs/ui-kit/pull/292 broke the icon name enum export.

This PR resolves those issues and unblocks me from being able to replace the icon sprite and properly render the CheckboxTable in dcos-ui.


package.json
```diff
@@ -17,12 +17,10 @@
     "build:storybook": "build-storybook",
     "prebuild:storybook": "npm run build:design-tokens && npm run build:svg",
     "prebuild:design-tokens": "npm run clean:design-tokens",
-    "prebuild:svg": "npm run clean:svg",
     "build:design-tokens": "node ./packages/design-tokens/scripts/build.js",
     "build:svg": "node ./packages/icons/scripts/build.js",
     "clean": "rimraf dist",
     "clean:design-tokens": "rm -rf ./dist/packages/design-tokens && rm -rf ./packages/design-tokens/dist/",
-    "clean:svg": "rm -rf ./dist/packages/icons && rm -rf ./packages/icons/dist/",
```
mperrotti: Removing this "clean" step because when we merge to master, Jenkins runs the Typescript compiler first, and then runs the icon build again (including the "clean" step) before building and deploying Storybook.

This resulted in the JavaScript files that were generated in `dist/icons/` to be removed right before we publish to NPM.

Check this log to see what I'm talking about: https://jenkins.mesosphere.com/service/jenkins/job/frontend/job/UI-Kit%20Pipeline/job/master/208/console

packages/icon/components/Icon.tsx
```diff
@@ -35,7 +36,7 @@ export interface IconProps {
 
 const Icon = (props: IconProps) => {
   const { color, size, shape, ariaLabel } = props;
-  const svgColor = color || textColorPrimary;
+  const svgColor = color || "currentColor";
```
natmegs: Should `"currentColor"` be a variable? Or how does this work?

mperrotti: @natmegs - "currentColor" is a keyword in CSS that ends up being whatever color it would be if it was just a piece of text.
So if we had
```
.blueText {
    color: blue;
}
.blueText div {
    border-color: currentColor;
}
```
the border of any divs that are children of `.blueText` would be blue

PR #304, opened 2019-02-25
fix(icons): fix attributes from kebab-case to camel case

Poltergeist:
React is expecting CamelCase attributes this fixes the kebab-case attributes on icon to CamelCase


mperrotti:
@Poltergeist - For what it's worth, there's a (very stale) spike day PR that gets rid of this file altogether

PR #305, opened 2019-03-04
DCOS-48552: add the ability to disable CheckboxTable rows

mperrotti:
Added the ability to set CheckboxTable rows to be non-selectable and/or visually muted

screen shot 2019-03-04 at 3 57 55 pm

### DIFF
--- a/packages/table/components/CheckboxTable.tsx
+++ b/packages/table/components/CheckboxTable.tsx
@@ -14,6 +14,14 @@ export interface CheckboxTableProps extends TableProps {
    */
   selectedRows?: {};
   /**
+   * Rows that cannot be selected
+   */
+  disabledRows?: {};
+  /**
+   * Rows that are in some incomplete state. Typically used with disabledRows
+   */
+  mutedRows?: {};
+  /**
    * The unique key in each row's data object.
    *
    * Ideally, the value of this key in your table's data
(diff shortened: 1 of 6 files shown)


packages/table/components/CheckboxTable.tsx
```diff
@@ -109,20 +139,26 @@ class CheckboxTable extends React.PureComponent<
       this.setState({ headerChecked: e.target.checked });
     };
 
+    console.log(`Boolean(selectedLength): ${Boolean(selectedLength)}`);
+    console.log(
+      `selectedLength === maxSelectedLength: ${selectedLength ===
+        maxSelectedLength}`
+    );
+    console.log(`this.state.headerChecked: ${this.state.headerChecked}`);
```
natmegs: 🎨

packages/table/stories/CheckboxTable.stories.tsx
```diff
@@ -151,4 +151,43 @@ storiesOf("Table/CheckboxTable", module)
       />
     </CheckboxTableHelper>
   ))
+  .addWithInfo("w/ disabled row", () => (
+    <CheckboxTableHelper
+      data={items}
+      disabledRows={{ [items[0].name]: true }}
```
GeorgiSTodorov: This seems like an unnecessarily complex data structure to me. Can we do something like
`disabledRows = [items[0].name, items[1].name]`?

mperrotti: I wanted to be consistent with the data structure for `selectedRows`. Or do you think it's so complex that it's not worth being consistent with `selectedRows`?

GeorgiSTodorov: In that case I prefer consistency then.

PR #306, opened 2019-03-05
DCOS-49609 refactor(Toaster): children as source-of-truth

TattdCodeMonkey:
Refactored Toaster to always trust it's children as the source of truth
for which toasts should be displayed. This requires the app using
toaster to keep and accurate list of toasts and handle removing toasts
from that list whenever onDismiss callback is invoked.

Updated `autodismiss` code to also ensure onDismiss callback is called.

### DIFF
--- a/packages/toaster/tests/Toast.test.tsx
+++ b/packages/toaster/tests/Toast.test.tsx
@@ -64,21 +64,14 @@ describe("Toast", () => {
     expect(action).toHaveBeenCalled();
   });
 
-  it("calls dismissToast and the onDismiss callback when the dismiss button is clicked", () => {
+  it("calls dismissToast when the dismiss button is clicked", () => {
     const dismissToastSpy = spyOn(Toaster.prototype, "dismissToast");
-    const onDismiss = jest.fn();
     const component = mount(
-      <Toaster
-        children={[
-          <Toast onDismiss={onDismiss} title="I Am Toast" key={0} id={0} />
-        ]}
-      />
+      <Toaster children={[<Toast title="I Am Toast" key={0} id={0} />]} />
     );
     const dismissBtn = component.find('span[role="button"]');
-    expect(onDismiss).not.toHaveBeenCalled();
     expect(dismissToastSpy).not.toHaveBeenCalled();
     dismissBtn.simulate("click");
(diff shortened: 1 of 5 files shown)


packages/toaster/tests/Toast.test.tsx
```diff
@@ -64,21 +64,14 @@ describe("Toast", () => {
...
     const component = mount(
-      <Toaster
-        children={[
-          <Toast onDismiss={onDismiss} title="I Am Toast" key={0} id={0} />
-        ]}
-      />
+      <Toaster children={[<Toast title="I Am Toast" key={0} id={0} />]} />
     );
     const dismissBtn = component.find('span[role="button"]');
-    expect(onDismiss).not.toHaveBeenCalled();
```
TattdCodeMonkey: `onDismiss` is now called by `dismissToast` which is being replaced here. It is tested in Toaster though.

packages/toaster/tests/Toaster.test.tsx
```diff
@@ -33,17 +33,19 @@ describe("Toaster", () => {
     expect(component.state("toasts")).not.toContain(testToast[0].props.id);
   });
 
-  it("dismisses toasts after specified time", () => {
```
TattdCodeMonkey: this test was broken and doesn't seem valid. We only want toasts automatically removed if `autodismiss=true` which we have a different test for.

packages/toaster/tests/Toaster.test.tsx
```diff
@@ -33,17 +33,19 @@ describe("Toaster", () => {
...
 
-    expect(Object.keys(component.state("toasts")).length).toBe(1);
-    setTimeout(() => {
-      expect(Object.keys(component.state("toasts")).length).toBe(0);
-    }, DELAY_TIME + 1);
+    expect(dismissCallback).toHaveBeenCalledTimes(1);
   });
 
-  it("dismisses autodismiss toasts after specified time", () => {
+  it("dismisses autodismiss toasts after specified time", done => {
```
TattdCodeMonkey: added done so that test worked. without this would always pass.

mperrotti: Nice, I didn't know about that parameter

PR #309, opened 2019-03-09
Table cell perf fix: only re-render classname for cells

mperrotti:
We noticed in Protoss that the table cells were rewriting the DOM nodes for table cells instead of changing the attributes of the table cell nodes. This was causing images to flicker on state change and components like `Tooltip` inside the table to unnecessarily do expensive re-renders.

Before:
TableCellReRender-before

After:
TableCellReRender-after

### DIFF
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -111,6 +111,22 @@ export const fillColumns = (
   }
 };
 
+const ContentCell = styled("div")`
+  ${props => {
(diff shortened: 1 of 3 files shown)


AmrAbdelrazik:
I wonder if there are other areas of the DC/OS UI that are already doing this expensive re-render

mperrotti:
It's unlikely - the change that caused this has only been merged for a week

PR #310, opened 2019-03-11
fix: don't create styled component in render function

brandonc:
There are a few more examples of this in Sidebar and Headerbar, but this in particular is much easier to fix and is affecting protoss performance

### DIFF
--- a/packages/dropdownable/tests/__snapshots__/Dropdownable.test.tsx.snap
+++ b/packages/dropdownable/tests/__snapshots__/Dropdownable.test.tsx.snap
@@ -259,19 +259,9 @@ exports[`Dropdownable is visible after opening 2`] = `
         className="emotion-1"
         overlayRoot={
           <body>
-            .emotion-0 {
-  top: 0px;
-  left: 0px;
-  position: absolute;
-  -webkit-transform: scale(1);
-  -ms-transform: scale(1);
-  transform: scale(1);
-  opacity: 1;
-}
-
-<div>
+            <div>
               <div
-                class="emotion-0"
+                class="emotion-1"
               >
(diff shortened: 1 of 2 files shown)


brandonc:
@jaxesn had to update a snapshot

packages/dropdownable/tests/__snapshots__/Dropdownable.test.tsx.snap
```diff
@@ -259,19 +259,9 @@ exports[`Dropdownable is visible after opening 2`] = `
...
-  left: 0px;
-  position: absolute;
-  -webkit-transform: scale(1);
-  -ms-transform: scale(1);
-  transform: scale(1);
-  opacity: 1;
-}
-
-<div>
+            <div>
```
jaxesn: I dont understand why this changed?  This looks like an unrelated class?

brandonc: I was surprised at first, but if styled was re-creating a component every render, it could be that it creates these styles each render as well, which was happening multiple times in this test. Now that the component is only created once, there's less style code created?

brandonc: I'm not too familiar with what emotion does under the hood

jaxesn: @mperrotti we might need you to weigh in on this

mperrotti: @brandonc's assessment sounds correct to me

PR #311, opened 2019-03-14
DCOS-50093: table perf - remove content-based column width sizing

mperrotti:
This removes the Table's ability to measure all cells and calculate column widths based on content sizing. We're removing this because that functionality was making the Table's performance very poor in dcos-ui.

I investigated alternatives that would allow us to keep the content-based column width sizing, but could not find a way to do it that didn't seriously hurt Table performance.

### DIFF
--- a/packages/table/components/CheckboxTable.tsx
+++ b/packages/table/components/CheckboxTable.tsx
@@ -3,6 +3,9 @@ import { ThemeProvider } from "emotion-theming";
 import { Table, Column, Cell, HeaderCell } from "..";
 import { TableProps } from "./Table";
 import CheckboxInput from "../../checkboxInput/components/CheckboxInput";
+import { WidthArgs } from "./Column";
+import { display } from "../../shared/styles/styleUtils";
+import { spaceM } from "../../design-tokens/build/js/designTokens";
 
 export interface CheckboxTableProps extends TableProps {
   /**
@@ -48,6 +51,7 @@ class CheckboxTable extends React.PureComponent<
           data.length - Object.keys(disabledRows).length
     };
   }
+  private checkboxRef = React.createRef<HTMLDivElement>();
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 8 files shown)


packages/table/components/CheckboxTable.tsx
```diff
@@ -48,6 +51,8 @@ class CheckboxTable extends React.PureComponent<
           data.length - Object.keys(disabledRows).length
     };
   }
+  // public checkboxRef: any = {};
```
GeorgiSTodorov: Forgotten comment.

packages/table/components/Table.tsx
```diff
@@ -285,19 +261,7 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
-        cache={this.cellMeasureCache}
-        columnIndex={columnIndex}
-        key={key}
-        parent={parent}
-        rowIndex={rowIndex}
-      >
-        {this.getCell(args)}
-      </CellMeasurer>
-    );
+    return this.getCell(args);
```
juliangieseke: am I correct that this change breaks existing functionality? If so, please add a notice to commit message 🙇 

e.g.
```
fix(table): remove content-based column width sizing

BREAKING CHANGE:
This change <whatitdoes>

Before: <short explanation>
After: <short explanation>

Closes DCOS-50093
```

going forward we should take care of breaking changes and our commit messages in general - and especially in ui-kit, since we're going to release dcos-ui quite often and need to be able to maintain older versions. 
remember the issues we had with `reactjs-components` 😿 

ref: https://www.conventionalcommits.org/en/v1.0.0-beta.3/#specification

mperrotti: Ok, I was undecided whether to add a breaking change for this commit since everything will still build and run just fine.
But I guess the fact that table columns without `width` passed won't behave the same as they used to is a good enough reason to flag it as a breaking change 👍

PR #317, opened 2019-03-21
DCOS-50412: account for scrollbars for Table col widths

mperrotti:
Before, table column widths were being calculated based on the width of the area the table filled. However, on Windows and Linux, a scrollbar is displayed on the right and causes a horizontal scroll.

Now, we measure the scrollbar width and factor that into the column width calculations.

TableScrollbarFix


mperrotti:
macOS users, there's a way for you to test this too :)

Go to System Preferences > General, then check the radio button for "Always" under the "Show scroll bars" section.

PR #320, opened 2019-04-03
feat(table): add prop overloading to column

DanielMSchmidt:


packages/table/components/Column.tsx
```diff
@@ -14,7 +18,11 @@ export interface ColumnProps {
   /**
    * cellRenderer is the function which is creating the cell contents for this column.
    */
-  cellRenderer: (data: any, width: number) => React.ReactNode;
+  cellRenderer: (
+    data: any,
+    width: number,
+    additionalProps: OverloadedProps
```
mperrotti: Is there a reason this can't be optional?

DanielMSchmidt: It should at least be an empty object

GeorgiSTodorov: This will break existing tables in dcos-ui, then?

GeorgiSTodorov:
A test is failing.
Screenshot from 2019-04-04 10-33-38

packages/table/components/Column.tsx
GeorgiSTodorov: It doesn't, I might have misunderstood.

DanielMSchmidt: It shouldn't, you can just ommit the argument in the cell renderer :)

GeorgiSTodorov:
@DanielMSchmidt , file conflicts.

mperrotti:
@DanielMSchmidt - could you either update this or close it?

PR #321, opened 2019-04-04
DCOS-46337: resizable table columns

mperrotti:
Table columns can be resized by setting the `resize` prop on the respective `Column`

TableResizeDemo_04092018


mperrotti:
Hold merge until design gives their final signoff

mperrotti:
@GeorgiSTodorov - I added a maximum width a column can be resized. I set it so no column can take up more than 80% of the table container's width.

GeorgiSTodorov:
I broke it again.
Peek 2019-04-05 17-23

mperrotti:
Clever, @GeorgiSTodorov. Were you a QA engineer in a past life? :p

I'll try and guard against that scenario, but I don't know if that's a realistic state a user would ever get to.

GeorgiSTodorov:
@mperrotti , can we set a maximum width for the table?

mperrotti:
I don't understand why we'd do that. Tables should be able to be as wide as they need to be.
I pushed up a fix I'm pretty sure you won't be able to break this time ;)

GeorgiSTodorov:
Does this count?
Peek 2019-04-05 20-51

mperrotti:
@GeorgiSTodorov - Your thorough testing never ceases to amaze me :)

mperrotti:
Reviewed with design today. Two things left to decide on:
- Visual design of the resize handle
- Whether resizing is opt-in or opt-out (use a `noResize` prop instead of `resizable`?)

AmrAbdelrazik:
@mperrotti are we getting double click to resize & fit content?
@izharbuendia may I suggest opt-out please forgetting to block sizing of a column is less important than forgetting to enable sizing on an important column (it would be very annoying)

mperrotti:
@AmrAbdelrazik 
We are not getting double-click to resize and fit content in this PR. I tried implementing it, and it hurt performance when scrolling and hovering in a DC/OS soak cluster

GeorgiSTodorov:
This might be an edge case.

mperrotti:
@GeorgiSTodorov - that's very edge case-y. I feel confident users won't get themselves into that situation.

mperrotti:
@natmegs and @GeorgiSTodorov - please re-review. I made visual design updates to the resize handle requested by the design team, and moved the `onResize` prop from Table to Column because it seemed like a better component API

PR #324, opened 2019-04-12
rm table wrapper div that breaks integration tests

mperrotti:
For some reason, when the table is wrapped in a div, it will not be rendered by Cypress' browser until you explicitly set the correct height on that wrapper div.

Using a `Fragment` instead of a DOM node will also fix this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "react-virtualized": "9.20.1"
   },
   "peerDependencies": {
-    "react": ">=15.4.1 <16.9.0 || 16.7.0-alpha.0",
-    "react-dom": ">=15.4.1 <16.9.0 || 16.7.0-alpha.0"
+    "react": ">=16.2.0 <16.9.0 || 16.7.0-alpha.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -55,8 +55,8 @@
     "react-virtualized": "9.20.1"
   },
   "peerDependencies": {
-    "react": ">=15.4.1 <16.9.0 || 16.7.0-alpha.0",
-    "react-dom": ">=15.4.1 <16.9.0 || 16.7.0-alpha.0"
+    "react": ">=16.2.0 <16.9.0 || 16.7.0-alpha.0",
```
nLight: that's technically a breaking change 👼 

https://www.conventionalcommits.org/en/v1.0.0-beta.2/#commit-message-with-description-and-breaking-change-in-body

mperrotti: Ah, you're right! I'll update the commit with the appropriate information.

We're using React 16+ in the Mesosphere projects that use ui-kit, so it shouldn't be a problem to upgrade

mperrotti:
@GeorgiSTodorov - it should be there now 👍

PR #325, opened 2019-04-12
Fix Storybook deployment

mperrotti:
I'm _pretty_ sure this will work, but there's not really a way to test locally.
I tested it by running `deploy:storybook` with the same options I set in the Jenkinsfile, and it worked

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ tmp/
 .vscode/
 .env
 coverage
-storybook-static/
+storybook_static/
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
From Jenkins:
```
Archiving artifacts

‘storybook-static/**/*’ doesn’t match anything, but ‘**/*’ does. Perhaps that’s what you mean?

No artifacts found that match the file pattern "storybook-static/**/*". Configuration error?
```

PR #327, opened 2019-04-22
DCOS-45978: Add label tooltip to text input component

GeorgiSTodorov:
Add an option for TextInput to have a tooltip with an icon.

Closes https://jira.mesosphere.com/browse/DCOS-45978

## Testing
Verify that a `TextInput` with `tooltipContent` works and so do the other `TextInput`s.
Please review critically.

## Screenshots
Screenshot from 2019-04-23 10-48-30

Screenshot from 2019-04-25 09-36-44

Screenshot from 2019-04-23 10-49-33


packages/textInput/components/TextInput.tsx
```diff
@@ -90,7 +103,7 @@ export class TextInput<
           flush("top"),
           margin("bottom", "xxs"),
           textWeight("medium"),
-          display("block"),
+          this.props.tooltipText ? display("inline-block") : display("block"),
```
GeorgiSTodorov: This was needed for the tooltip to be on the same line.

packages/textInput/components/TextInput.tsx
```diff
@@ -65,13 +76,15 @@ export class TextInput<
...
     const labelContent = this.getLabelContent();
+    const tooltipContent = this.getTooltipContent();
     const containerProps: { className?: string } = {};
     if (this.props.className) {
       containerProps.className = this.props.className;
     }
     return (
       <div {...containerProps}>
         {labelContent}
+        {tooltipContent}
```
mperrotti: We get nicer vertical alignment if we use a flexbox container to wrap both the `<label>` and the `<Tooltip>`, and move the bottom margin from the `<label>` to the flexbox container:
```
<div className={cx(margin("bottom", "xxs"), flex({ align: "center" }))}>
    {labelContent}
    {tooltipContent}
</div>
```

Then, we can also get rid of the ternary in `getLabelContent()`: https://github.com/dcos-labs/ui-kit/pull/327/files#diff-d6ddbf113c06302e95d5b248c029f35fR106

packages/textInput/components/TextInput.tsx
```diff
@@ -177,6 +192,29 @@ export class TextInput<
...
+      <span className={cx(margin("left", "xs"))}>
+        <Tooltip
+          trigger={
+            <Icon
+              color={greyLightDarken2}
+              shape={SystemIcons.CircleQuestion}
+              size={iconSizeXs}
+            />
+          }
+          id="tooltipTextinput"
```
mperrotti: We'll get duplicate IDs if we just use a string for this. We could use the `id` prop to make the tooltip IDs unique

mperrotti:
One other thing to consider:
I've seen some long pieces of text in some of the field label tooltips in dcos-ui. Should we pass a default `maxWidth` to the `Tooltip` component here? 200px feels reasonable

packages/textInput/components/TextInput.tsx
```diff
@@ -177,6 +192,29 @@ export class TextInput<
...
     );
   }
+
+  protected getTooltipContent(): React.ReactNode {
+    if (!this.props.tooltipContent) {
+      return null;
+    }
+
+    return (
+      <span className={cx(margin("left", "xs"))}>
```
mperrotti: No need for the `cx` here. You can just do `className={margin("left", "xs")}`

packages/textInput/components/TextInput.tsx
```diff
@@ -177,6 +192,29 @@ export class TextInput<
...
+        <Tooltip
+          trigger={
+            <Icon
+              color={greyLightDarken2}
+              shape={SystemIcons.CircleQuestion}
+              size={iconSizeXs}
+            />
+          }
+          id={this.props.id}
```
mperrotti:
```suggestion
          id={`labelTooltip-${this.props.id}`}
```

mperrotti: Otherwise the tooltip will have the same ID as the text input

packages/textInput/components/TextInput.tsx
```diff
@@ -177,6 +192,29 @@ export class TextInput<
...
+        >
```
mperrotti:
```suggestion
          maxWidth={200}
        >
```

mperrotti: This would apply the default `maxWidth` I mentioned :)

mperrotti:
Also we don't need these classes anymore. Once we remove them, the vertical alignment gets better:
https://github.com/dcos-labs/ui-kit/pull/327/files#diff-d6ddbf113c06302e95d5b248c029f35fR106
https://github.com/dcos-labs/ui-kit/pull/327/files#diff-d6ddbf113c06302e95d5b248c029f35fR106

GeorgiSTodorov:
@mperrotti 
> Also we don't need these classes anymore. Once we remove them, the vertical alignment gets better:
https://github.com/dcos-labs/ui-kit/pull/327/files#diff-d6ddbf113c06302e95d5b248c029f35fR106
https://github.com/dcos-labs/ui-kit/pull/327/files#diff-d6ddbf113c06302e95d5b248c029f35fR106

You posted the same line of code twice. Did you mean to do that?

GeorgiSTodorov:
@mperrotti , I addressed your feedback. Regarding the comment about removing classes, you posted the same line twice, so I only removed one class. Are there any other classes that need to be removed?

mperrotti:
Oops, I did not mean that. I meant `display("block")`

mperrotti:
Snapshots just need to be updated :)

GeorgiSTodorov:
Merging the latest master was causing some snapshots, unrelated to this PR, to fail. I fixed them now. @pierrebeitz @TattdCodeMonkey , please re-review.]
Screenshot from 2019-05-07 11-25-59

PR #328, opened 2019-05-01
DCOS-52782: add Typeahead component

mperrotti:
TypeaheadDemo

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   ],
   "typings": "dist/packages/index",
   "dependencies": {
+    "downshift": "3.2.10",
     "emotion": "9.2.12",
     "emotion-theming": "9.2.9",
     "focus-trap-react": "4.0.1",
--- a/packages/dropdowMenu/README.md
+++ b/packages/dropdowMenu/README.md
@@ -1,0 +1,3 @@
+# DropdownMenu Name
+
+DropdownMenu documentation
--- a/packages/dropdownable/components/Dropdownable.tsx
+++ b/packages/dropdownable/components/Dropdownable.tsx
@@ -24,14 +24,14 @@ const DEFAULT_DIRECTION_PREFERENCES = [
 
 const Container = styled("div")`
   position: relative;
-  display: inline-block;
 `;
 
 export interface DropdownableProps {
   open: boolean;
   dropdown: React.ReactElement<any>;
   preferredDirections?: Direction[];
-  onClose: () => void;
+  onClose?: () => void;
+  overlayRoot?: HTMLElement;
 }
 
 interface PositionCoord {
@@ -110,7 +110,7 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
 
   getOverlay() {
     const { position } = this.state;
-    const { open } = this.props;
+    const { open, overlayRoot } = this.props;
 
     const overlayStyle = css({
       top: `${position.top}px`,
@@ -120,7 +120,11 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
       opacity: open ? 1 : 0,
       "z-index": zIndexDropdownable
     });
-    return <Overlay className={overlayStyle}>{this.getDropdown()}</Overlay>;
+    return (
+      <Overlay overlayRoot={overlayRoot} className={overlayStyle}>
+        {this.getDropdown()}
+      </Overlay>
+    );
   }
 
   getDropdown() {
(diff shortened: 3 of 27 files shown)


packages/dropdowMenu/README.md
```diff
@@ -0,0 +1,3 @@
+# DropdownMenu Name
```
mperrotti: Will document with Daniel when we meet in-person on Friday

packages/typeahead/README.md
```diff
@@ -0,0 +1,3 @@
+# Typeahead Name
```
mperrotti: Will document with Daniel when we meet in-person on Friday

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -24,14 +24,14 @@ const DEFAULT_DIRECTION_PREFERENCES = [
...
   position: relative;
-  display: inline-block;
 `;
 
 export interface DropdownableProps {
   open: boolean;
   dropdown: React.ReactElement<any>;
   preferredDirections?: Direction[];
-  onClose: () => void;
+  onClose?: () => void;
```
mperrotti: @brandonc - the Typeahead component doesn't need to manage the state for whether the dropdown menu is visible. I'm using render props from [downshift](https://github.com/downshift-js/downshift) to update the `open` prop

If you want `onClose` to be required when using `Dropdownable`, I can take a different path

packages/typeahead/components/Typeahead.tsx
```diff
@@ -0,0 +1,252 @@
...
+   */
+  multiSelect?: boolean;
+  /**
+   * callback for when an item is selected
+   */
+  onSelect?: (selectedItems: string[]) => void;
+  /**
+   * which DOM node the dropdown menu will attach to
+   */
+  overlayRoot?: HTMLElement;
```
mperrotti: I had to add this to support usage inside of modals in dcos-ui. If it portals to `document.body`, it will be covered by a modal because modals in dcos-ui have `z-index: 999`.

package.json
```diff
@@ -40,6 +40,7 @@
   ],
   "typings": "dist/packages/index",
   "dependencies": {
+    "downshift": "3.2.10",
```
mperrotti: [Downshift](https://github.com/downshift-js/downshift) is a great little library that handles keyboard navigation within the menu and other a11y goodies like updating ARIA roles

GeorgiSTodorov:
Looks good. Some questions:
- Should the multi-select show the selected values in the input? Having an empty input seems useless.
- I need to un-focus the input to be able to select again without changing the value. Is this intended?
Peek 2019-05-02 11-43
- I can't select the text values in order to copy them. Should I be able to? The cursor when hovering them is the one that shows that they can be selected.
- Should we make the input have customizable width? Sometimes we need a smaller or a bigger input.

mperrotti:
> Should the multi-select show the selected values in the input? Having an empty input seems useless.

That is intentional. The idea is that you could use `onSelect` to do something with those values. Here's an example from the design team's Sketch symbols for how those might be used: https://cl.ly/45baadd535ff

Implementing this input is out of scope for this ticket

---

> I need to un-focus the input to be able to select again without changing the value. Is this intended?

I'll update the dropdown to open on click as well as on focus.

---

> I can't select the text values in order to copy them. Should I be able to? The cursor when hovering them is the one that shows that they can be selected.

Can't select them from the list or from the text input?

---

> Should we make the input have customizable width? Sometimes we need a smaller or a bigger input.

The width of the parent element decides the width of the input

packages/typeahead/stories/Typeahead.stories.tsx
```diff
@@ -0,0 +1,125 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+import { Typeahead, TextInput } from "../../index";
+import MultiselectTypeahead from "./helpers/MultiselectTypeahead";
+import FilteredListTypeahead from "./helpers/FilteredListTypeahead";
+import { padding } from "../../shared/styles/styleUtils";
+import { complexItems, items } from "./helpers/itemMocks";
+import { css } from "emotion";
```
GeorgiSTodorov: Non-blocker, but non-local imports should be before local imports.

GeorgiSTodorov:
> Can't select them from the list or from the text input?

From the list.

mperrotti:
@GeorgiSTodorov - I don't think that's a problem

PR #330, opened 2019-05-21
[Spike] chore(scripts): fix create component script to support new storybook version

Poltergeist:


GeorgiSTodorov:
It might be useful to add this to the spike board so that people can easily find it.

PR #332, opened 2019-05-21
DCOS-54457 feat(configurationmap): add configuarationmap component

Poltergeist:
This adds a minimal ConfigurationMap component

image

Is part of DCOS-54457


packages/configurationmap/components/ConfigurationMap.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+// import { style } from "../style";
```
GeorgiSTodorov: Forgotten comment.

GeorgiSTodorov:
It might be useful to add this to the spike board so that people can easily find it.

packages/configurationmap/components/ConfigurationMap.tsx
```diff
@@ -0,0 +1,26 @@
...
+class ConfigurationMap extends React.PureComponent<ConfigurationMapProps, {}> {
+  public render() {
+    const { data } = this.props;
+
+    const tableData = Object.entries(data);
+
+    return (
+      <Table data={tableData}>
+        <Column header="" cellRenderer={cellRenderer(0)} />
+        <Column header="" cellRenderer={cellRenderer(1)} />
```
GeorgiSTodorov: Why do we have two rows hard-coded?

Poltergeist: A Configuration map is this in DC/OS 
image
The most minimal approach to this is a table with 2 Columns. This is why there are 2 Columns hard coded.

GeorgiSTodorov: But still, having a custom number of rows wouldn't be much less minimal and it would be useful.

Poltergeist: Could you give me an example? I think I am misunderstanding your intend.

GeorgiSTodorov: In practice we can have a different number of rows and columns. So for this component to be useful, it should support that.

Poltergeist: I will discuss this with design, to come to an agreement. Currently I would suggest to do it only with 2 columns since this will cover most of the use cases. Since this is a minimal implementation this should be enough ground covered.

DanielMSchmidt: @GeorgiSTodorov Where do we have multiple number of columns? Wouldn't this be more of a table than a config map?

packages/configurationmap/README.md
```diff
@@ -0,0 +1,3 @@
+# ConfigurationMap Name
+
+ConfigurationMap documentation
```
DanielMSchmidt: Much extensive, very communication. img (Seriously, I'm fine with auto generating, but we could give this at least a sentence for what it is for

packages/configurationmap/components/ConfigurationMap.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+// import { style } from "../style";
+import { Table, Column } from "../../";
+
+export interface ConfigurationMapProps {
+  data: { [key: string]: string };
```
DanielMSchmidt: Is the value required to be string or could it also be a number?

packages/configurationmap/components/ConfigurationMap.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const cellRenderer = column => data => <span>{data[column]}</span>;
```
DanielMSchmidt: This is a problem for rerenders: Every time data changes, we are going to have a new method being passed to the cell renderer, making every part below that rerendered. 

```suggestion
const cellRenderer = column => data => <span>{data[column]}</span>;
+ const keyRenderer = cellRenderer(0);
+ const valueRenderer = cellRenderer(1);
```

Making the references static would solve the problem

packages/configurationmap/style.ts
```diff
@@ -0,0 +1,3 @@
+import { css } from "emotion";
```
DanielMSchmidt: Do we need this file?

mperrotti: I'd be in favor of removing this until we actually have styles

packages/configurationmap/components/ConfigurationMap.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+class ConfigurationMap extends React.PureComponent<ConfigurationMapProps, {}> {
+  public render() {
+    const { data } = this.props;
+
+    const tableData = Object.entries(data);
+
+    return (
+      <Table data={tableData}>
```
mperrotti: Do you think we'll use the `Table` component for the real ConfigurationMap implementation? I'd prefer we didn't - it carries a lot of JavaScript to execute, and we'd have to come up with a way to override some of the styles (e.g.: the row hover background color change)

packages/configurationmap/README.md
mperrotti: I agree with @DanielMSchmidt , we should have at least a brief description of what the configuration map is and when it is used.

packages/configurationmap/components/ConfigurationMap.tsx
GeorgiSTodorov: @DanielMSchmidt , I don't know if we have more than two anywhere, but I can imagine situations where we want to have more than two. I want our minimal component to be usable in every case.

DanielMSchmidt: I think if we try to build something that works for all use cases we will spend time that is not needed. Let's iterate on it and if we have a config map with more then a key value relationship we can see if it's really a config map or a table or a special new component

GeorgiSTodorov: Okay, agreed.

packages/configurationmap/components/ConfigurationMap.tsx
Poltergeist: I don't think that we are going to use the table in the final implementation. My thinking was like a configuration map is a two column table which keeps key's and values. And the way of the least effort was to just use the table for it. Alternative could have been to use html table. And I agree that the table is not the best implementation for the final configuration map. Although the row highlight could be helpful ;-)

packages/configurationmap/README.md
Poltergeist: Added some documentation.

packages/configurationmap/components/ConfigurationMap.tsx
GeorgiSTodorov: I would vote for using the default html table.

PR #334, opened 2019-05-23
DCOS-54458 feat(chart): add chart component

Poltergeist:
This adds a minimal chart component which is using react-chartist to create a chart from provided
data. Provided data needs to be in the format `{%label%: %data%}`.

Is part of DCOS-54458


packages/chart/components/Chart.tsx
```diff
@@ -0,0 +1,39 @@
...
+      labels: Object.keys(data),
+      series: [Object.values(data)]
+    };
+    return (
+      <ChartistGraph
+        type={type}
+        data={graphData}
+        options={{
+          // @ts-ignore
+          fullWidth: true
```
GeorgiSTodorov: Why is this needed?

GeorgiSTodorov:
Looks good, but the last year doesn't fit on the screen.
Screenshot from 2019-05-27 14-38-55

packages/chart/components/Chart.tsx
Poltergeist: Because fullWidth is not part of the interface in the options of the library we are using.

packages/chart/index.ts
```diff
@@ -0,0 +1 @@
+export { default as Chart } from "./components/Chart";
```
mperrotti: Could we be more specific and call this component `LineChart`?

packages/chart/components/LineChart.tsx
```diff
@@ -0,0 +1,39 @@
...
+export interface LineChartProps {
+  data: {
+    [label: string]: number;
+  };
+  type: string;
+}
+
+class LineChart extends React.PureComponent<LineChartProps, {}> {
+  public static defaultProps = {
+    type: "Line"
```
brandonc: Why can you change LineChart to a different chart?

Poltergeist: Good catch. Forgot to remove thins when I changed this to LineChart.

Poltergeist:
I used chartist since it was the easiest way to create a chart component, we are only providing a very thin wrapper so that we can exchange chartist for something else or even create something ourself at a later point in time. So chartist is not here to stay and we only provide the abstraction which is as universal as possible.

PR #336, opened 2019-05-24
fix: extend DropdownMenuItem bg color with text

mperrotti:
If a menu item's text is one long string with no spaces, the background color would only extend to the width of the menu.

This seems like an unlikely scenario, but we discovered it by testing a Typeahead that could contain long ID strings.

## Screenshots
Background has been changed to pink for demonstration purposes

Before:
TypeaheadTextHighlight_before

After:
TypeaheadTextHighlight_after


mperrotti:
[Merging with only 1 review](https://media.giphy.com/media/2L4YDs59H7ssM/giphy.gif) because I'm confident this won't break anything

PR #337, opened 2019-05-24
fix(clickable): add cursor pointer to clickable elements

Poltergeist:
This adds cursor pointer to clickable items.

Testing:

This is a change to the `Clickable` component which can be tested in Storybook here: 

`?path=/story/clickable--default`


GeorgiSTodorov:
I can't find any difference to master, which clickable items does this affect?

Poltergeist:
Oh sorry :-( I thought it was clear enough will add this to the description.

PR #338, opened 2019-05-24
DCOS-54459: feat(tabs): add tabs component

Poltergeist:
This adds the minimal version of a tabs component. The API look like this:
```JSX
  class Example extends React.Component<{}, Partial<{ selectedIndex: number }>> {
      state = { selectedIndex: 0 };
      handleSelect = selectedIndex => {
        this.setState({ selectedIndex });
      };
      render() {
        const { selectedIndex } = this.state;
        return (
          <Tabs selectedIndex={selectedIndex} onSelect={this.handleSelect}>
            <TabItem>
              <TabTitle>Tab 1 Name</TabTitle>
              First tab Content
            </TabItem>
            <TabItem>
              <TabTitle>Tab 2 Name</TabTitle>
              Second Tab Content
            </TabItem>
          </Tabs>
        );
      }
    }
```

Depends on #337 

Part of DCOS-54459

### DIFF
--- a/packages/tabs/README.md
+++ b/packages/tabs/README.md
@@ -1,0 +1,31 @@
+# Tabs Name
+
+State of the Tabs is controlled with in the component.
+
+Example Component:
+
+```JS
+class Example extends React.Component<{}, Partial<{ selectedIndex: number }>> {
+  state = { selectedIndex: 0 };
+  handleSelect = selectedIndex => {
+    this.setState({ selectedIndex });
+  };
+  render() {
+    const { selectedIndex } = this.state;
+    return (
+      <Tabs selectedIndex={selectedIndex} onSelect={this.handleSelect}>
+        <TabItem>
+          <TabTitle>Tab 1 Name</TabTitle>
+          First tab Content
+        </TabItem>
+        <TabItem>
+          <TabTitle>Tab 2 Name</TabTitle>
+          Second Tab Content
+        </TabItem>
+      </Tabs>
+    );
+  }
+}
+```
+
+Under the hood we are using [react-tabs](https://github.com/reactjs/react-tabs)
--- a/packages/tabs/components/TabItem.tsx
+++ b/packages/tabs/components/TabItem.tsx
@@ -1,0 +1,15 @@
+import * as React from "react";
+
+export interface TabItemProps {
+  children?: React.ReactNode;
+}
+
+class TabItem extends React.PureComponent<TabItemProps, {}> {
+  public render() {
+    const { children } = this.props;
+
+    return children;
+  }
+}
+
+export default TabItem;
--- a/packages/tabs/components/Tabs.tsx
+++ b/packages/tabs/components/Tabs.tsx
@@ -1,0 +1,122 @@
+import * as React from "react";
+import { Tabs as ReactTabs, TabList, TabPanel } from "react-tabs";
+import { injectGlobal } from "emotion";
+
+import { TabItemProps } from "./TabItem";
+import { TabTitle } from "..";
+
+// Copy & paste from node_modules/react-tabs/style/react-tabs.css
+// This is needed to give the tabs a style
+/* tslint:disable */
+injectGlobal`
+.react-tabs {
+  -webkit-tap-highlight-color: transparent;
+}
+
+.react-tabs__tab-list {
+  border-bottom: 1px solid #aaa;
+  margin: 0 0 10px;
+  padding: 0;
+}
+
+.react-tabs__tab {
+  display: inline-block;
+  border: 1px solid transparent;
+  border-bottom: none;
+  bottom: -1px;
+  position: relative;
+  list-style: none;
+  padding: 6px 12px;
+  cursor: pointer;
+}
+
+.react-tabs__tab--selected {
+  background: #fff;
+  border-color: #aaa;
+  color: black;
+  border-radius: 5px 5px 0 0;
+}
+
+.react-tabs__tab--disabled {
+  color: GrayText;
+  cursor: default;
+}
+
+.react-tabs__tab:focus {
+  box-shadow: 0 0 5px hsl(208, 99%, 50%);
+  border-color: hsl(208, 99%, 50%);
+  outline: none;
+}
+
+.react-tabs__tab:focus:after {
+  content: "";
+  position: absolute;
+  height: 5px;
+  left: -4px;
+  right: -4px;
+  bottom: -5px;
+  background: #fff;
+}
+
+.react-tabs__tab-panel {
+  display: none;
+}
+
+.react-tabs__tab-panel--selected {
+  display: block;
+}
+`;
+/* tslint:enable */
+
+export type TabSelected = string;
+export interface TabsProps {
+  children: Array<React.ReactElement<TabItemProps>>;
+  selectedIndex: number;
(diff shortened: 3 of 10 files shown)


GeorgiSTodorov:
@Poltergeist , are you not merging to master on purpose?

GeorgiSTodorov:
It works, but the selection border is somewhat ugly and unnecessary.
Firefox
Peek 2019-05-27 15-57
Chrome
Peek 2019-05-27 15-57-2

packages/tabs/README.md
```diff
@@ -0,0 +1,37 @@
...
+    return {
+      TI1: `This is Tab Name 1 content`,
+      TI2: `This is Tab Name 2 content`
+    }[id];
+  }
+  render() {
+    const { active } = this.state;
+    return (
+      <Tabs selected={active}>
+        {{
```
DanielMSchmidt: I would like to challenge the API design.

Having an object as a child is not really a conventional React pattern, so I would like to understand why we should go with it. I also think having two distinct things in here shows that the modeling of the tab behavior is not clear.

[React Tabs](https://github.com/reactjs/react-tabs) uses the tabs of a specific kind as children, similar to your work on Table with the Columns. [React Native Tabs](https://github.com/react-native-community/react-native-tab-view) passes them in as a prop, which is also an option IMHO.

Coming from a DX perspective and assuming I would not want the tabs to be part of my routing (which makes sense as we could otherwise also use links) I would suggest an API like this:

```js
<TabableContent>
  <Tab title="A">
	My A content
  </Tab>
  <Tab title="B">
	My B content
  </Tab>
</TabableContent>
```

This would enable me to split up every tab as a separate component, encapsulating it's naming and content in a reusable way. My assumption about the implementation would be that only the content of the currently active tab is being rendered and that the entire selection part is hidden from me as an implementation detail.

What do you think? Which use case am I forgetting about?

mperrotti: I agree with Daniel - the children as an object is a little strange. I think the [React Tabs component API](https://github.com/reactjs/react-tabs#basic-example) is really nice.

Poltergeist: First I want to start of with an article about children compositions patterns [here](https://medium.com/@martin_hotell/react-children-composition-patterns-with-typescript-56dfc8923c64) 

I wanted to let the complete control over which tab is active and which content is shown to the overarching application. So my intend was to skip state in this component. But this is not explaining why I choose for an object as children instead of component like we did with the table. The example you are giving is removing the need for the object since we only have one type of children. Which still works, however we would need to cycle through all children to generate a list for the tabs and to pick the content for the active tab. 

The following could be an example I could agree on, although I want to raise that the pattern of using an object for named slots is not a bad approach, the question is if it is applicable here. For example with the `PageHeader` component the reasoning is a lot better.

```JSX
<TabableContent active="A"> {/* could also be a id*/}
  <Tab title="A">
	My A content
  </Tab>
  <Tab title="B">
	My B content
  </Tab>
</TabableContent>
```

The important question is what do we want to discuss here object as children (named slots) or if named slots are a good pattern to use for the tabs component? In order to move this PR forward I propose that we only discuss the named slot children for tabs as I feel the general pattern discussion is a bigger discussion where we should include more people from the team. 
Another discussion I would like to raise is if we want to keep state inside of the component shouldn't we just use `react-tabs`, the library mentioned by @danielMSchmidt ?

Poltergeist:
@mperrotti Is the WAI-ARIA spec a minimal requirement to the tabs component? If yes I would be happy if we could pair on the implementation. The ARIA spec is not my strong suit and you strike me to be more proficient.

packages/tabs/README.md
mperrotti: > In order to move this PR forward I propose that we only discuss the named slot children for tabs as I feel the general pattern discussion is a bigger discussion where we should include more people from the team.

I'm interested in this discussion. Could you schedule something or add it to the weekly sync?

---

> Another discussion I would like to raise is if we want to keep state inside of the component

I'm sure we could come up a good component API that would let us render tabs without state, but I'm not opposed to letting this component be stateful.

---

> Shouldn't we just use react-tabs, the library mentioned by @DanielMSchmidt ?

I'd be comfortable with that - it's pretty small and allows us to pass class names for styling

mperrotti:
@Poltergeist - since these components are so raw, I think we wait until the next iteration to make them accessible/keyboard-navigable.

Poltergeist:
Okay I will change this PR to use react-tabs instead.

packages/tabs/components/TabItem.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+
+export interface TabItemProps {
+  children?: React.ReactNode;
+  name: string;
```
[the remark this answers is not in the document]
Poltergeist: Hmm good question I would say we might not want to have the name as a prop but rather have a `TabTitel` component which is a Child of TabItem like

```html
<detail>
  <summary>Title</summary
  Example detail text
</detail>
```

packages/tabs/stories/Tabs.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+          Second Tab Content
+        </TabItem>
+      </Tabs>
+    );
+  }
+}
+
+storiesOf("Tabs", module)
+  .addDecorator(withReadme([readme]))
+  .add("default", () => <Example />);
```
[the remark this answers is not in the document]
Poltergeist: Because we needed state to provide a actual example, here.

packages/tabs/components/TabItem.tsx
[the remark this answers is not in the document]
Poltergeist: Should I change the component?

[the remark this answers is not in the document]
mperrotti: Good idea @Poltergeist - I like that pattern

packages/tabs/stories/Tabs.stories.tsx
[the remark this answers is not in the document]
mperrotti: Wooops, I missed the state 🙈

packages/tabs/components/Tabs.tsx
```diff
@@ -0,0 +1,119 @@
+import * as React from "react";
+import { Tabs as ReactTabs, TabList, TabPanel } from "react-tabs";
+import { injectGlobal } from "emotion";
+
+import { TabItemProps } from "./TabItem";
+import { TabTitle } from "..";
+
+// Copy & paste from node_modules/react-tabs/style/react-tabs.css
+// This is needed to give the tabs a style
+injectGlobal`
```
mperrotti: This is being flagged by tslint:
```
packages/tabs/components/Tabs.tsx[10, 1]: unused expression, expected an assignment or function call
```

Maybe we should make it a function and call it in the `render` method of `Tabs`?

Poltergeist: Na, I am going to ignore it for tslint calling this on render could be a huge performance loss. Since we might change the DOM. on every call.

GeorgiSTodorov:
It seems like if we have anything other than plain text inside `<TabItem>`, it doesn't work. I know that this is supposed to be a minimal component, but this seriously limits the usefulness since we can't style it.
Screenshot from 2019-06-17 08-35-43
Screenshot from 2019-06-17 08-35-55

PR #342, opened 2019-05-27
DCOS-54460 feat(card): add minimal card component

Poltergeist:
This provides a minimal card component which will create a div around children which has a border
and padding associated with it.

CLOSES DCOS-54460


packages/card/tests/Card.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+import { Card } from "../";
```
GeorgiSTodorov: Mixing local and non-local imports.

PR #331, opened 2019-05-21
DCOS-54456 feat(breadcrumb): add breadcrumb component

Poltergeist:
This is a minimal version for a breadcrumb component.

image

CLOSES DCOS-54456

PR #343, opened 2019-05-27
DCOS-54461 feat(pageheader): add minimal component implementation

Poltergeist:
This adds a minimal implementation for a page header.

CLOSES DCOS-54461

Depends on #331

image

### DIFF
--- a/packages/pageheader/README.md
+++ b/packages/pageheader/README.md
@@ -1,0 +1,1 @@
+# PageHeader
--- a/packages/pageheader/components/PageHeader.tsx
+++ b/packages/pageheader/components/PageHeader.tsx
@@ -1,0 +1,31 @@
+import * as React from "react";
+
+import { default as Breadcrumb } from "../../breadcrumb/components/Breadcrumb";
+
+import { flush } from "../../shared/styles/styleUtils";
+
+import { style } from "../style";
+
+export interface PageHeaderProps {
+  breadcrumbElements: React.ReactNodeArray;
+  actions?: Array<React.ReactElement<any>>;
+}
+
+class PageHeader extends React.PureComponent<PageHeaderProps, {}> {
+  public render() {
+    const { breadcrumbElements, actions = [] } = this.props;
+
+    return (
+      <div className={style}>
+        <Breadcrumb>{breadcrumbElements}</Breadcrumb>
+        <ul className={flush("all")}>
+          {actions.map(action => {
+            return <li key={`${action.key}`}>{action}</li>;
+          })}
+        </ul>
+      </div>
+    );
+  }
+}
+
+export default PageHeader;
--- a/packages/pageheader/stories/PageHeader.stories.tsx
+++ b/packages/pageheader/stories/PageHeader.stories.tsx
@@ -1,0 +1,25 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { withReadme } from "storybook-readme";
+
+import { PageHeader } from "../index";
+import { Clickable } from "../../clickable";
+
+const readme = require("../README.md");
+const action = () => alert("Action 1 triggered");
+
+storiesOf("PageHeader", module)
+  .addDecorator(withReadme([readme]))
+  .add("default", () => (
+    <PageHeader
+      breadcrumbElements={[
+        <span key="ElementOne">One</span>,
+        <span key="ElementTwo">Two</span>
+      ]}
+      actions={[
+        <Clickable key="Action1" action={action}>
+          <div>Action 1</div>
+        </Clickable>
+      ]}
+    />
+  ));
--- a/packages/pageheader/style.ts
+++ b/packages/pageheader/style.ts
@@ -1,0 +1,6 @@
+import { css } from "emotion";
+
+export const style = css`
+  display: grid;
+  grid-template-columns: repeat(2, 50%);
+`;
--- a/packages/pageheader/tests/PageHeader.test.tsx
+++ b/packages/pageheader/tests/PageHeader.test.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import { render } from "enzyme";
+import toJSON from "enzyme-to-json";
+
+import { PageHeader } from "../";
+import { Clickable } from "../../clickable";
+
+describe("PageHeader", () => {
+  it("default", () => {
+    const action = () => 1;
+    expect(
+      toJSON(
+        render(
+          <PageHeader
+            breadcrumbElements={[
+              <span key="One">One</span>,
+              <span key="Two">Two</span>
+            ]}
+            actions={[
+              <Clickable key="Action1" action={action}>
+                <div>Action 1</div>
+              </Clickable>
+            ]}
+          />
+        )
+      )
+    ).toMatchSnapshot();
+  });
+});
--- a/packages/pageheader/index.ts
+++ b/packages/pageheader/index.ts
@@ -1,0 +1,1 @@
+export { default as PageHeader } from "./components/PageHeader";
--- a/packages/pageheader/tests/__snapshots__/PageHeader.test.tsx.snap
+++ b/packages/pageheader/tests/__snapshots__/PageHeader.test.tsx.snap
@@ -1,0 +1,31 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PageHeader default 1`] = `
+<div
+  class="css-jiaax5"
+>
+  <div
+    class="css-ewxxaj"
+  >
+    <span>
+      One
+    </span>
+    <span>
+      Two
+    </span>
+  </div>
+  <ul
+    class="css-10pfh7o"
+  >
+    <li>
+      <div
+        class="css-1h5x3dy"
+        role="button"
+        tabindex="-1"
+      >
+        Action 1
+      </div>
+    </li>
+  </ul>
+</div>
+`;


packages/pageheader/style.ts
```diff
@@ -0,0 +1,3 @@
+import { css } from "emotion";
+
+export const style = css``;
```
GeorgiSTodorov: Are we exporting empty css?

packages/pageheader/tests/PageHeader.test.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+
+import { PageHeader } from "../";
```
GeorgiSTodorov: Mixing local and non-local imports.

packages/pageheader/components/PageHeader.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+
+import { BreadcrumbProps } from "../../breadcrumb/components/Breadcrumb";
+
+export interface PageHeaderProps {
+  children: {
```
mperrotti: Why did you choose to make `children` an object instead of just adding the props directly to the component?

packages/pageheader/components/PageHeader.tsx
```diff
@@ -0,0 +1,31 @@
...
+  };
+}
+
+class PageHeader extends React.PureComponent<PageHeaderProps, {}> {
+  public render() {
+    const { title, breadcrumb, actions = [] } = this.props.children;
+
+    return (
+      <div>
+        <h1>{title}</h1>
```
mperrotti: I've never seen a page header in dcos-ui that shows both a page title _and_ breadcrumbs. Is the idea to visually hide the `h1` but leave it rendered in the DOM for accessibility?

If not, I think we just just have one prop (either `breadcrumb` or `title`) that renders something on the left

Poltergeist: Oh I guess I misunderstood the notes I made. It is not title + breadcrumb + actions but just breadcrumbs + actions as the title is part of the breadcrumb. Will change.

packages/pageheader/components/PageHeader.tsx
Poltergeist: I have added it to children since I felt like the contents are children of the component not attributes which one would rather expose as an attribute property.

brandonc: I find it strange as well. This requires children to be an object, but the element itself is already a suitable object container. Why not just have title/breadcrumb/actions props? This limits the flexibility of the component in the future since children are usually understood to be react elements.

packages/pageheader/components/PageHeader.tsx
```diff
@@ -0,0 +1,31 @@
...
+        {breadcrumb}
+        <ul>
+          {actions.map((action, index) => (
+            <li key={`action${index}`}>{action}</li>
```
brandonc: Indexes are unsuitable as react keys. If an action is added and an action is deleted in the same update, React would render the wrong thing because the key is not unique to any particular action

packages/pageheader/README.md
```diff
@@ -0,0 +1,24 @@
+# PageHeader Name
```
brandonc: What, if any, relationship does this have with the AppChrome package? Is this component reusable by two apps? Is there a strategy to migrate DCOS UI to use this?

packages/pageheader/components/PageHeader.tsx
Poltergeist: I feared that, what would you propose as a better key?

packages/pageheader/README.md
Poltergeist: The App chrome is the surrounding frame. Where as this Component is just the heading. Which is containing the Breadcrumb and the actions.

image

This is the PageHeader in DC/OS ui

packages/pageheader/components/PageHeader.tsx
Poltergeist: Just so that we are all on the same page you would prefer to have something like that:
```JSX
<PageHeader breadcrumb={
  <Breadcrumb>
    <span>One</span>
    <span>Two</span>
  </Breadcrumb>
  } actions={[
  <Clickable key="Action1" action={action}>
    <div>Action 1</div>
  </Clickable>
]} />
```

I really feel the attribute props should not be used for ReactNodes but this is my opinion I am happy to discuss.

brandonc: That is it, yes. props is already a "named slot" I'm not sure what the benefit is in overtaking children for this purpose

packages/pageheader/README.md
brandonc: The basis of my question was to understand the strategy of having page header in ui-kit which appears to be app-specific. At this point, we have dcos-ui-like app kit components such as AppChrome in ui-kit and I'm wondering if this gets us further in that direction... Or if it is completely dedicated to Dumbledore

packages/pageheader/components/PageHeader.tsx
brandonc: You'd need to delegate this to callers...

`actions={[{ actionId: "myaction1", action: <Clickable /> }]}`

brandonc: Another option is to use a global counter so you always get a unique key, but this blows the whole point of keys and forces a dom change on update.

packages/pageheader/components/PageHeader.tsx
Poltergeist: Yes this is correct. I feel like the children approach is better maintainable and overall a better user experience. Please compare the above code with the following:
```JSX
<PageHeader>
  {{
    breadcrumb: (
      <Breadcrumb>
        <span>One</span>
        <span>Two</span>
      </Breadcrumb>
    ),
    actions: [
      <Clickable key="Action1" action={action}>
        <div>Action 1</div>
      </Clickable>
    ]
  }}
</PageHeader>
```

To me I can better see where the PageHeader begins and ends. The named slots are similar to distinguish so there is no drawback nor a potential gain. Also semantically I feel like the breadcrumb and the actions are a children property of the PageHeader and not an attribute of the PageHeader. 

Here is a third proposal would you say that this one is also incorrect or would you say that it is favourable over the attribute approach?

```
<PageHeader>
  <Breadcrumb>
    <span>One</span>
    <span>Two</span>
  </Breadcrumb>
  <ActionList>
    <Clickable key="Action1" action={action}>
      <div>Action 1</div>
    </Clickable>
  </ActionList>
</PageHeader>
```

packages/pageheader/README.md
Poltergeist: The PageHeader and all other components should be generic to all applications, although they should be themable to differentiate between different apps.

packages/pageheader/components/PageHeader.tsx
brandonc: I've heard some agreement that children "named slots" is a strange convention and I remain not convinced we should adopt it. You argued that you can better see where the PageHeader begins and ends, but this could also be taken to mean that an end tag is now necessary. Furthermore, there is a real drawback- you lose the flexibility of having normal, renderable children. I'm still fond of the props version, even as ReactNodes. Maybe we can get a fourth voice in here?

natmegs: I prefer the props pattern. IMO the children object pattern is unfamiliar and could be unnatural to others who are trying to consume this component.

Poltergeist: I tried to write up pros and cons, I figured that most of the pros are all applicable to putting the children into props as well. So it all comes down to wether or not we want to put children into the children prop or rather into other props. Although I want to raise that I think it is a huge anti pattern to put children into none children props. And if there are multiple `areas` within one component and one would not like to add the complexity of iterating through children to actually find the relevant areas. One might use this pattern to differentiate between different areas in a lot less complex manner. 
About the closing tag I don't feel that this is a drawback, since most HTML tags do need a closing tag this should be a widely used pattern. 
@natmegs I agree with your point that it is unfamiliar and that it feels unnatural if this would be the only component which is using this pattern which is correct right now. How would you feel about this pattern if this is a pattern we adopt for similar situations?

natmegs: I feel like props are already doing the same thing as the children object method, but without any additional boilerplate. My concern about it being unfamiliar/unnatural is not only just in comparison to other components in this library but more in terms of React patterns generally, so I don't think it would help to adopt this widely. I think it's commonly assumed that children will be a ReactNode, and for that reason it would be easy to misuse this component if it requires children that are an object of a particular shape.

Poltergeist: I thought a little more about that and discussed this with @DanielMSchmidt and @nLight today. I came to the conclusion that the object as a child pattern could also have performance issues as pointed out by @DanielMSchmidt, which is a good reason not to use it. But while talking about it we kind of came to the conclusion that passing ReactNode's as props is an anti pattern and most of the times when used we probably leave too much surface to the api. We probably want to pass something else for example for the `PageHeader` instead of expecting a `breadcrumb` and a list of `actions` we probably want to have a list of `breadcrumbElements` and a list of `actions`  which will change the type to look like this:
```TS
interface PageHeaderProps = {
  breadcrumbElements: React.ReactNode[];
  actions?: React.ReactNode[]; 
}
```

Also I agreed with @nLight that we should create guidelines on how to create components. Similar to `eslint` so that we have a team agreement on which patterns we use. This should unify on how components look. And this will erase eventually discussions about opinions. 

Just to mentioned it here at least we also came to the conclusion that if there is ever a need to pass ReactNode's as props one should use the composable pattern by exposing either pseudo components like `Column` or expect certain components e.g.:
```JS
<PageHeader>
  <Breadcrumb>
    <span>One</span>
    <span>Two</span>
  </Breadcrumb>
  <ActionList>
    <Clickable key="Action1" action={action}>
      <div>Action 1</div>
    </Clickable>
  </ActionList>
</PageHeader>
```

This will be reflected eventually in a guideline design doc.

packages/pageheader/components/PageHeader.tsx
mperrotti: Sorry @Poltergeist - that's my fault for misinterpreting how page headers work in DC/OS UI

packages/pageheader/components/PageHeader.tsx
mperrotti: > Also I agreed with @nLight that we should create guidelines on how to create components.

Keep me posted - I'd love to be a part of creating this guideline design doc.
---

> we kind of came to the conclusion that passing ReactNode's as props is an anti pattern and most of the times when used we probably leave too much surface to the api

100% agreed that it creates a big surface area. My favorite [blog post about component APIs](https://jxnblk.com/blog/defining-component-apis-in-react/) touches on this topic.

mperrotti:
I like the component API better, but I think my last comment got overlooked in all the excitement:

> I know this is supposed to be a very lo-fi representation of a page header, but could we at least introduce a little layout styling to put the breadcrumbs/title on the left, and the actions on the right?

Poltergeist:
Yes I overlooked the comment although this is something we can adjust later I will change this it in this PR

packages/pageheader/stories/PageHeader.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+          </Clickable>
+        ]}
+      />
+    );
+  }
+}
+
+storiesOf("PageHeader", module)
+  .addDecorator(withReadme([readme]))
+  .add("default", () => <Example />);
```
mperrotti: Why aren't we just adding the JSX right in the `.add`?

packages/pageheader/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+PageHeader documentation
+
+Example Implementation
+
+```JSX
```
mperrotti: What's the thinking behind adding example implementation in the README? Storybook can do that automatically.

mperrotti: What's the thinking behind adding example implementation in the README? Storybook can do that automatically.

packages/pageheader/stories/PageHeader.stories.tsx
Poltergeist: Because I missed replacing it, will adjust.

packages/pageheader/README.md
Poltergeist: This is also a left over from before. Storybook had problems with having objects as Children and I thought it would be helpful to have the component used in code at least once, will remove the documentation here.

packages/pageheader/components/PageHeader.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+
+import { default as Breadcrumb } from "../../breadcrumb/components/Breadcrumb";
+
+import { flush } from "../../shared/styles/styleUtils";
+
+import { style } from "../style";
+
+export interface PageHeaderProps {
+  breadcrumbElements: React.ReactNodeArray;
```
brandonc: Given the way that it is used, this could be a ReactNode, which is a superset containing ReactNodeArray

Poltergeist: I will check this to make sure the type is preventing what I wanted to prevent with it. Thanks for mentioning this. So the type `ReactNodeArray` is not equal to `ReactNode[]` ? That was what I wanted to achieve.

PR #344, opened 2019-05-31
DCOS-54072: themable color scheme

mperrotti:
The UIKitThemeProvider component takes a prop, `appTheme`, that applies a theme to ui-kit components that are children.

This component makes the theme properties available via emotion's `ThemeProvider` component or `withTheme` HoC, and as CSS custom properties written to the document root.
I chose to take advantage of CSS custom properties because:
- it's easy to set fallback colors in the case that a property has been excluded from the `appTheme`
- it avoids breaking changes in apps using ui-kit that are not wrapped in an emotion ThemeProvider
- styles can access theme properties in styles without passing the theme from a component to the imported style

## Testing:
You can use a Storybook decorator to wrap every Storybook story in a UIKitThemeProvider, and check that everything behaves as expected:
[decorators/themeDecorator.tsx](https://gist.github.com/mperrotti/b1f9cef41ca46230a4429875ddaa1ecb) (new file)
[.storybook/config.js](https://gist.github.com/mperrotti/0b93389d5cab7face254529c9561dcc4) (storybook config update)

Theme colors should be obeyed, and if a color property hasn't been set in the theme, it should fall back to the DC/OS colors as defined in `fallbacks.json`

## Tradeoffs
- A color scheme can't be fully changed without a page refresh or without un-mounting and re-mounting the components
- [getCSSVarVal issue](https://github.com/dcos-labs/ui-kit/pull/344#discussion_r290977324)
- The color functions still only support hex colors, so it's dangerous to theme using other color formats. We should definitely support other color formats in a future ticket
- Global styles don't unmount when the UIKitThemeProvider unmounts, but this is fixed in Emotion 10. This is a problem when you go to the UIKitThemeProvider story and then navigate to another story

**Reviewers:** I highly recommend taking advantage of the different commits


packages/badge/style.ts
```diff
@@ -1,56 +1,54 @@
...
+      getCSSVarValue(themeTextColorPrimary),
+      getCSSVarValue(themeTextColorPrimaryInverted)
+    )};
+  `;
+};
+
+const badgeAppearance = appearance => {
+  switch (appearance) {
+    case "default":
+      return badgeAppearanceStyle(greyLight);
```
natmegs: Does this mean that default appearance is not themeable?

packages/infobox/style.ts
```diff
@@ -5,49 +5,54 @@ import {
...
+      return css`
+        background-color: ${lighten(getCSSVarValue(themeError), 4)};
+        border-bottom-color: ${lighten(getCSSVarValue(themeError), 3)};
+        color: ${darken(getCSSVarValue(themeError), 4)};
+      `;
+    case "info":
+      return css`
+        background-color: ${blueLighten4};
+        border-bottom-color: ${blueLighten3};
+        color: ${blueDarken4};
```
natmegs: Similar Q here - are default and info states meant to be static/not themeable?

packages/dropdowMenu/components/DropdownMenuItem.tsx
```diff
@@ -19,6 +21,21 @@ export interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
 
 const DropdownMenuItem = (props: DropdownMenuItemProps) => {
   const { isActive, isSelected, index, listLength, disabled, ...other } = props;
+  // Importing these was causing issues with `getCSSVarValue` where it would
```
mperrotti: I'm having an issue where the custom properties aren't set on :root yet when I try and use `getCSSVarVal` outside of a component.

It works if I use it inside a component, or if the imported Emotion style is called as a function from the component.

I'd appreciate some help on tracking down the root cause of this :)

packages/badge/style.ts
mperrotti: Wooops, this should have been `themeBgNeutral`. Good catch!

packages/infobox/style.ts
mperrotti: We're getting rid of this one because the design team only wants to use the grey one, so I didn't bother theming it.

I should probably add a comment and remove it from Storybook

PR #346, opened 2019-06-17
DCOS-54889: add a donut chart component

mperrotti:
This is just a rough draft of this component meant to be used as a placeholder. We will enhance the component as needed.

Using this donut chart component removes the need for [react-donut-graph](https://github.com/CreativeBuilds/react-donut-graph/). Then, we can will resolve the issue where all text in apps that use react-donut-chart becomes centered.

## Screenshots:
Screen Shot 2019-06-17 at 11 37 27 AM


packages/donutChart/components/DonutChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+class DonutChart extends React.PureComponent<DonutChartProps, {}> {
+  public render() {
+    const { data, label, text } = this.props;
+    const strokeWidth = 2;
+    const radius = 100 / (Math.PI * 2);
+    const diameter = radius * 2 + strokeWidth;
+    const circleCenter = diameter / 2;
+
+    const circleEls = (data: DonutChartDatum[]) =>
```
natmegs: Small nit: could this just be `const circleEls = data.map ...` instead of assigning to a fn and calling it?

mperrotti: Yea, that makes way more sense 🤦‍♂

PR #348, opened 2019-06-20
[spike]DCOS_OSS-5217: improve Storybook documentation

mperrotti:
I made some improvements to component and color documentation:
- improved how we show the source of stories
- cleaned up colors table and made all color names fully visible with a Tooltip
- made form component stories more consistent

One of the improvements to showing story source is to use Storybook's [storysource addon](https://www.npmjs.com/package/@storybook/addon-storysource). This addon moves the story source into the addons panel at the bottom of Storybook. If any reviewers don't like using this addon, I can put the story source inline with the component demo where it used to live.

### Improved story source
**Before** (useful code example is obscured by a helper component)
Screen Shot 2019-06-20 at 4 07 35 PM

**After** (useful code examples are visible, and the full story source can be browsed)
StorySourceDemo_smaller

### Cleaned up color grid
**Before**
ColorTooltipDemo_before

**After**
ColorTooltipDemo


danielone:
@mperrotti 
awesome stuff!!!
I was going to file a jira actually from some needs here!

- Is it possible to see the colors values from such in some manner ? (right on the tile, sidebar if click, on hover, etc..) 
- find out easily what hex value is and be able to copy and paste. hsla nice too especially if capture transparencies in future but that may be in future.

image

image

mperrotti:
@DanielMSchmidt - definitely possible. File a ticket and I'll tackle that in the next spike day.

It would also be cool to include color contrast accessibility info :)

PR #349, opened 2019-06-20
[spike]DCOS-55570: distribute typed design tokens

mperrotti:
The JS distribution of design tokens is now typed and compiled with Typescript.

Bonus:
You can now import design tokens into your app like this:
```
import { designTokens } from "@dcos/ui-kit";

const { red, blue, green } = designTokens;
```

I tested these changes in `dcos-ui` and `kommander`


packages/design-tokens/scripts/formats.js
```diff
@@ -23,10 +23,10 @@ const colorsForStyleguide = {
 };
 
 const commonJS = {
-  name: "javascript/commonJS",
+  name: "typescript/es6",
   formatter: dictionary =>
     `${dateHeader()}\n${dictionary.allProperties
-      .map(prop => `exports.${prop.name} = '${prop.value}';`)
+      .map(prop => `export const ${prop.name}: string = "${prop.value}";`)
```
mperrotti: Setting specific types for each category of design tokens felt like overkill, so I'm just using the `string` type for everything.

I can be more specific if anybody disagrees

GeorgiSTodorov:
Running the linter locally shows no errors, but we do have linter errors in the CI.

PR #351, opened 2019-06-26
chore: add top level package export for configuration map

natmegs:
^^ What the PR title says 😄

### DIFF
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -25,6 +25,7 @@ export {
   DangerDropdownButton
 } from "./button";
 export { CheckboxInput } from "./checkboxInput";
+export { ConfigurationMap } from "./configurationmap";
 export { DonutChart } from "./donutChart";
 export { Dropdownable } from "./dropdownable";
 export { Icon } from "./icon";


mperrotti:
Woops, I should do this in https://github.com/dcos-labs/ui-kit/pull/352

Poltergeist:
Do we want to support the global imports? Last I heard we haven't solved the tree shaking issue and we should stick to deep imports. that's why I haven't added the global exports in the first place. @mperrotti

Poltergeist:
How can a global export be blocking a Pr? There is always the way of deep imports.

natmegs:
@Poltergeist  The only import path I could find was from `"@dcos/ui-kit/packages/configurationmap"`, which is the TS file directly. This is causing a bunch of TS errors originating from ui-kit in node_modules. For other deep imports in ui-kit we use the `dist` folder (ie `"@dcos/ui-kit/dist/packages/design-tokens/build/js/designTokens"`) but configurationmap is not in this folder.

Poltergeist:
Totally I forgot that we only compile stuff which is linked in the index.ts file. Thanks for explaining.

PR #352, opened 2019-06-26
DCOS-54898: finished ConfigurationMap, added HashMap

mperrotti:
## Screenshots
Screen Shot 2019-06-25 at 9 08 57 PM

### DIFF
--- a/packages/configurationmap/components/ConfigurationMapHeading.tsx
+++ b/packages/configurationmap/components/ConfigurationMapHeading.tsx
@@ -1,0 +1,44 @@
+import * as React from "react";
+import { cx } from "emotion";
+import { padding, border } from "../../shared/styles/styleUtils";
+import { configurationMapHeadingStyle } from "../style";
+
+interface ConfigurationMapHeadingProps {
+  /**
+   * What to render as the heading content
+   */
+  children: React.ReactNode;
+  /**
+   * Priority of the heading. Numbers map to <h1> through <h6>
+   */
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+}
+
+const ConfigurationMapHeading: React.StatelessComponent<
+  ConfigurationMapHeadingProps
+> = props => {
+  const { children, headingLevel = 1 } = props;
+  // TODO: when we have our typographic scale and component for headings,
+  // use those header components instead of HTML headings
+  const Heading: keyof React.ReactHTML = `h${headingLevel}` as
+    | "h1"
+    | "h2"
+    | "h3"
+    | "h4"
+    | "h5"
+    | "h6";
+
+  return (
+    <Heading
+      className={cx(
+        configurationMapHeadingStyle,
+        padding("bottom", "xs"),
+        border("bottom")
+      )}
+    >
+      {children}
+    </Heading>
+  );
+};
+
+export default ConfigurationMapHeading;
--- a/packages/configurationmap/components/ConfigurationMapRow.tsx
+++ b/packages/configurationmap/components/ConfigurationMapRow.tsx
@@ -1,0 +1,38 @@
+import { configurationMapRow } from "../style";
+import { css } from "emotion";
+import styled from "react-emotion";
+import {
+  flex,
+  padding,
+  border,
+  visuallyHidden,
+  undoVisuallyHidden
+} from "../../shared/styles/styleUtils";
+import { rowActionStaticClassname } from "./ConfigurationMapRowAction";
+
+interface ConfigurationMapRowActionProps {
+  children: React.ReactNode;
+  onlyShowActionOnHover?: boolean;
+}
+
+const showActionOnHoverStyle = (props: ConfigurationMapRowActionProps) => {
+  if (props.onlyShowActionOnHover) {
+    return css`
+      .${rowActionStaticClassname} {
+        ${visuallyHidden};
+      }
+
+      &:hover .${rowActionStaticClassname} {
+        ${undoVisuallyHidden};
+      }
+    `;
+  }
+};
+
+export default styled("div")`
+  ${configurationMapRow};
+  ${showActionOnHoverStyle};
+  ${border("bottom")};
+  ${flex()};
+  ${padding("vert", "xs")};
+`;
--- a/packages/configurationmap/components/ConfigurationMapRowAction.tsx
+++ b/packages/configurationmap/components/ConfigurationMapRowAction.tsx
@@ -1,0 +1,31 @@
+import * as React from "react";
+import SecondaryButton from "../../button/components/SecondaryButton";
+import { cx } from "react-emotion";
+import { flexItem } from "../../shared/styles/styleUtils";
+
+interface ConfigurationMapRowActionProps {
+  /**
+   * What to render as the action content
+   */
+  children: React.ReactNode;
+  /**
+   * Click handler for the action
+   */
+  onClick: (event?: React.SyntheticEvent<HTMLElement>) => void;
+}
+
+export const rowActionStaticClassname = "static_configurationMapRowAction";
+
+const ConfigurationMapRowAction: React.StatelessComponent<
+  ConfigurationMapRowActionProps
+> = props => {
+  const { onClick, children } = props;
+
+  return (
+    <span className={cx(rowActionStaticClassname, flexItem("shrink"))}>
+      <SecondaryButton onClick={onClick}>{children}</SecondaryButton>
+    </span>
+  );
+};
+
+export default ConfigurationMapRowAction;
--- a/packages/configurationmap/components/ConfigurationMapValueWithDefault.tsx
+++ b/packages/configurationmap/components/ConfigurationMapValueWithDefault.tsx
@@ -1,0 +1,18 @@
+import * as React from "react";
+import ConfigurationMapValue from "./ConfigurationMapValue";
+
+interface ConfigurationMapValueWithDefaultProps {
+  /**
+   * The value to display
+   */
+  value?: React.ReactNode;
+}
+const ConfigurationMapValueWithDefault: React.StatelessComponent<
+  ConfigurationMapValueWithDefaultProps
+> = props => {
+  const { value } = props;
+
+  return <ConfigurationMapValue>{value || "—"}</ConfigurationMapValue>;
+};
+
+export default ConfigurationMapValueWithDefault;
--- a/packages/configurationmap/components/HashMap.tsx
+++ b/packages/configurationmap/components/HashMap.tsx
@@ -1,0 +1,150 @@
+import * as React from "react";
+
+import ConfigurationMapSection from "./ConfigurationMapSection";
+import ConfigurationMapHeading from "./ConfigurationMapHeading";
+import ConfigurationMapRow from "./ConfigurationMapRow";
+import ConfigurationMapLabel from "./ConfigurationMapLabel";
+import ConfigurationMapValue from "./ConfigurationMapValue";
+
+const METHODS_TO_BIND = ["formatValue", "isHashMap"];
+
+type headingLevel = 1 | 2 | 3 | 4 | 5 | 6;
+
+interface HashMapProps {
+  /**
+   * Content to render if no value is available
+   */
+  defaultValue?: string;
+  /**
+   * The data to render. Keys are rendered as labels, values are rendered as the value
+   */
+  hash: { [key: string]: React.ReactNode };
+  /**
+   * Priority of the heading that the `headline` prop renders in. Numbers map to <h1> through <h6>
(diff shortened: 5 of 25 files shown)


mperrotti:
@Poltergeist:

> why we are introducing a breaking change to the API. Also why are we breaking the API just shortly after we introduced it in #332

That's a great point. We did _just_ introduce this component. I changed it for two reasons:
1. I thought this component API, taken from the `ConfigurationMap` in DC/OS, was closer to what we want our component APIs to look like
2. Aligning the component API to what's in DC/OS would make it easier to use this component in DC/OS
3. I wanted to support row actions in the configuration map, and I couldn't think of a great way to do that with the `data` prop as you had it before.

---

If you or other reviewers think think these changes are too much too soon, I'll gladly change the component API back to what it was.

If you and other reviewers think it's ok to move forward with these changes, I'll make the appropriate changes to my commit message.

GeorgiSTodorov:
@Poltergeist , good catch on the breaking changes.
However, I think it's better to change the commit message and move forward with the current changes.
Also, @mperrotti , we have a conflicting file.

packages/configurationmap/index.ts
```diff
@@ -1 +1,23 @@
 export { default as ConfigurationMap } from "./components/ConfigurationMap";
+export {
+  default as ConfigurationMapSection
```
pierrebeitz: i like skipping the prefix of `ConfigurationMap*` to be able to either 

```typescript
import { Section } from "components/ConfigurationMap"

// OR

import * as CMap from "components/ConfigurationMap"

render(
  <CMap.Section />
)
```

but that's personal preference. no action item here.

packages/configurationmap/components/HashMap.tsx
```diff
@@ -0,0 +1,152 @@
...
+   */
+  emptyValue?: string;
+  /**
+   * The data to render. Keys are rendered as labels, values are rendered as the value
+   */
+  hash: { [key: string]: React.ReactNode };
+  /**
+   * Priority of the heading that the `headline` prop renders in. Numbers map to <h1> through <h6>
+   */
+  headingLevel?: headingLevel;
```
pierrebeitz: is there a concrete need for `headingLevel` already?

my gut says that the library should decide on this most or all of the time.

packages/configurationmap/components/HashMap.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    if (!value && this.props.emptyValue) {
+      value = this.props.emptyValue;
+    }
+
+    return value;
+  }
+
+  isHashMap(value) {
```
pierrebeitz: sidenote: this "method" does not use `this` and thus does not need to be a method.

no action item, please just close this comment.

packages/configurationmap/tests/__snapshots__/HashMap.test.tsx.snap
```diff
@@ -0,0 +1,141 @@
...
+      class="css-1pld8at"
+    >
+      bar
+    </dd>
+  </div>
+</dl>
+`;
+
+exports[`HashMap renders with nested object hash 1`] = `
+<dl
```
pierrebeitz: is that markup here the way we want it?

the double-`dl` makes me suspicious :)
```
dl 
  dl 
    h2 foo
    div
      dt bar
      dd baz
```

packages/configurationmap/stories/HashMap.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+      hash={{
+        name: "Jane Doe",
+        role: "UX Designer",
+        city: "San Francisco",
+        state: "CA",
+        zipcode: 95125
+      }}
+    />
+  ))
+  .add("with nested objects", () => (
```
natmegs: I think this is overwritten below

packages/configurationmap/stories/HashMap.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    <HashMap
+      hash={{
+        name: "Jane Doe",
+        role: "UX Designer",
+        city: "San Francisco",
+        state: "CA",
+        zipcode: 95125,
+        Favorites: {
```
natmegs: Is there a way to specify `renderKeys` for nested keys?

packages/configurationmap/components/HashMap.tsx
mperrotti: I've seen a bunch of cases in `dcos-ui` where we manually set the `headingLevel`, so I figured I'd keep that here.

packages/configurationmap/tests/__snapshots__/HashMap.test.tsx.snap
mperrotti: I agree, this is a bit strange, but I can't find anything in the spec that says it's not ok: https://www.w3.org/TR/html401/struct/lists.html#h-10.3.

I'm open to suggestions for how we might change the parent `dl` though if you have any ideas :)

packages/configurationmap/stories/HashMap.stories.tsx
mperrotti: Woops, you're right. Good catch 😅

packages/configurationmap/tests/__snapshots__/HashMap.test.tsx.snap
pierrebeitz: if it's okay for you, it's okay for me! :)

packages/configurationmap/stories/HashMap.stories.tsx
mperrotti: No, it only goes one level deep. Do you think this can be a follow-up ticket? Or changed in this PR?

Poltergeist:
I still intend to review this thoroughly, please don't merge yet.

packages/configurationmap/stories/HashMap.stories.tsx
natmegs: Follow up would be fine - it's probably not a terribly common use case!

packages/table/README.md
```diff
@@ -4,13 +4,6 @@
 
 Use tables to display information in a way that’s easy to scan, so that user can look for patterns and insights. Users will use tables to troubleshoot, manage and monitor their systems.
 
-## Types of Tables
```
Poltergeist: I think this has been removed by accident?

packages/typeahead/tests/__snapshots__/Typeahead.test.tsx.snap
```diff
@@ -1,6 +1,28 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Typeahead renders 1`] = `
+.emotion-3 {
```
Poltergeist: Why are we updating the typeahead snapshots? Did we touch anything here?

packages/table/README.md
mperrotti: Nope, the "detail table" describes a configuration map, so it felt weird to have it documented in here

packages/typeahead/tests/__snapshots__/Typeahead.test.tsx.snap
mperrotti: We didn't. I have no idea why the snapshots changed - they styles look the same

packages/configurationmap/stories/ConfigurationMap.stories.tsx
```diff
@@ -1,14 +1,171 @@
...
+          <ConfigurationMapValue>San Francisco</ConfigurationMapValue>
+          <ConfigurationMapRowAction>Action</ConfigurationMapRowAction>
+        </ConfigurationMapRow>
+      </ConfigurationMapSection>
+    </ConfigurationMap>
+  ))
+  .add("with row action shown on hover", () => (
+    <ConfigurationMap>
+      <ConfigurationMapSection>
+        <ConfigurationMapRow showActionOnHover={true}>
```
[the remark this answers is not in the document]
mperrotti: Technically it is, but our linter yells at me for not explicitly passing a boolean value

packages/configurationmap/components/ConfigurationMapValueWithDefault.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+interface ConfigurationMapValueWithDefaultProps {
+  /**
+   * The value to display
+   */
+  value?: React.ReactNode;
+  /**
+   * Content to render if no value is available
+   */
+  defaultValue?: React.ReactNode;
```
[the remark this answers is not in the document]
mperrotti: I'm not certain, but I think sometimes we use `N/A`, and sometimes we use `–`.
I'm ok with not allowing this to be configurable for now, and changing it if we need to

packages/configurationmap/components/ConfigurationMapRowAction.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface ConfigurationMapRowActionProps {
+  /**
+   * What to render as the action content
+   */
+  children: React.ReactNode;
+  /**
+   * Click handler for the action
+   */
+  onClick?: (event?: React.SyntheticEvent<HTMLElement>) => void;
```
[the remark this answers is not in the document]
mperrotti: Good point - it should probably be required

packages/configurationmap/components/ConfigurationMapRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+  padding,
+  border,
+  visuallyHidden,
+  undoVisuallyHidden
+} from "../../shared/styles/styleUtils";
+import { rowActionStaticClassname } from "./ConfigurationMapRowAction";
+
+interface ConfigurationMapRowActionProps {
+  children: React.ReactNode;
+  showActionOnHover?: boolean;
```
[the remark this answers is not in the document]
mperrotti: We could have an action that's always shown, or we could have an action that's only shown on hover

packages/configurationmap/components/ConfigurationMapHeading.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface ConfigurationMapHeadingProps {
+  /**
+   * What to render as the heading content
+   */
+  children: React.ReactNode;
+  /**
+   * Priority of the heading. Numbers map to <h1> through <h6>
+   */
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
```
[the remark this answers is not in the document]
mperrotti: > 1. Do we really need 6 level deep headings
A `HashMap` with nested objects increases it's heading level (up to `6`) with each level of nesting

> 2. How this is implemented we could have a headingLevel 7 in the default value since that one does not get type checked if we want to have a default value we need a implementation like this:

I don't think I'm following what you're saying here. Maybe we could hop on a Zoom and talk more?

> 3. I would prefer if we would expose ConfigurationMapHeading1... ConfigurationMapHeading6 as different components instead of 1 flexible component.

Why do you prefer that?

packages/configurationmap/components/ConfigurationMapRow.tsx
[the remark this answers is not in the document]
Poltergeist: Okay, gotcha, I propose to rename the prop to `onlyShowActionOnHover` to remove the  confusion. My assumption was that the normal case is that the action is hidden and it should get shown on hover.

packages/configurationmap/components/ConfigurationMapHeading.tsx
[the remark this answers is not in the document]
Poltergeist: Let's talk about this offline it is also not blocking.

packages/configurationmap/components/ConfigurationMapValueWithDefault.tsx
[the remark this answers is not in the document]
Poltergeist: Let's keep on using it. Although I would prefer if we could bring this to a better consistency by thinking about the different cases and maybe generate multiple components. For each case.

packages/configurationmap/components/ConfigurationMapRow.tsx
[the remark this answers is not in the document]
Poltergeist: Okay, gotcha, I propose to rename the prop to `onlyShowActionOnHover` to remove the  confusion. My assumption was that the normal case is that the action is hidden and it should get shown on hover.

GeorgiSTodorov:
Is it expected that the prop types look weird?
On master:
Screenshot from 2019-07-09 13-05-43

Here:
Screenshot from 2019-07-09 13-06-41

mperrotti:
@GeorgiSTodorov - that's an easy fix in Storybook. I'll take care of it before we merge :)

PR #354, opened 2019-06-27
DCOS-55837: Add an avatar component

mperrotti:
## Screenshots
Screen Shot 2019-06-27 at 1 25 10 PM


GeorgiSTodorov:
Wouldn't it be better to have a default icon instead of an empty one when the source is empty?

mperrotti:
@GeorgiSTodorov - a placeholder would be nice, I just don't know what it should be. We can always add it later

PR #357, opened 2019-07-03
DCOS-56024: feat(tabs): add styling

Poltergeist:
This adds the styling to tabs.

image

CLOSES DCOS-56024

### DIFF
--- a/packages/tabs/components/Tabs.tsx
+++ b/packages/tabs/components/Tabs.tsx
@@ -4,8 +4,17 @@ import { injectGlobal } from "emotion";
 
 import { TabItemProps } from "./TabItem";
 import { TabTitle } from "..";
+import {
+  themeBorder,
+  fontWeightMedium,
+  themeTextColorPrimary,
+  themeBrandPrimary,
+  themeBgHover
+} from "../../design-tokens/build/js/designTokens";
+import { padding, margin } from "../../shared/styles/styleUtils";
 
 // Copy & paste from node_modules/react-tabs/style/react-tabs.css
+// Also changed to better fit the ui kit styles.
 // This is needed to give the tabs a style
 /* tslint:disable */
 injectGlobal`
@@ -14,27 +23,43 @@ injectGlobal`
 }
 
 .react-tabs__tab-list {
-  border-bottom: 1px solid #aaa;
-  margin: 0 0 10px;
-  padding: 0;
+  border-bottom: 1px solid ${themeBorder};
+  ${margin("bottom", "s")}
+  ${padding("vert", "none")}
+  ${padding("horiz", "l")}
 }
 
 .react-tabs__tab {
   display: inline-block;
-  border: 1px solid transparent;
-  border-bottom: none;
-  bottom: -1px;
   position: relative;
   list-style: none;
-  padding: 6px 12px;
+  ${padding("vert", "s")}
+  ${margin("right", "s")}
   cursor: pointer;
+  font-weight: ${fontWeightMedium};
+  color: ${themeTextColorPrimary};
+  &:after{
+    content: "";
+    height: 2px;
+    position: absolute;
+    width: 100%;
+    bottom: 0;
+    left: 0;
+    background: ${themeBrandPrimary};
+    display:none;
+  }
+}
+
+.react-tabs__tab:focus {
+  outline: none;
+  background: ${themeBgHover}
 }
 
 .react-tabs__tab--selected {
-  background: #fff;
-  border-color: #aaa;
-  color: black;
-  border-radius: 5px 5px 0 0;
+  &:after{
+    display: block;
+  }
+  color: ${themeBrandPrimary};
 }
 
 .react-tabs__tab--disabled {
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
@GeorgiSTodorov I can not see this border, neither on Chrome nor on Firefox. Did you do anything special to get it?

GeorgiSTodorov:
@DanielMSchmidt , I saw it accidentally, I think it appeared by clicking of the border of the tab rectangle. If you still can't reproduce it, I will create a gif.

Poltergeist:
The border is the focus around the clickable area. Although we don't have it in DC/OS ui it is still considered to be a best practice and also something we need to address in terms of a11y so  I added the focus.

GeorgiSTodorov:
@Poltergeist , I think it appears in an unusual way - only when clicking the border. I think most users usually click the text, not the border.

Poltergeist:
It should only appear on the active tab and it focuses when you click the second time.

GeorgiSTodorov:
Why would you click the active tab a second time?

Poltergeist:
Well I don't know why it is implemented like this but this is how the browser behaves. I can see us eventually only showing the focus if the user uses the keyboard.

GeorgiSTodorov:
Since we are adding a border, can we also change the way it shows?

Poltergeist:
@GeorgiSTodorov Okay I removed the focus style for now.

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +21,60 @@ injectGlobal`
 }
 
 .react-tabs__tab-list {
```
mperrotti: Why not pass our own classname/s generated by emotion?

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +21,60 @@ injectGlobal`
...
+    content: "";
+    height: 2px;
+    width: 100%;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    background: ${purple};
+    display: block;
+  }
+  color: ${purple};
```
mperrotti: Use `themeBrandPrimary` instead

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +21,60 @@ injectGlobal`
...
-  border-bottom: none;
-  bottom: -1px;
   position: relative;
   list-style: none;
-  padding: 6px 12px;
+  ${padding("vert", "s")}
+  ${margin("right", "s")}
   cursor: pointer;
+  font-weight: 600;
+  color: ${greyDarkLighten3};
```
mperrotti: Use `themeTextColorPrimary` instead

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +21,60 @@ injectGlobal`
...
-  border: 1px solid transparent;
-  border-bottom: none;
-  bottom: -1px;
   position: relative;
   list-style: none;
-  padding: 6px 12px;
+  ${padding("vert", "s")}
+  ${margin("right", "s")}
   cursor: pointer;
+  font-weight: 600;
```
mperrotti: Use `fontWeightMedium` instead

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +21,60 @@ injectGlobal`
 }
 
 .react-tabs__tab-list {
-  border-bottom: 1px solid #aaa;
-  margin: 0 0 10px;
-  padding: 0;
+  border-bottom: 1px solid ${greyLight};
```
mperrotti: Use `themeBorder` instead

packages/tabs/components/Tabs.tsx
Poltergeist: Because react tabs has not that much of configuration possibilities.

packages/tabs/components/Tabs.tsx
```diff
@@ -14,52 +23,61 @@ injectGlobal`
...
   position: relative;
   list-style: none;
-  padding: 6px 12px;
+  ${padding("vert", "s")}
+  ${margin("right", "s")}
   cursor: pointer;
+  font-weight: ${fontWeightMedium};
+  color: ${themeTextColorPrimary};
+  &:after{
+    content: "";
```
pierrebeitz: why specify everything here and then hide it with `display: none;`?

i'd vow to either 

```
.base {
  all: the;
  sty: les;
  display: none;
}

.base--selected {
  display: block;
}
```

or 
```
// skip .base-styling

.base--selected {
  all: the;
  sty: les;
  display: block;
}
```

GeorgiSTodorov:
@pierrebeitz , is your feedback blocking? If not, I will merge.

packages/tabs/components/Tabs.tsx
Poltergeist: I'll adjust.

danielone:
@mperrotti @Poltergeist this component deserves another pass as related to keyboard focus, perhaps in grander scheme of keyboard focus?

Agree It is important to show focus of a tab especially when using keyboard navigation. Otherwise no affordance to user what has been selected
Example: https://lightningdesignsystem.com/components/tabs/#Keyboard-Interactions

Taking action item to file backlog item

PR #358, opened 2019-07-04
Setup work to support style utility components

mperrotti:
There are no visual changes, these are just changes to lower level code (like design tokens) to support the other style utility PRs.

Just check that tests still pass and the code looks good. See other style utility PRs for additional context

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -987,7 +987,7 @@
       "dependencies": {
         "chalk": {
           "version": "2.3.1",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.3.1.tgz",
+          "resolved": "http://registry.npmjs.org/chalk/-/chalk-2.3.1.tgz",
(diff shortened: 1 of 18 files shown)


GeorgiSTodorov:
`npm run dist` is failing locally, as well as in CI.
Screenshot from 2019-07-04 10-00-49

package-lock.json
```diff
@@ -10437,9 +10437,9 @@
       }
     },
     "csstype": {
-      "version": "2.6.3",
```
GeorgiSTodorov: Why did we have csstype in package-lock if we are only adding it now?

pierrebeitz: it's a dependency of `@emotion`!

GeorgiSTodorov: That makes sense, I guess.

pierrebeitz: here's a boiled down section of the `package-lock.json` that hopefully makes some sense:

```
```

PR #362, opened 2019-07-04
DCOS-43353: add vertical list components

mperrotti:
Adds components for rendering vertical lists

## Screenshots
Screen Shot 2019-07-04 at 1 46 00 AM

Screen Shot 2019-07-04 at 1 46 25 AM


packages/list/components/List.tsx
```diff
@@ -0,0 +1,35 @@
...
+    | Array<React.ReactElement<HTMLLIElement>>
+    | React.ReactElement<HTMLLIElement>;
+  tag: "ul" | "ol";
+}
+
+interface ListProps extends SharedListProps {
+  /**
+   * The style of the text bullets. See https://www.w3.org/TR/CSS21/generate.html#list-style for more info
+   */
+  markerStyle?: React.CSSProperties["listStyleType"];
```
Poltergeist: Shouldn't this be defined by ui-kit for consistency?

packages/list/components/List.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const List = (props: ListProps) => {
+  const { children, markerStyle, tag } = props;
+  const StyledList = styled(tag)`
```
Poltergeist: Like before this might belong outside of the function scope.

packages/list/components/BorderedList.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
+import { borderedList } from "../style";
+import styled from "react-emotion";
+import { SharedListProps } from "./List";
+import { listReset } from "../../shared/styles/styleUtils";
+
+const BorderedList = (props: SharedListProps) => {
```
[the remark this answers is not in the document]
mperrotti: Yes, this seems better. I'll update other PRs with this too

packages/list/components/BorderedList.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const { tag, children } = props;
+  const StyledList = styled(tag)`
```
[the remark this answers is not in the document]
mperrotti: Thanks for reminding me of that! I'm used to looking for it in the `render` method, but it still applies here

packages/list/components/List.tsx
mperrotti: Yea, we could limit which values we accept in ui-kit

PR #363, opened 2019-07-05
fix(index): add missing components to build

Poltergeist:
This adds components to the index of all packages which will make them build since
`packages/index.ts` is our entry point of typescript compiler.


GeorgiSTodorov:
@Poltergeist , build is failing.

packages/index.ts
```diff
@@ -53,6 +56,7 @@ export {
   SortableHeaderCell,
   NumberCell
 } from "./table";
+export { TabItem, TabTitle, Tabs } from "./tabs";
```
mperrotti: Alphabetized! 😍

Poltergeist: Almost `Tabs` need to be before `TabTitle`.

PR #364, opened 2019-07-09
[spike] refactor(breadcrumb): refactor breadcrumb to reduce iterations

Poltergeist:


pierrebeitz:
are we talking about the loop-iterations here? if so, have we benchmarked stuff and identified that there's a need to optimize the performance of that? 

asking as i think we're sacrificing on readability here a bit.

GeorgiSTodorov:
@pierrebeitz , do you want to block this?

pierrebeitz:
i'm sorry for creating noise in a PR that already has 2 approvals but i really think we should optimize for readability in this case (i'm *pretty* sure we won't have 100000 items in a breadcrumb). 

an `intersperse`-helper (can be easily reused for all sorts of things) was introduced and the way we set our padding has been changed. you might also want to read https://github.com/dcos-labs/ui-kit/pull/364/commits/bca2e57905601bc93b6fa862b7677ec4035c3d23.

sorry for the noise and feel free to just delete that commit again and merge the PR in its previous state if you don't think my arguments make sense.

PR #368, opened 2019-07-10
feat(modal): change closeText to a optional prop on dialog with footer

Poltergeist:
This changes the closeText prop to be optional on the DialogModalWithFooter component to cover the
rare case were we want to show a informational message.

### DIFF
--- a/packages/modal/components/DialogModalWithFooter.tsx
+++ b/packages/modal/components/DialogModalWithFooter.tsx
@@ -1,14 +1,15 @@
 import * as React from "react";
-import DialogModal, { DialogModalProps } from "./DialogModal";
-import SecondaryButton from "../../button/components/SecondaryButton";
+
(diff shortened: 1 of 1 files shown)


packages/modal/components/DialogModalWithFooter.tsx
```diff
@@ -22,9 +23,11 @@ class DialogModalWithFooter extends React.PureComponent<
       <DialogModal
         footerContent={
           <div className={flex({ align: "center", justify: "space-between" })}>
-            <div className={flexItem("shrink")}>
-              <SecondaryButton onClick={onClose}>{closeText}</SecondaryButton>
-            </div>
+            {closeText !== "" || closeText != null ? (
```
mperrotti: Non-blocking nitpick: wouldn't `closeText && ()` work just as well because an empty string is falsy?

PR #370, opened 2019-07-11
Fix Modal height regression

mperrotti:
## Screenshots
**Before**
Screen Shot 2019-07-11 at 6 33 11 PM

**After**
Screen Shot 2019-07-11 at 6 30 30 PM


packages/modal/components/ModalBase.tsx
```diff
@@ -82,15 +81,10 @@ class ModalBase extends React.PureComponent<ModalBaseProps, {}> {
...
-                className={cx(
-                  modal,
-                  modalWidth[modalSize],
-                  flex({ direction: "column" }),
-                  {
-                    [modalPreTransitionStyle(animationDuration)]: isAnimated,
-                    [modalTransitionStyles[state]]: isAnimated
-                  }
-                )}
+                className={cx(modal, modalWidth[modalSize], {
```
mperrotti: Mostly Prettier noise. All I did was remove the `flex` style

packages/modal/style.ts
```diff
@@ -43,9 +43,12 @@ export const scrim = css`
 export const modal = css`
   border-radius: ${borderRadiusDefault};
   background-color: ${themeBgPrimary};
+  display: flex;
+  flex-direction: column;
   left: 50%;
   max-height: calc(100vh - ${modalInset * 2}px);
   outline: none;
+  overflow: auto;
```
mperrotti: This has nothing to do with this regression.

While testing this I found an edge case where if a user's viewport height is ~200px or less, they can't click footer buttons

GeorgiSTodorov:
"FullscreenModal with additional header content" and "FullscreenModal w/ flushed content" are both wider that the screen for me. Tested on Firefox and Chrome.
Screenshot from 2019-07-12 09-53-38
Screenshot from 2019-07-12 09-54-04

mperrotti:
@GeorgiSTodorov - fixed

PR #372, opened 2019-07-11
Add aspectRatio prop to Card

mperrotti:
These cards in Konvoy and Kommander are limited by a 4:5 aspect ratio, and I think we may end up using this again.
Screen Shot 2019-07-11 at 7 40 50 PM

## Screenshot:
Screen Shot 2019-07-11 at 7 41 01 PM


GeorgiSTodorov:
This might be known, but it's not exactly 2:1. Same when I enter a 1:4 ratio.
Screenshot from 2019-07-12 09-58-14
Screenshot from 2019-07-12 10-01-24

mperrotti:
Good catch @GeorgiSTodorov! I fixed it

PR #374, opened 2019-07-12
Allow AppChrome components' padding to be 0 via theming

mperrotti:
This change is needed for the latest Kommander/Konvoy header bar designs - where the logo is at the top left with no padding.

This can be tested by choosing "none" from the padding-related select knobs in the Knobs panel

### DIFF
--- a/packages/appChrome/components/HeaderBar.tsx
+++ b/packages/appChrome/components/HeaderBar.tsx
@@ -18,7 +18,6 @@ export interface HeaderProps {
 class Header extends React.PureComponent<HeaderProps, {}> {
   public render() {
     const { children } = this.props;
-    /* tslint:disable:no-string-literal */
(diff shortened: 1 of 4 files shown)


packages/appChrome/components/HeaderBar.tsx
```diff
@@ -30,19 +29,22 @@ class Header extends React.PureComponent<HeaderProps, {}> {
         );
         return css`
           background-color: ${bgColor};
-          padding-left: ${spaceSizes[props.theme.headerPaddingHor] ||
```
mperrotti: `spaceSizes.none` is `0`, which is falsy. This caused the it to fall back to the value after the `||` operator

mperrotti: All the changes are the same, just with different theme property names

PR #379, opened 2019-07-17
DCOS-56537: add data attributes to simplify selectors for Cypress testing

mperrotti:


mperrotti:
Reviewers: This PR isn't as big and scary as it looks - it's all simple changes :)

packages/appChrome/components/SidebarItem.tsx
```diff
@@ -14,6 +14,11 @@ export interface SidebarItemProps {
 class SidebarItemComponent extends React.PureComponent<SidebarItemProps, {}> {
   public render() {
     const { children, isActive, onClick } = this.props;
+    const dataCy = [
```
GeorgiSTodorov: I can't find the data-cy for `SidebarItem`s.

packages/appChrome/components/SidebarSubMenuItem.tsx
```diff
@@ -64,9 +64,13 @@ class SidebarSubMenuItem extends React.PureComponent<
         )]: isActive
       }
     );
+    const dataCy = [
```
GeorgiSTodorov: I can't see the data-cy for this element.

packages/badge/components/badgeButton.tsx
```diff
@@ -40,7 +40,7 @@ export class BadgeButton extends React.PureComponent<BadgeButtonProps, {}> {
       ${badgeButton(appearance)};
     `;
     return (
-      <Clickable {...props}>
+      <Clickable {...props} data-cy="badgeButton">
```
GeorgiSTodorov: I can't see the data-cy for this element.

packages/breadcrumb/components/Breadcrumb.tsx
```diff
@@ -37,6 +37,7 @@ class Breadcrumb extends React.PureComponent<BreadcrumbProps, {}> {
             className={cx(textWeight("medium"), textSize("l"), {
               [padding("left", "xs")]: i !== 0
             })}
+            data-cy="breadcrumb"
```
GeorgiSTodorov: If we have One > Two, then "One", ">" and "Two" all have this data-cy. Is this expected?

packages/clickable/components/clickable.tsx
```diff
@@ -23,6 +23,10 @@ export interface ClickableProps {
    * Whether or not to reset the :focus outline style
    */
   disableFocusOutline?: boolean;
+  /**
+   * human-readable selector used for writing tests
+   */
+  dataCy?: string;
```
GeorgiSTodorov: This element doesn't have a data-cy.

pierrebeitz:
i'm totally in favour of having this as it will make our tests way better!  

the only concern that came to my mind was the bundle size. but as that does not accumulate like it would in static HTML but only adds a couple bytes per component, we won't see a noticeable increase, right?

packages/table/components/CheckboxTable.tsx
```diff
@@ -154,7 +154,11 @@ class CheckboxTable extends React.PureComponent<
     };
 
     return (
-      <div ref={this.checkboxRef} className={display("inline-block")}>
+      <div
```
GeorgiSTodorov: I can't see the data-cy for this element.

packages/table/components/Table.tsx
```diff
@@ -421,6 +421,11 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
     const onStartCallback = () => {
       this.setState({ resizeIndex: columnIndex });
     };
+    const dataCy = [
```
GeorgiSTodorov: I can't see the data-cy for this element.

packages/textInput/components/TextInput.tsx
```diff
@@ -75,12 +75,20 @@ export class TextInput<
...
+
     if (this.props.className) {
       containerProps.className = this.props.className;
     }
     return (
-      <div {...containerProps}>
-        <div className={flex({ align: "center" })}>
+      <div {...containerProps} data-cy={dataCy}>
+        <div className={flex({ align: "center" })} data-cy="textInput-label">
           {labelContent}
```
GeorgiSTodorov: Is there a reason why we don't have a data-cy for the actual input here? I think it's the most used part of this component.

packages/textarea/components/Textarea.tsx
```diff
@@ -98,8 +105,14 @@ class Textarea extends React.PureComponent<TextareaProps, {}> {
               required={required}
               {...{ ...other, onChange }}
             />
-            {getHintContent}
-            {hasError && getValidationErrors}
+            {getHintContent ||
+              (hasError &&
+                getValidationErrors && (
+                  <div data-cy="textarea-hintContent">
```
GeorgiSTodorov: I can't see the data-cy for hints.

packages/tooltip/components/Tooltip.tsx
```diff
@@ -58,6 +58,7 @@ class Tooltip extends React.PureComponent<TooltipProps, TooltipState> {
         onFocus={this.handleOpen}
         onBlur={this.handleClose}
         tabIndex={0}
+        data-cy="tooltip"
```
GeorgiSTodorov: I can't see the data-cy for this element.

README.md
```diff
@@ -93,6 +93,21 @@ describe('Badge', () => {
...
 
+## Testing with Cypress
+To make it easier to select DOM nodes of our components, DOM nodes have a `data-cy` attribute.
+
+### `data-cy` value naming convention
+**Parent nodes:** The value of `data-cy` for component's parent node is the same as the component name, but camelCased. For example: The parent node for `<PrimaryButton>` will have `data-cy="primaryButton"`
+
+**Child nodes:** If a child node has a `data-cy` added, there will be a dash between the parent node's name and a string to describe the child node. For example: The footer element of a `<DialogModal>` will have `data-cy="fullscreenModal-footer"`
+
+**States and variants:** If a node has a special "state", `data-cy` will prepend a string describing that state after a dot.
```
pierrebeitz: # 😻

packages/appChrome/components/SidebarItem.tsx
pierrebeitz: @GeorgiSTodorov 	is right. we need to search for things that are unwrapped. e.g. `Clickable` will be unwrapped, so only it's children are actually rendered. thus we lose our `data-cy`-prop.

mperrotti:
@pierrebeitz - I had a similar concern about bundle size, but it will be a very small increase

packages/breadcrumb/components/Breadcrumb.tsx
mperrotti: That is expected since each of those is a breadcrumb. Do you (or any other reviewers) think we should exclude `data-cy` on the ">" breadcrumbs?

packages/clickable/components/clickable.tsx
mperrotti: It's not supposed to. `Clickable` just renders it's children, so if you're using it on it's own, you just have to pass the `dataCy` prop

packages/breadcrumb/components/Breadcrumb.tsx
GeorgiSTodorov: I don't have a strong opinion, I just though it was a little strange. If I don't know this, I would think that the second breadcrumb would be "Two", not ">".

packages/table/components/CheckboxTable.tsx
mperrotti: Really? I do
Screen Shot 2019-07-19 at 10 44 40 AM

packages/textInput/components/TextInput.tsx
mperrotti: That's a great point. I'll add it

packages/table/components/CheckboxTable.tsx
GeorgiSTodorov: Which element did you select?
Screenshot from 2019-07-19 17-51-59

packages/tooltip/components/Tooltip.tsx
mperrotti: I do
Screen Shot 2019-07-19 at 10 56 07 AM

packages/table/components/Table.tsx
mperrotti: I'm seeing it:
Screen Shot 2019-07-19 at 10 45 48 AM

packages/table/components/CheckboxTable.tsx
mperrotti: Sorry, my screenshots weren't very helpful :x

Screen Shot 2019-07-19 at 10 58 12 AM

That `data-cy` is just for the checkbox in the table header

mperrotti:
I addressed @GeorgiSTodorov's feedback. In response to [this comment](https://github.com/dcos-labs/ui-kit/pull/379#discussion_r304819856), I also added a `data-cy` to each <input> (or <textarea> or <select>) element. This feels a little redundant to me, but might make writing selectors for form elements easier.

Reviewers: let me know what you think

packages/table/components/CheckboxTable.tsx
GeorgiSTodorov: What if we want a table row?

packages/table/components/Table.tsx
GeorgiSTodorov: This is just for the header?

packages/infobox/components/InfoBox.tsx
```diff
@@ -67,10 +67,14 @@ export class InfoBox extends React.PureComponent<InfoBoxProps, {}> {
       onDismiss,
       className
     } = this.props;
-
     const hasActions = primaryAction || secondaryAction;
-
     const isUrgentMessage = appearance === "danger" || appearance === "warning";
+    const dataCy = [
+      "infoBox",
+      ...(appearance && appearance !== "default"
```
pierrebeitz: saw this pattern a couple of times now and realized that i would have lazily gone with having `appearance.default` and justifying it to myself with the slightly lighter footprint on implementation cost (yes, the conditional). 

no action item, just wanted to share

packages/appChrome/components/SidebarSubMenuItem.tsx
pierrebeitz: works for me!

Storybook 2019-07-22 10-49-05

packages/table/components/Table.tsx
mperrotti: Yes

packages/table/components/CheckboxTable.tsx
mperrotti: You could use the checkbox id or select using the `data-cy` attributes on the parent cell, but I've added a data-cy attribute specifically for checkbox cells for simplicity: `table-checkboxCell table-checkboxCell.row${rowIndex}`

PR #380, opened 2019-07-18
[spike] DCOS-56471: add `url` prop for buttons that are links

mperrotti:
Passing a `url` prop will render an anchor tag styled like a button


GeorgiSTodorov:
Some snapshots fail for me locally.
Screenshot from 2019-07-19 12-14-36

mperrotti:
@GeorgiSTodorov - it's ok to update that snapshot. I made a change to those styles shortly after those components were merged

packages/button/components/ButtonBase.tsx
```diff
@@ -98,41 +102,66 @@ class ButtonBase extends React.PureComponent<ButtonBaseProps, {}> {
...
+            tabIndex={0}
+          >
+            {this.getButtonContent()}
+          </a>
+        ) : (
+          // tslint:disable react-a11y-anchors
+          //
+          // this rule was erroring because the anchor had no href, but in this
+          // case that is intentional beacuse it's disabled
+          <a className={buttonClassName} aria-disabled="true" tabIndex={-1}>
```
pierrebeitz: how about using a `button`-tag instead for the case no url is given and it should be disabled?

PR #385, opened 2019-07-31
Fix FF bug where overflowed flex items are cut off

mperrotti:
Instead of only applying 'min-height: 0' on the flex container when flex-direction is column, it always gets applied

For more info on this bug, see:
https://github.com/philipwalton/flexbugs\#flexbug-1
https://bugs.chromium.org/p/chromium/issues/detail\?id\=580196

## Screenshots
**Before (does not scroll all the way to the bottom of the content)**
flexboxOverflow_before

**After (does scroll all the way to the bottom of the content)**
flexboxOverflow_after


packages/shared/styles/styleUtils/layout/flexbox.ts
```diff
@@ -68,19 +65,21 @@ export const flex = (
...
   return css`
     ${getResponsiveStyle("align-items", flexProps.align)};
     ${getResponsiveStyle("height", height)};
     ${getResponsiveStyle("flex-direction", flexProps.direction)};
     ${getResponsiveStyle("flex-wrap", flexProps.wrap)};
     ${getResponsiveStyle("justify-content", flexProps.justify)};
-    ${getResponsiveStyle("min-height", minHeight)};
     box-sizing: border-box;
     display: flex;
+    min-height: 0;
```
mperrotti: This is the only real change. Everything else is a snapshot update

GeorgiSTodorov:
@mperrotti , where can I test this in the ui-kit?

mperrotti:
@GeorgiSTodorov - I was just testing in Kommander, but we can pair to write a story that would test this in ui-kit if you prefer

GeorgiSTodorov:
@mperrotti , I think testing it in Kommander is good enough, especially since the bug was discovered there.

PR #386, opened 2019-08-01
DCOS-56894: Fix Sidebar open/close toggle bugs

mperrotti:
Bugs fixed:
- The Sidebar would not toggle closed when `sidebarWidth` was set in the AppChrome theme
- The Sidebar open/close animation was broken
- `styled` was being called in the render method, which hurts React performance

## Testing:
1. Go to the story called "Sidebar Customizable w/ Knobs"
2. Uncheck the "isOpen" checkbox
3. Confirm the Sidebar opens and closes, and animates when it does

## Screenshots:
**Before - toggle bug:**
sidebarWidthThemeBug_before

**After - toggle bug:**
sidebarWidthThemeBug_after

**Before - animation bug:**
sidebarAnimationBug_before

**After - animation bug:**
sidebarAnimationBug_after

### DIFF
--- a/packages/appChrome/components/Sidebar.tsx
+++ b/packages/appChrome/components/Sidebar.tsx
@@ -1,8 +1,7 @@
 import * as React from "react";
-import { cx, css } from "emotion";
+import { css } from "emotion";
 import styled from "react-emotion";
 import { sidebar, sidebarAnimator } from "../style";
-import { atMediaUp } from "../../shared/styles/breakpoints";
 import {
   themeTextColorPrimary,
   themeTextColorPrimaryInverted,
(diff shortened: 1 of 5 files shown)


GeorgiSTodorov:
For me, both in Firefox and in Chrome, there is some delay after clicking the button and the animation is kind of fast so it looks like it disappears and appears suddenly instead of gradually.
Firefox
Peek 2019-08-02 08-43

Chrome
Peek 2019-08-02 08-44

mperrotti:
@GeorgiSTodorov - I think that's specific to the way Storybook knobs cause a re-render of the story. Does it behave that way when you click the "☰" in the header of the "AppChrome DCOS clone" story?

GeorgiSTodorov:
@mperrotti , then it's kind of fast, but still fine.

PR #387, opened 2019-08-01
fix: hides tooltip when mouse leaves disabled button

mperrotti:
## Testing
1. In the "default" Tooltip story, replace the `trigger` prop with:
```
trigger={<PrimaryButton disabled={true}>Hover me</PrimaryButton>}
```
2. Hover over the button and then off the button

The tooltip should disappear when your cursor leaves the button

## Screenshots
**Before:**
disabledButtonTooltipBug_before

**After:**
disabledButtonTooltipBug_after

### DIFF
--- a/packages/button/style.ts
+++ b/packages/button/style.ts
@@ -65,6 +65,7 @@ export const filledButton = (
 const mutedButton = css`
   ${tintContent(themeTextColorDisabled)};
   cursor: default;
+  pointer-events: none;
 
   &:hover,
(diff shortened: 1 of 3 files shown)


mperrotti:
@GeorgiSTodorov - that's what the changes in the `.stories.tsx` file were for :)

The trigger was in a block-level element, so if you hovered anywhere in that row, you'd be hovering the div that has the mouse handlers

PR #392, opened 2019-08-15
DCOS-56745: allow SpacingBox to set spacing per side

mperrotti:
There was a case in Kommander where I needed spacing on all sides but the bottom, and I think we'll also have situations where we want one spacing value for the left and right sides,  but a different spacing value for the top and bottom sides.

## Screenshot
Screen Shot 2019-08-15 at 7 48 52 PM

### DIFF
--- a/packages/shared/styles/styleUtils/modifiers/modifierUtils.ts
+++ b/packages/shared/styles/styleUtils/modifiers/modifierUtils.ts
@@ -20,8 +20,7 @@ export type BoxSides =
   | "bottom"
   | "left"
   | "horiz"
-  | "vert"
-  | undefined;
+  | "vert";
 export type SpaceSizes = "xxs" | "xs" | "s" | "m" | "l" | "xl" | "xxl" | "none";
 export type SpaceSize = BreakpointConfig<SpaceSizes>;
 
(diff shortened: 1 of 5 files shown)


mperrotti:
I was also playing with an alternative API for this component, but it would be a breaking change:
```
different spacing on top and bottom:
<SpacingBox top="xl" bottom="m">children</SpacingBox>

different spacing on horizontal and vertical sides:
<SpacingBox horiz="xl" vert="m">children</SpacingBox>

xl spacing on all sides:
<SpacingBox all="xl">children</SpacingBox>

default spacing ("m") on all sides:
<SpacingBox>children</SpacingBox>
```
Instead of configuring an object in the `spacingSizePerSide` prop, the strings from the `BoxSides` union type would be the props, and they'd accept values of the `SpaceSize` type. We'd no longer have the `side` or `spacingSize` props.

---

If any reviewers prefer the component API described above, we could either decide to introduce a breaking change, or file a ticket to follow up on this.

GeorgiSTodorov:
Also, I prefer the API here.

mperrotti:
@GeorgiSTodorov - the side padding isn't supposed to go to 0, it's supposed to just get smaller. I've set the prop like this:
```
    vert: {
        default: "none",
        medium: "l"
    },
    horiz: {
        default: "m",
        medium: "xl"
    }
}}
```

I can see how that feels a little strange. I'm going to update it so padding is "none" on all sides for smaller viewports

PR #393, opened 2019-08-16
Remove dead code from "utilities" directory

mperrotti:
It looks like most of what was in the "utilities" directory wasn't being used. I also searched for usage in projects that use ui-kit and didn't find anything


GeorgiSTodorov:
Shouldn't removing these be considered a breaking change?

mperrotti:
In master, the code I deleted isn't able to be imported anyway.
However, the directory structure change _would_ be a breaking change if anybody was using `getVarValue` or `InteractionModeEngine`.

I'll change the directory structure back to how it was so we don't have to introduce a breaking change for something so minor

PR #394, opened 2019-08-20
DCOS-57723: added a segmented control component

mperrotti:
## Screenshots
SegmentedControlDemo

### DIFF
--- a/packages/segmentedControl/README.md
+++ b/packages/segmentedControl/README.md
@@ -1,0 +1,20 @@
+# SegmentedControl
+
+A segmented control is group of two or more buttons that represent a list of options users can select from. Only one option can be selected at the same time.
+
+A segmented control can be a good alternative to a dropdown or a select input because it requires less interactions to make a selection. They can be a good alternative to a group of radio buttons when there are space limitations or if form controls look visually awkward.
+
+## Icons in button labels
+Button labels can contain icons. When using an icon in a button label, the default icon size to use is `iconSizeXs`.
+
+When an icon is the only button label content, be sure to provide text content to keep the buttons screenreader accessible. If you're using the `Icon` component to render the icon, this is as simple as passing a string to the `ariaLabel` prop.
+
+## Do's and Dont's
+
+### Do
+- Limit the number of segments to 5 or less. If you find yourself needing more than 5 segments, consider a group of radio buttons or a SelectInput
+- Keep the content in each segment consistent as much as possible. For example, all segments should just contain text or all segments should just contain icons
+
+### Don't
+- Use a segmented control for navigation instead of Tabs
+- Put long strings in the segments
--- a/packages/segmentedControl/components/SegmentedControlButton.tsx
+++ b/packages/segmentedControl/components/SegmentedControlButton.tsx
@@ -1,0 +1,70 @@
+import * as React from "react";
+import { cx } from "emotion";
+import shortid from "shortid";
+import {
+  segmentedControlButton,
+  segmentedControlButtonActive,
+  staticKeyboardFocusClassname
+} from "../style";
+import { visuallyHidden, display } from "../../shared/styles/styleUtils";
+import FocusStyleManager from "../../shared/components/FocusStyleManager";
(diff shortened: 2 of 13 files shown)


mperrotti:
Don't merge this until I've reviewed it with the designers

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+   */
+  id?: string;
+  /**
+   * Whether the button has been selected
+   */
+  isActive?: boolean;
+  /**
+   * The name passed to the input. Shared by all inputs in the segmented control
+   */
+  name?: string;
```
GeorgiSTodorov: If I am understanding correctly, all the button in a group have the same name?

packages/segmentedControl/README.md
```diff
@@ -0,0 +1,13 @@
+# SegmentedControl
+
+A segmented control is group of 2 or more buttons users can select a value from. A segmented control can be a good alternative to a dropdown or a select input because it requires less interactions to make a selection. A segmented control can be a good alternative to a group of radio buttons when there are space limitations or form controls would look visually awkward.
```
pierrebeitz: just out of curiosity: what's the rationale behind not calling it a "Button Group" like e.g. Bootstrap and Foundation do?

packages/segmentedControl/components/SegmentedControlButton.tsx
pierrebeitz: yes! and that happens to be the `id`-prop of the `SegmentedControl` component.

one noticable implication is that we can never have two buttons "active" at the same time. so this component can't be (ab)used as a multiselect.

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { id = shortid.generate(), isActive, onChange, name, value } = props;
+  return (
+    <label
+      className={cx(display("inline-block"), segmentedControlButton, {
+        [segmentedControlButtonActive]: props.isActive
+      })}
+      data-cy="segmentedControlButton"
+      htmlFor={id}
+    >
+      {/* TODO: more detail on why this is disabled */}
```
pierrebeitz: care to explain now? or is this a todo for later?

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The value of the input
+   */
+  value: string;
+}
+
+const SegmentedControlButton: React.SFC<
+  SegmentedControlButtonProps
+> = props => {
+  const { id = shortid.generate(), isActive, onChange, name, value } = props;
```
pierrebeitz: not that i see that as a really problematic thing, but on every change of the selection, we'll generate a new `shortid` for all the buttons within a `SegmentedControl` because of the `.bind` we use in there.

it somehow feels a little odd to me but i can't think of critical arguments against that behaviour (it's only recreationg of 5 DOM-nodes at max if used properly. hooray for the documentation btw!).

packages/segmentedControl/components/SegmentedControlButton.tsx
mperrotti: This is a TODO for now, I just forgot to do it 🙈 

There was a a11y linting error that was asking me to add ARIA props that would have been redundant and potentially create an awkward screenreader experience

packages/segmentedControl/README.md
mperrotti: I think of button groups as just being a styled list of buttons that have related actions.
This component is a list of options a user chooses from. I think of it more like a fancy group of radio buttons.

packages/segmentedControl/components/SegmentedControlButton.tsx
mperrotti: Yes, because it's a group of checkbox inputs

PR #395, opened 2019-08-22
DCOS-57267: Dropdown component

mperrotti:
## Screenshots
DropdownDemo

## Trade-offs:
This lets your click a menu item and take an action, but item's don't have a "selected" state like they would with a native `<select>` dropdown or the `Typeahead` component.


mperrotti:
I want to give a little more thought to the component names

GeorgiSTodorov:
Overall it looks good.
- Is it intentional that the gray hovered rectangle is smaller that the row and that the dropdown items don't have a special cursor?
Screenshot from 2019-08-23 10-14-55

- Also, when I hover a row and they go outside the dropdown with the cursor, the last hovered row remains hovered.
Peek 2019-08-23 10-15

- Unlike other dropdown I have seen, this one does not close when we resize the window or when we click on the title of the browser page.

- With sections looks like this for me.
Screenshot from 2019-08-23 10-20-29

- Same with (with icons and avatars).
Screenshot from 2019-08-23 10-23-47
Screenshot from 2019-08-23 10-25-18
Screenshot from 2019-08-23 10-25-26

These are the same in Chrome, and there I get a disabled scrollbar.
Screenshot from 2019-08-23 10-26-33

packages/dropdown/README.md
```diff
@@ -0,0 +1,26 @@
+# Dropdown
+
+A dropdown toggles an overlay menu of actions. It appears as a result of a user interacting with some trigger (usually a button or icon) to open the menu.
+
+Dropdown menus can be opened via cursor or via keyboard navigation, and it's menu items can navigated and selected via cursor or via keyboard navigation.
+
+## Dropdown triggers
+By default, use one of the dropdown button components (e.g.: `PrimaryDropdownButton`) as the dropdown trigger.
```
pierrebeitz: I know i'm questioning stuff outside this PR, but: What's the reasoning to go with very specific components (like `PrimaryDropdownButton`) instead of making them parameterized with e.g. an appearance?

Appearance could even be stringly-typed so that users would not have to import the `DropdownButtonAppearance`-enum.

```
<DropdownButton appearance="secondary">Edit</DropdownButton>
```

That's a very opinionated question and i'm totally fine with a "🤷‍♂".

packages/dropdowMenu/style.ts
```diff
@@ -14,6 +14,7 @@ export const menuList = css`
```
pierrebeitz: the `overflow-y: scroll` looks weird in case you don't use an apple's trackpad and the scrollbar is rendered. that might be the issue @GeorgiSTodorov has screenshotted. https://github.com/dcos-labs/ui-kit/blob/9a6c61c4ce3fef8867ca6bff790c1c8ca577fe6f/packages/dropdowMenu/style.ts#L12

mperrotti:
@GeorgiSTodorov - does the dropdown look like that on the Typeahead component as well? If so, does it look like that in master as well?

packages/dropdown/components/Dropdown.tsx
```diff
@@ -0,0 +1,173 @@
...
+    >
+  ) => void;
+  /**
+   * which DOM node the dropdown menu will attach to
+   */
+  overlayRoot?: HTMLElement;
+  /**
+   * The preferred direction to open the menu in relation to the trigger
+   */
+  preferredDirections?: Direction[];
```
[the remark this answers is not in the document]
mperrotti: Yea, I have to check and make sure I'm using that prop correctly. It works as expected on the `Dropdownable` component in Storybook

packages/dropdown/components/Dropdown.tsx
```diff
@@ -0,0 +1,173 @@
...
+        return {
+          sections: [...sections, childrenWithKeys],
+          menuItemIndex: acc.menuItemIndex
+        };
+      },
+      { sections: [], menuItemIndex: -1 }
+    );
+
+  return (
+    <Downshift
```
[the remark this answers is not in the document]
mperrotti: Everything in ui-kit is keyboard accessible 😃

packages/dropdown/components/DropdownActions.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { DropdownActionItemProps } from "../components/DropdownActionItem";
+
+export interface DropdownActionsProps {
+  children:
+    | React.ReactElement<DropdownActionItemProps>
+    | Array<React.ReactElement<DropdownActionItemProps>>;
+}
+
+const DropdownActions: React.SFC<DropdownActionsProps> = ({ children }) => (
+  <React.Fragment>{children}</React.Fragment>
```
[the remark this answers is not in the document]
mperrotti: You're correct, this _could_ technically be represented by an ordinary div, but I chose to make it a Fragment to avoid DOM nesting

packages/dropdown/README.md
mperrotti: We chose to do buttons this way to simplify the prop interface.

packages/dropdowMenu/style.ts
mperrotti: I think you're right. I'll try and be smarter about when we apply this style

GeorgiSTodorov:
> Does the dropdown look like that on the Typeahead component as well? If so, does it look like that in master as well?

Typeahead looks worse on master. Not only does the hover not take the entire row, but we have a useless scrollbar.
Screenshot from 2019-08-23 19-48-10

> Which dropdown are you referring to? I don't see this as an issue
For example, the current dropdowns in dcos-ui.

mperrotti:
Yea, I was aware of the horizontal scroll, which is fixed in this PR

mperrotti:
Tagging Brandon for review to confirm what I did to `Dropdownable` makes sense

packages/dropdowMenu/components/DropdownMenuItemAvatar.tsx
```diff
@@ -0,0 +1,21 @@
...
+   */
+  position?: "start" | "end";
+}
+
+const DropdownMenuItemAvatar: React.SFC<DropdownItemAvatarProps> = props => (
+  <Avatar {...props} />
+);
+
+DropdownMenuItemAvatar.defaultProps = {
+  size: iconSizeXs,
```
GeorgiSTodorov: Why is size not in the interface?

GeorgiSTodorov:
It looks better than before, but now the max height dropdown has a horizontal scrollbar. And I still think that the dropdown items need a special cursor.
Screenshot from 2019-08-26 09-15-54

GeorgiSTodorov:
That's true, but section can mean one of several sections.

GeorgiSTodorov:
@mperrotti , the original renaming you suggested is also fine with me.

packages/dropdownMenu/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Icons and avatars in items
+Dropdown menu items can display an icon or avatar before or after the menu item text. By default, the avatar or icon is sized at `iconSizeXs`.
+
+## Do's and Dont's
+
+### Do
+- Use concise language to improve scannability
+- Use verbs for menu items that are actions
+- Avoid hiding actions in dropdowns when possible - especially critical actions
```
GeorgiSTodorov: What is meant by this?

packages/popover/style.ts
```diff
@@ -9,10 +9,18 @@ export const menuList = css`
   background-color: ${themeBgPrimary};
   border-radius: ${borderRadiusDefault};
   box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
-  overflow-y: scroll;
+  overflow: auto;
+  -ms-overflow-style: -ms-autohiding-scrollbar;
+`;
+
+// Hides unnecessary horizontal scrollbars on Linux and Microsoft
```
GeorgiSTodorov: Shouln't this say "Windows"? :D

packages/dropdownMenu/README.md
GeorgiSTodorov: No, I was just wondering what is the dropdown good for if not for actions.

PR #396, opened 2019-08-26
DCOS-57915: adds ProgressBar component

mperrotti:
## Screenshots:
Screen Shot 2019-08-26 at 2 53 10 PM
Screen Shot 2019-08-26 at 2 53 28 PM
Screen Shot 2019-08-26 at 2 53 38 PM

### DIFF
--- a/packages/utilities/components/percentage.ts
+++ b/packages/utilities/components/percentage.ts
@@ -2,7 +2,10 @@ function calculate(a, b) {
   return (b / 100) * a;
 }
 
-function percentage(a?: number, b?: number) {
+// Using "any" to get around a Typescript error that popped up
+// when upgrading to 3.5.3.
+// This function isn't even used anywhere.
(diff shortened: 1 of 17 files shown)


GeorgiSTodorov:
Do we want to enable renovate here? We have https://github.com/dcos-labs/ui-kit/pull/240 , but it's broken.

GeorgiSTodorov:
Yes, I would like to. I don't know either, but we can check together.

packages/utilities/components/percentage.ts
```diff
@@ -2,7 +2,10 @@ function calculate(a, b) {
   return (b / 100) * a;
 }
 
-function percentage(a?: number, b?: number) {
+// Using "any" to get around a Typescript error that popped up
+// when upgrading to 3.5.3.
+// This function isn't even used anywhere.
```
GeorgiSTodorov: Can we just remove it then?

PR #401, opened 2019-09-11
DCOS-57982: component for rendering badges in a text input

mperrotti:
## Testing
You should be able to add badges by hitting "Enter" after you type some value into the input
You should be able to add badges using the Typeahead variants
You can't add duplicates

## Screenshots
**Default:**
badgeDemo_default

**Pre-filled:**
Screen Shot 2019-09-11 at 6 02 04 PM

**w/ a Typeahead:**
badgeDemo_typeahead

---

Some small things I'd like to address before merging, but probably aren't blockers:
- [ ] The vertical centering of the text in the badges is a little off when the `<input>` isn't on the same line as the badges
- [ ] I'd like to add some handler for when a user tries to add a duplicate badge
- [x] I want to reconsider "sanitizing" the string that goes into the `value` property of the badge data

### DIFF
--- a/packages/badge/components/badge.tsx
+++ b/packages/badge/components/badge.tsx
@@ -1,14 +1,16 @@
 import * as React from "react";
 import { badge } from "../style";
 
+export type BadgeAppearance =
+  | "default"
+  | "success"
+  | "primary"
+  | "warning"
+  | "danger"
+  | "outline";
+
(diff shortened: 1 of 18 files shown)


mperrotti:
Good catch @GeorgiSTodorov - I can fix that

PR #403, opened 2019-09-13
Fix SidebarSubMenu issue w/ horiz padding on large screens

mperrotti:
## Screenshots
**Before:**
Screen Shot 2019-09-12 at 9 48 17 PM

**After:**
Screen Shot 2019-09-12 at 9 48 05 PM


mperrotti:
@GeorgiSTodorov - the issue should be fixed now

PR #415, opened 2019-10-10
Allow modals to choose which node they mount in

mperrotti:


packages/modal/components/ModalBase.tsx
```diff
@@ -38,6 +38,14 @@ export interface ModalBaseProps {
    * human-readable selector used for writing tests
    */
   dataCy?: string;
+  /**
+   * which DOM node the dropdown menu will attach to
```
TattdCodeMonkey: which DOM node the modal will attach to

PR #418, opened 2019-10-17
[DEMO/DISCUSSION] possible way of solving double-scrollbar in kommander

pierrebeitz:
allow style-attr to be passed in order to let consumers decide on how/where to embed a table.

i abused a story to reproduce the kommander-issue and demo it. those changed would need to be removed again of course.

i have no idea on how to do idiomatic ui-kit-stuff, so please excuse this potentially very unsophisticated approach.

⚠️ this might be totally opposing the philosophy of ui-kit and everyone should use the flex-helpers to solve problems like we currently have in kommander? 🤔 

# before 

Table - column width fill remaining ⋅ Storybook 2019-10-17 12-27-39

# after

Table - column width fill remaining ⋅ Storybook 2019-10-17 12-24-48

### DIFF
--- a/packages/table/stories/Table.stories.tsx
+++ b/packages/table/stories/Table.stories.tsx
@@ -126,46 +126,57 @@ storiesOf("Table", module)
     </div>
   ))
   .add("column width fill remaining", () => (
-    <div
-      style={{
-        height: "175px",
-        width: "100%",
-        fontSize: "14px"
-      }}
-    >
-      <Table data={items}>
-        <Column
-          header={<HeaderCell>name</HeaderCell>}
-          cellRenderer={nameCellRenderer}
-          growToFill={true}
-          minWidth={100}
-          maxWidth={150}
-        />
-        <Column
-          header={<HeaderCell>role</HeaderCell>}
-          cellRenderer={roleCellRenderer}
-          growToFill={true}
-        />
-        <Column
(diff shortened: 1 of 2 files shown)


packages/table/stories/Table.stories.tsx
```diff
@@ -126,46 +126,57 @@ storiesOf("Table", module)
...
+      >
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+        <div>Lorem ipsum hassenichgesehn</div>
+
+        <Table data={items} style={{ flexGrow: 1 }}>
```
pierrebeitz: we can now pass arbitrary styling in case we need to embed the table as e.g. a flex-child.

maybe this is bad and we should encourage everyone to use the flex-utils provided by ui-kit to solve the problem at hand... 🤷‍♂

pierrebeitz:
after reading https://github.com/bvaughn/react-virtualized/blob/master/docs/usingAutoSizer.md#can-i-use-autosizer-within-a-flex-container i think that's an utterly bad approach.

PR #419, opened 2019-10-17
docs(designdoc): add design doc file to repository

Poltergeist:
This adds the design doc to the repository to keep the living design doc close to the code.

### DIFF
--- a/design-doc.md
+++ b/design-doc.md
@@ -1,0 +1,179 @@
+## Table of Contents
+
+
+{{TOC}}
+
+
+
+## Links to Old Documents {#links-to-old-documents}
+
+
+
+*   [https://docs.google.com/document/d/1b7Or5gidpfQenkGR4dysz9ShEWgrWxxPJuQUBrnRXvs/edit#heading=h.3svxzuabblb4](https://docs.google.com/document/d/1b7Or5gidpfQenkGR4dysz9ShEWgrWxxPJuQUBrnRXvs/edit#heading=h.3svxzuabblb4)
+*   [https://docs.google.com/document/d/1smOrLXBs4E9KvRcjFzVFSGRLSOdJFioi8HHv_xvMlCI/edit](https://docs.google.com/document/d/1smOrLXBs4E9KvRcjFzVFSGRLSOdJFioi8HHv_xvMlCI/edit)
+*   [https://docs.google.com/document/d/1hFMPtjQnlscUwIauxcjAPxFmRAgruZ-6TBnKGEW4PWc/edit#heading=h.3svxzuabblb4](https://docs.google.com/document/d/1hFMPtjQnlscUwIauxcjAPxFmRAgruZ-6TBnKGEW4PWc/edit#heading=h.3svxzuabblb4)
+*   [https://docs.google.com/document/d/1-V07kO9L1mab8GV1-OQoLTAIDPZyWi5P_PRoIsEHeik/edit#heading=h.3svxzuabblb4](https://docs.google.com/document/d/1-V07kO9L1mab8GV1-OQoLTAIDPZyWi5P_PRoIsEHeik/edit#heading=h.3svxzuabblb4)
+*   [https://docs.google.com/document/d/1fiRlWzmeHYXGtMXEUCcs82sLQDyZc1oSdY5lUs566jc/edit](https://docs.google.com/document/d/1fiRlWzmeHYXGtMXEUCcs82sLQDyZc1oSdY5lUs566jc/edit)
+*   [https://docs.google.com/document/d/1p8IW_QuzMO6EhLiBOPJY0FqD0ZoS64-7M_aZvP9y2-8/edit#](https://docs.google.com/document/d/1p8IW_QuzMO6EhLiBOPJY0FqD0ZoS64-7M_aZvP9y2-8/edit#)
+*   [https://github.com/dcos-labs/ui-kit/blob/master/CONTRIBUTING.md](https://github.com/dcos-labs/ui-kit/blob/master/CONTRIBUTING.md)
+*   [https://docs.google.com/document/d/1p-CnUnhNFVz6WomPQvMEsTybbGgV0FHWgsUzBqX64cs/edit#heading=h.apvnjfls8tkq](https://docs.google.com/document/d/1p-CnUnhNFVz6WomPQvMEsTybbGgV0FHWgsUzBqX64cs/edit#heading=h.apvnjfls8tkq)
+
+
+## Living Document
+
+This is a living document[^1]. Feel free to add or change information. Everybody should be able to contribute to this document.
+
+
+## General Guidelines
+
+
+### Why are we building the design system?
+
+Have one source of truth, build a consistent product across company
+
+and platforms, increase developer productivity and feature delivery.
+
+
+### When does a component go into ui-kit
+
+Ui-kit follows the rules of three, which means once a component is used in 3 different places it should be added to ui-kit.
+
+
+### Flexibility
+
+We are trying to avoid unnecessary flexibility of a component. There is always room for building on the existing component or creating a new component for a different use case. Always when adding flexibility for another use case consider adding a new component instead or ensure that the complexity of the component is not exploding.
+
+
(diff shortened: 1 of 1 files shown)


design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+#### Avoid weak elements
+
+Weak elements are elements which can’t nest block elements, for example, a `p` tag is not able to wrap around a div which will reduce the comparability of a component. This is why we try to rather use a strong element like `section` or `div` instead of `p`.
+
+
+#### Declare Prop Types
+
+Since the `ui-kit` is using typescript please make sure to define proper prop types for your components. This will improve the documentation, which creates dynamically and will provide better error messages for the users. And also this will provide auto-completion in some IDE’s. Describe the props as well as possible, so also define object shapes or array shapes.
```
GeorgiSTodorov: Suggestions:
"which creates dynamically" -> "which **is created** dynamically"
"And also this will provide" -> "And also **these types** will provide"
"Describe the props as well as possible" -> "Describe the props as **good** as possible"
"so also define object shapes or array shapes." -> "**including** object **types** **and** array **types**."

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+This is a living document[^1]. Feel free to add or change information. Everybody should be able to contribute to this document.
+
+
+## General Guidelines
+
+
+### Why are we building the design system?
+
+Have one source of truth, build a consistent product across company
```
GeorgiSTodorov: Isn't present continuous tense better here?
 `Having one source of truth, building a consistent product across company and platforms, increasing developer productivity and feature delivery.`

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+#### Declare Prop Types
+
+Since the `ui-kit` is using typescript please make sure to define proper prop types for your components. This will improve the documentation, which creates dynamically and will provide better error messages for the users. And also this will provide auto-completion in some IDE’s. Describe the props as well as possible, so also define object shapes or array shapes.
+
+
+#### Never hard code HTML ID’s
+
+This might conflict at some point. If you need an id on the element expose a prop for that so the user can define it based on the application	.
```
GeorgiSTodorov: Space before the dot.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+Weak elements are elements which can’t nest block elements, for example, a `p` tag is not able to wrap around a div which will reduce the comparability of a component. This is why we try to rather use a strong element like `section` or `div` instead of `p`.
+
+
+#### Declare Prop Types
+
+Since the `ui-kit` is using typescript please make sure to define proper prop types for your components. This will improve the documentation, which creates dynamically and will provide better error messages for the users. And also this will provide auto-completion in some IDE’s. Describe the props as well as possible, so also define object shapes or array shapes.
+
+
+#### Never hard code HTML ID’s
```
GeorgiSTodorov: The plural of ID is IDs.
https://english.stackexchange.com/questions/212040/is-the-plural-form-of-id-spelled-ids-or-id

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+### Why are we building the design system?
+
+Have one source of truth, build a consistent product across company
+
+and platforms, increase developer productivity and feature delivery.
+
+
+### When does a component go into ui-kit
+
+Ui-kit follows the rules of three, which means once a component is used in 3 different places it should be added to ui-kit.
```
GeorgiSTodorov: Isn't this one rule? So the rule of three.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+We are trying to avoid unnecessary flexibility of a component. There is always room for building on the existing component or creating a new component for a different use case. Always when adding flexibility for another use case consider adding a new component instead or ensure that the complexity of the component is not exploding.
+
+
+### Considerations
+
+
+#### Avoid weak elements
+
+Weak elements are elements which can’t nest block elements, for example, a `p` tag is not able to wrap around a div which will reduce the comparability of a component. This is why we try to rather use a strong element like `section` or `div` instead of `p`.
```
GeorgiSTodorov: Doesn't "we prefer to use a strong element" sound better?

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+#### Never hard code HTML ID’s
+
+This might conflict at some point. If you need an id on the element expose a prop for that so the user can define it based on the application	.
+
+
+#### Use Logical Defaults
+
+Use prop defaults which make sense this will reduce the efforts needed to adopt this component.
```
GeorgiSTodorov: Suggestion:
"Use prop defaults which make sense**. This** will reduce the efforts needed to adopt this component."

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+#### A11y Support
+
+A good start on that topic is to read [this website](https://a11yproject.com/) While creating your component think about users: - Which uses keyboards - expose tab indexes as props - Use semantic correct HTML - ARIA
```
GeorgiSTodorov: No dot before "While". Shouldn't each of these dashes be on a new line?

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+
+*   consistency
+*   less typing needed
+*   less error prone
+
+But please avoid using configuration Objects if you have 3 or fewer properties.
+
+A example pattern for this can be found [here](#configuration-objects).
```
GeorgiSTodorov: **An** example.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+#### Single Responsibility Principle
+
+Keep the components simple by applying them to a single responsibility. In that way, the component will stay simple in most cases. Try not to introduce unnecessary props. Each prop is a configuration which will automatically increase the complexity. Before adding a property to an existing component ask yourself if this could be a new component.
+
+
+#### Keep Components Simple
+
+It's important to keep components as simple as possible, add logic to a component if it improves performance or quality of that component. For example, paginating a list if the number of items is exceeding a certain amount or having a show toggle to show more content like a dropout.
```
GeorgiSTodorov: "It's important to keep components as simple as possible, **only** add logic to a component if it improves **the** performance or **the** quality of that component."

I don't understand what the second sentence is trying to say.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+    </ul>
+  </Sidebar>
+</Page>
+```
+
+
+
+## Component Guidelines
+
+Each component should follow a common guideline, so that all components are reasonable within our ideals and they feel like a collection of similar working component. This will increase the ease of use for customers. To ensure this we try to put the most applicable things into this document.
```
GeorgiSTodorov: I don't fully understand the first sentence.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+  data,
+  Advanced: true
+}} />
+```
+
+
+
+#### Named slots
+
+Named slots are a way to describe certain areas of a component like `sidebar,` `header` and `content `to name some very general examples. These could be exposed as props but this often should not be the case. Since props are intended to be attributes. We should consider creating a sub component like `<Sidebar>`,` <Header>` and `<Content>` in the example. The intend here is that the component stays composable by that approach instead of being configurable.
```
GeorgiSTodorov: There is a comma and a whitespace inside the `.

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+Other stuff we might add
```
GeorgiSTodorov: Do we need this here?

design-doc.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+### Specific Patterns
+
+
+[^1]: A **living document**, also known as an **evergreen document** or **dynamic document**, is a [document](https://en.wikipedia.org/wiki/Document) that is continually edited and updated.([https://en.wikipedia.org/wiki/Living_document](https://en.wikipedia.org/wiki/Living_document))
```
GeorgiSTodorov: I don't think that linking the wikipedia article for a document is necessary.

PR #422, opened 2019-10-21
DCOS-60141: does not set position if dropdownable is not open

mperrotti:
This is a performance improvement that should resolve [DCOS-60141](https://jira.mesosphere.com/browse/DCOS-60141) when ui-kit gets updated in that Kommander with these changes.

## Testing
- Run a build of ui-kit using this branch
- Copy the build from `ui-kit/dist/` to `{kommander root}/packages/ui-components/node_modules/@dcos/ui-kit/dist/`
- In Kommander, navigate to the clusters cards view and resize the window

The card layout should not feel as slow as before, and the FPS should remain steady when resizing.

### Before
[Video](https://cl.ly/c47ebfc9e5cc)

DevTools perf:
Screen Shot 2019-10-21 at 1 44 14 PM

### After
[Video](https://cl.ly/01088df763b9)

DevTools perf
Screen Shot 2019-10-21 at 1 28 13 PM


mperrotti:
There were some snapshot updates as a result of this change. I tested `Tooltip` and other components that use `DropdownMenu` in ui-kit and inn Kommander, and everything is still behaving as expected.

PR #423, opened 2019-10-24
DCOS-60200: support for disabled dropdown menu items

mperrotti:
This was done in support of [DCOS-60200](https://jira.mesosphere.com/browse/DCOS-60200).

I also took care of this for the Typeahead dropdown while I was making changes

## Screenshot
DisabledDropdownDemo


mperrotti:
@GeorgiSTodorov - yes, because that item is still selected. If I was to press key "Enter", that item would be selected

PR #424, opened 2019-10-25
DCOS-57973: adds support for vertical tabs

mperrotti:
## Screenshot
Screen Shot 2019-10-25 at 3 32 05 PM

### DIFF
--- a/packages/tabs/style.ts
+++ b/packages/tabs/style.ts
@@ -1,0 +1,127 @@
+import { css, cx } from "emotion";
+import { border } from "../shared/styles/styleUtils";
+import {
+  SpaceSizes,
+  spaceSizes
+} from "../shared/styles/styleUtils/modifiers/modifierUtils";
+import { atMediaUp } from "../shared/styles/breakpoints";
+import {
+  TabDirection,
+  TabDirections,
+  defaultTabDirection
+} from "./components/Tabs";
+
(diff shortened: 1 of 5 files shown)


GeorgiSTodorov:
If I make the browser window small, I don't get a scrollbar or anything like that. Is this intentional?
Screenshot from 2019-10-28 10-55-26

mperrotti:
@GeorgiSTodorov - Yes. The tab content just fills the remaining space

GeorgiSTodorov:
Does that mean that we need some other element to make the tab content scrollable?

mperrotti:
We should be switching to horizontal tabs on smaller viewports. Maybe I should make the `direction` prop responsive. e.g.: `direction={{default: "horiz", medium: "vert"}}`

mperrotti:
Hold merge until I add the responsive support

packages/tabs/style.ts
```diff
@@ -0,0 +1,127 @@
...
+              css`
+                ${atMediaUp[breakpoint](css`
+                  ${getHorizOrVertStyle(direction[breakpoint])};
+                `)};
+              `
+            );
+
+            return acc;
+          },
+          Array()
```
DanielMSchmidt: Just curious, why not `[]`?

mperrotti: Typescript doesn't like `[]` for some reason ¯\_(ツ)_/¯

PR #426, opened 2019-11-01
Adds FullscreenView component

mperrotti:


packages/fullscreenView/README.md
```diff
@@ -0,0 +1,5 @@
+# FullscreenView
+
+The FullscreenView component should be used when we want to bring the user into a flow that does not have any of the usual app's chrome.
+
+The FullscreenModal component can be used in cases where we don't want to remove all other content from the page.
```
GeorgiSTodorov: We are not removing the content, we are just hiding it as far as I know.

PR #430, opened 2019-11-06
Fix sortable table

GeorgiSTodorov:
Go to a [table that should be sortable](http://localhost:6006/?path=/story/table--with-sortable-column), and verify that it is.

I accidentally wrote scrollable instead of sortable in the branch name and commit message.

### DIFF
--- a/packages/table/components/HeaderCell.tsx
+++ b/packages/table/components/HeaderCell.tsx
@@ -1,31 +1,9 @@
-import * as React from "react";
-
-import { default as Cell, CellProps } from "./Cell";
+import { default as Cell } from "./Cell";
 import styled from "react-emotion";
-import { headerCellCss, textCapitalize } from "../style";
+import { headerCellCss } from "../style";
 import { textTruncate } from "../../shared/styles/styleUtils";
 
-export interface HeaderCellProps extends CellProps {
-  /**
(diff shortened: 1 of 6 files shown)


packages/table/components/HeaderCell.tsx
```diff
@@ -12,20 +12,11 @@ export interface HeaderCellProps extends CellProps {
...
-      : styled(Cell)`
-          ${headerCellCss};
-          ${textCapitalize};
-          ${textTruncate};
-        `;
-    return <HeaderCell>{this.props.children}</HeaderCell>;
-  }
-}
+const HeaderCell = (props: HeaderCellProps) => (
+  <Cell lowerCase={props.lowerCase} {...props} />
```
mperrotti: Why do we have to specifically pass `lowerCase`? Does it not come through with `...props`?

GeorgiSTodorov: I did not think of that, it probably does.

packages/modal/tests/__snapshots__/Modal.test.tsx.snap
```diff
@@ -1258,55 +1258,170 @@ exports[`Modal DialogModal renders DialogModal 1`] = `
             }
           >
             <div>
-              <FocusLockCombination>
```
GeorgiSTodorov: I am not sure why this file was changed.

PR #435, opened 2019-11-21
[POC?] don't rerender allTheCells allTheTime

pierrebeitz:
# ⚠️ Checkboxtable does not have highlighting for disabled/muted rows in the current state of this PR.

Have a look at ContentCell in the Table-Component. Its sole purpose is to
highlight rows once a CheckboxTable has a disabled or checked row.

To do that it's computing css in a very dynamic way. That dynamic computation
breaks reacts reconciliation - effectively forcing every cell to rerender on
every update (a scroll-event is an update for the MultiGrid we use).

We *should* get rid of the dynamic nature of `ContentCell`. Unfortunately i
don't know how to do that, so this is rather a POC that breaks disabled and
checked rows in the `CheckboxTable`.

# Before

Note all the flashing boxes indicating a rerender.

before

# After

A lot less flashing boxes 🎉

after

### DIFF
--- a/packages/table/components/Table.tsx
+++ b/packages/table/components/Table.tsx
@@ -412,10 +412,7 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
     >;
 
     const onStopCallback = (_, data) => {
-      this.resizeColumn({
-        dragDelta: data.x,
-        index: columnIndex.toString()
-      });
+      this.resizeColumn({ dragDelta: data.x, index: columnIndex.toString() });
       this.setState({ resizeIndex: -1 });
     };
     const onStartCallback = () => {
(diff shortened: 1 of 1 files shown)


packages/table/components/Table.tsx
```diff
@@ -587,26 +579,29 @@ export class Table<T> extends React.PureComponent<TableProps, TableState> {
...
-      <ContentCell
-        className={cx(cellCss, vAlignChildren, {
-          [css`
-            ${rowHoverStyles};
-          `]: rowIndex === this.state.hoveredRowIndex,
-          [preventSortOverlap(
-            RESIZE_ICON_SIZE
-          )]: this.cellNeedsPreventIconOverlap(header, column)
-        })}
+      <div
```
pierrebeitz: this is the main change.

mperrotti:
@pierrebeitz - I figured out a way we can do this with Context, but I'm wondering if there's a better solution 🤔 

If you have time, I'd love to spend a few minutes pairing tomorrow.

PR #436, opened 2019-11-22
fix: arrow direction / asc & desc

brandonc:
When sort direction is "ascending" the arrow should point up

Here is the fixed version:

Nov-21-2019 23-02-12


Poltergeist:
I feel like the arrow should point into the direction of the sorting. So ascending showing downwards since we list things from top to bottom. I will do some more research on this how do you feel about this @mperrotti ?

PR #437, opened 2019-11-26
DCOS-61091 & 61090: empty state components

mperrotti:
## Screenshots
Screen Shot 2019-11-26 at 2 38 38 PM
Screen Shot 2019-11-26 at 2 38 52 PM

### DIFF
--- a/packages/emptyState/components/EmptyStateWithGraphic.tsx
+++ b/packages/emptyState/components/EmptyStateWithGraphic.tsx
@@ -1,0 +1,63 @@
+import * as React from "react";
+import { SpacingBox } from "../../styleUtils/modifiers";
+import { HeadingText2 } from "../../styleUtils/typography";
+import { EmptyStateProps } from "./EmptyState";
+import EmptyStateActions from "./EmptyStateActions";
+import { emptyStateGraphic, emptyStateWithGraphicBody } from "../style";
+
+interface GraphicDimensions {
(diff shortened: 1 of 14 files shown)


packages/emptyState/components/EmptyStateWithGraphic.tsx
```diff
@@ -0,0 +1,63 @@
...
+const EmptyStateWithGraphic: React.SFC<EmptyStateWithGraphicProps> = ({
+  graphicSrc,
+  graphicDimensions,
+  heading,
+  children,
+  primaryAction,
+  secondaryAction
+}) => {
+  const hasActions = primaryAction || secondaryAction;
+  return (
```
GeorgiSTodorov: Can we reuse `EmptyState` here?

mperrotti: We could, but why would we?

GeorgiSTodorov: For consistency.

mperrotti: There is a subtle (but intentional) difference with the spacing between the headline and the body content

GeorgiSTodorov: Okay, makes sense.

PR #441, opened 2019-12-02
DCOS-61479: add components and docs for loading indicators

mperrotti:
## Screenshots
InlineLoadingIndicatorDemo
SectionLoadingIndicatorDemo

### DIFF
--- a/packages/loadingIndicator/README.md
+++ b/packages/loadingIndicator/README.md
@@ -1,0 +1,31 @@
+# Loading indicators
+Loading indicators communicate that some action is in progress. They are displayed in place of content that cannot be shown until some data is loaded or some action is processed.
+
+## When to show a loading indicator
+To helped improve the perceived performance of our apps, don't let loading indicators hide available content. Loading indicators are only displayed in place of content that cannot possibly be rendered until some action or request has been processed.
+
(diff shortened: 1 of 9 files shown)


packages/loadingIndicator/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Avoid using this component in place of a button. If we're waiting for data before a user can interact with a button, disable the button and show a tooltip explaning why that button is disabled.
+
+For example: showing a loading indicator in place of a package version number until the API call to get the package data.
+
+## Do's and Dont's
+
+### Do
+- Show as much content as possible to avoid showing loading indicators too often
+- Avoid showing more than one `SectionLoadingIndicator` on the screen at once
```
GeorgiSTodorov: Isn't this a `Don`t`?

mperrotti: I didn't want to put it in the "don't" category because I think there will be rare cases where we will need more than one. For example, imagine a screen where you review a configuration: we might show a `SectionLoadingIndicator` for each `ConfigurationMap` on that screen.

Maybe I'm being too cautious? 🙈 Happy to move this to the "Don't" category if you think it's confusing here

GeorgiSTodorov: I think it's fine either way.

PR #447, opened 2019-12-07
fix(checkboxtable): calculates checkbox col width after mount

mperrotti:
These changes fix a bug @tnguyen0 was seeing in the Learning Hub app. The checkbox column of the `CheckboxTable` was not calculated correctly because the `ref` that was used to get the width dimension of the checkbox was coming back undefined.


mperrotti:
I was not able to reproduce this bug in ui-kit or kommander. @tnguyen0 - will you please review this one by testing locally in `learning-hub-server`?

mperrotti:
I talked to @tnguyen0, and he can no longer reproduce the original issue. I still think it's a good idea  to merge these changes regardless

packages/table/components/CheckboxTable.tsx
```diff
@@ -72,27 +73,30 @@ class CheckboxTable extends React.PureComponent<
...
+  componentDidMount() {
     const checkbox = this.checkboxRef.current;
 
-    return checkbox
-      ? checkbox.getBoundingClientRect().width + parseInt(spaceM, 10)
-      : 24;
+    this.setState({
+      checkboxWidth: checkbox
+        ? checkbox.getBoundingClientRect().width + parseInt(spaceM, 10)
+        : 17 + parseInt(spaceM, 10) // 17 is the width of our CheckboxInput component at the time of this change
```
GeorgiSTodorov: Can we make 17 a constant?

PR #449, opened 2019-12-10
Optional icon for SelectInput

GeorgiSTodorov:
Снимка от 2019-12-10 17-34-07


packages/selectInput/components/SelectInput.tsx
```diff
@@ -42,6 +42,10 @@ export interface SelectInputProps extends React.HTMLProps<HTMLSelectElement> {
    * hintContent is text or a ReactNode that is displayed directly under the input with additional information about the expected input.
    */
   hintContent?: React.ReactNode;
+  /**
+   * Optional icon to be displayed before the input value.
+   */
+  icon?: React.ReactElement<HTMLElement> | string;
```
mperrotti: Can we rename this prop to `iconStart`? That's the prop name we use for text inputs and buttons

packages/selectInput/style.ts
```diff
@@ -39,3 +39,7 @@ export const selectIcon = css`
   top: 50%;
   transform: translateY(-50%);
 `;
+
+export const optionalIcon = css`
+  margin: auto;
```
mperrotti: The icon is right up against the text. I recommend changing this to `margin: auto ${spaceXs} auto 0`

mperrotti:
@GeorgiSTodorov - Looks good! I just pushed up a commit so that the `iconStart` prop can accept an icon shape ID as well as an element

GeorgiSTodorov:
Thank you, @mperrotti .

PR #454, opened 2019-12-14
Apply an optional maximum width to a DropdownMenu

mperrotti:
## Screenshot
DropdownMenuMaxWidthDemo

### DIFF
--- a/packages/dropdownMenu/components/DropdownMenu.tsx
+++ b/packages/dropdownMenu/components/DropdownMenu.tsx
@@ -25,6 +25,10 @@ export interface DropdownMenuProps {
    */
   menuMaxHeight?: number;
   /**
+   * Maximum width the menu can grow to
+   */
+  menuMaxWidth?: number;
+  /**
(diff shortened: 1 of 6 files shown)


danielone:
@mperrotti 

1) thoughts on having small/med/large menu item width sizes like lightning design system ? 
https://lightningdesignsystem.com/components/menus/#Width

2) what are you thoughts on truncation if hits width vs. scrolling?

mperrotti:
@danielone
> thoughts on having small/med/large menu item width sizes like lightning design system ?

I like this, we just have to define what those sizes would be

> What are you thoughts on truncation if hits width vs. scrolling?

I'm against truncating because it hides the full string from the user

danielone:
@mperrotti 
yes agree, not blockers. 
RE defined sizes - great!  adding note to track and define this with you.
RE truncation - Do most DS systems do this now? I've typically seen truncation with full text on hover. happy to go with your pref and discuss further offline.

PR #458, opened 2020-01-06
fix: broaden version of react peerDeps

pierrebeitz:
npm frequently complains about the peerDependencies of ui-kit not being fulfilled as we (dcos-ui) use react `16.12.0`. is there any particular reason to have the restriction to `<16.9.0`?


pierrebeitz:
thanks for the timely review, @mperrotti! 🎉

PR #459, opened 2020-01-08
Set button type attribute to "button" by default

mperrotti:
These changes were made to accommodate buttons nested inside of a <form> element. Without the 'type' attribute set, buttons inside of a <form> element default to 'type=submit', which causes unwanted/premature form submissions. When kommander upgrades to a version of ui-kit that includes these changes, Jira ticket DCOS-62594 should be resolved.


GeorgiSTodorov:
Do we also want this for `ClickToCopy` button?

mperrotti:
@GeorgiSTodorov - we don't need it on that because it's not a `button`

PR #460, opened 2020-01-08
Some fixes for react warnings

brandonc:
## Overview
In some cases, I converted componentWillReceiveProps to componentDidUpdate or just added the UNSAFE_ prefix if a more involved refactor would be necessary.

I also added a couple of properties to relevant props lists so they wouldn't be rendered onto an input dom element.

It appears that many react warnings could be resolved by upgrading to emotion 10. I found a migration guide that doesn't look too crazy: https://emotion.sh/docs/migrating-to-emotion-10

## Testing
I only used the storybook which admittedly doesn't show react warnings. I did not test the legacy modal.

### DIFF
--- a/packages/appChrome/components/Sidebar.tsx
+++ b/packages/appChrome/components/Sidebar.tsx
@@ -53,12 +53,12 @@ const StyledNav = styled("nav")`
 `;
 
 class Sidebar extends React.PureComponent<SidebarProps, {}> {
-  public componentWillReceiveProps(nextProps: SidebarProps) {
+  public componentDidUpdate(prevProps: SidebarProps) {
     const { onOpen, onClose } = this.props;
 
-    if (nextProps.isOpen && onOpen) {
+    if (prevProps.isOpen && onOpen) {
       onOpen();
-    } else if (!nextProps.isOpen && onClose) {
+    } else if (!prevProps.isOpen && onClose) {
       onClose();
     }
   }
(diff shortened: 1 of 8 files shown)


GeorgiSTodorov:
@brandonc , unit tests are failing.

mperrotti:
Those failures appear to be a problem with the test. The component works as expected in Storybook. There's a few issues like this one filed in Enzyme's GitHub repo: https://github.com/airbnb/enzyme/issues/2203

I was able to hack around this by directly calling `componentDidUpdate({isOpen: true|false})` directly on the component instance. Anybody have a better idea? :/

GeorgiSTodorov:
@mperrotti , I think this is fine for now, we can change it to something better later if we figure something out.
@brandonc , I added a commit that renames `componentWillMount` and `componentWillUpdate`. Feel free removing it if you don't like it.

pierrebeitz:
While I understand that those errors are annoying they're quite helpful from my perspective in that they remind us of addressing some technical debt.   
For one the react team might want to remove `UNSAFE`-things at some point - which would force us to address things if we wanted to update.  
Another aspect is that there seems to be room for improvement - which could lead to better APIs of those components (I could see that for things having `open` and `onClose`-props). 


Just my 2 cents.

brandonc:
@pierrebeitz I respectfully disagree because the non-UNSAFE versions will be removed in React 17 but not the UNSAFE_ prefixed versions. There will no doubt be _another_ warning in React 17 before these are ultimately removed. This change keeps us out of the pit of despair for a bit longer.

PR #461, opened 2020-01-09
DCOS-62619: globally style all anchor tags

mperrotti:
Before:
Screen Shot 2020-01-09 at 5 34 46 PM

After:
Screen Shot 2020-01-09 at 5 34 25 PM

### DIFF
--- a/packages/shared/styles/normalize.ts
+++ b/packages/shared/styles/normalize.ts
@@ -1,0 +1,9 @@
+import { themeTextColorInteractive } from "../../design-tokens/build/js/designTokens";
+
+// TODO: follow-up with Jira ticket DCOS-62772 (a more thorough set of normalize styles)
+export const normalize = `
+    a {
+        text-decoration: none;
+        color: ${themeTextColorInteractive};
+    }
+`;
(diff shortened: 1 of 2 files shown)


packages/shared/styles/normalize.ts
```diff
@@ -0,0 +1,9 @@
+import { themeTextColorInteractive } from "../../design-tokens/build/js/designTokens";
+
+// TODO: follow-up with Jira ticket DCOS-62772 (a more thorough set of normalize styles)
+export const normalize = `
+    a {
+        text-decoration: none;
```
juliangieseke: I think I'd keep this underlined in normalize file 🤔 we're removing _all_ underlines otherwise.

GeorgiSTodorov: @juliangieseke , aren't we just removing all underlines for link elements?

juliangieseke: yeah, thats whats happening here. My questions was if thats actually what we want?

GeorgiSTodorov: I misunderstood your comment as saying that we are removing all underlines for all elements. I think removing it for all link elements is what we want, but @mperrotti can answer better.

mperrotti: Removing the underline for all links is intentional

PR #463, opened 2020-01-13
DCOS-59641: use flexbox for centering to avoid subpixel blur

mperrotti:
When centering using positioning and transform, sometimes modal content would appear a little blurry on low-res screens.

## Testing
See Jira.


pierrebeitz:
Woohoo! Thanks @mperotti!

PR #464, opened 2020-01-14
DCOS-62895: generate id value for components with optional id prop

mperrotti:
These changes are to ensure that we follow accessibility best practices by providing an ID to associate form inputs with related content.

Before, it was possible to leave the `id` prop off of some components.
Now, we generate a value for the `id` prop if one is not set.


package.json
```diff
@@ -55,12 +55,12 @@
     "react-draggable": "3.3.2",
     "react-emotion": "9.2.12",
     "react-focus-lock": "2.0.5",
+    "react-id-generator": "3.0.0",
```
mperrotti: I decided to use a different library so we don't unnecessarily re-create the ID when a component re-renders

juliangieseke:
Will these IDs change between builds? Or: can we rely on these in system tests?

GeorgiSTodorov:
@juliangieseke , I think it is a good idea to supply our own ids for system tests and in general.

packages/toggleBox/components/ToggleBox.tsx
```diff
@@ -1,5 +1,4 @@
 import * as React from "react";
-import shortid from "shortid";
```
GeorgiSTodorov: Why are we removing this?

GeorgiSTodorov:
@mperrotti , I am not sure why are we only changing a handful of components here.

mperrotti:
@GeorgiSTodorov - only a handful of components needed to be changed. Most of them require an ID

packages/toggleBox/components/ToggleBox.tsx
mperrotti: Because ID is actually required for this component, so this was unnecessary

mperrotti:
@juliangieseke - these will change between builds. I agree with Georgi that we should be supplying our own IDs for system tests.

PR #467, opened 2020-01-21
Keep whitespace in table

GeorgiSTodorov:
I am not sure if this is the best way to do it.
This is needed for https://jira.mesosphere.com/browse/DCOS-60700 .

Screenshot from 2020-01-21 14-19-46


mperrotti:
I don't think we should do this with `white-space: pre`. In cases where we want to force the browser to render multiple spaces, we should replace whitespace characters with a [non-breaking space](http://www.fileformat.info/info/unicode/char/00a0/index.htm).

GeorgiSTodorov:
Thank you, @mperrotti . I will explore that solution.

PR #479, opened 2020-02-26
chore: upgrade to ts 3.8

pierrebeitz:
and fix the type errors that occured along the way.


tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "esModuleInterop": true,
     "strict": true,
     "moduleResolution": "node",
+    "skipLibCheck": true,
```
pierrebeitz: was needed because: 

```

ERROR in node_modules/react-chartist/dist/index.d.ts(1,10):
TS2440: Import declaration conflicts with local declaration of 'ChartitGraphProps'.

```

packages/popover/components/PopoverListItem.tsx
```diff
@@ -84,33 +84,19 @@ const PopoverListItem = (props: PopoverListItemProps) => {
...
-          React.isValidElement<PopoverListItemIcon | PopoverListItemAvatar>(
-            child
-          ) &&
-          isItemGraphic(child) &&
-          child.props.position === "start"
+        child => isItemGraphic(child) && child.props.position === "start"
       );
       const itemGraphicEnd = itemChildren.find(
-        child =>
-          React.isValidElement<PopoverListItemIcon | PopoverListItemAvatar>(
```
Poltergeist: Why did you decide to drop the `isValidElement`?

GeorgiSTodorov: It's not dropped, it's moved - https://github.com/dcos-labs/ui-kit/pull/479/files#diff-cac05b4db96f03850867c34aed16603aR90 .

PR #480, opened 2020-02-26
refactor: remove immutable

pierrebeitz:
it seems overkill to have a big library for handling two places that need
immutable data.

this should trim down the size of the build a bit (like ~15kb gzipped on dcos-ui).


Poltergeist:
I am agreeing with Pierre here, it does not justify to have the complete library. Although if we really need immutable did we consider using "immer" instead of immutable.js?

PR #481, opened 2020-02-26
Adds a prop to open button links in a new tab

mperrotti:
There are already plans to improve how we deal with links in ui-kit ([D2IQ-61570](https://jira.d2iq.com/browse/D2IQ-61570)), but I'm not sure when that ticket will get worked on. 

This is an issue that causes enough headaches that I think this quick fix makes sense.


danielone:
@mperrotti  perhaps we build in that external icon link shows up by default in such a case but "escape hatch" to potentially turn off in case opening in new tab but not an external link.

mperrotti:
We haven't defined any rules around when we do or don't show the OpenExternal icon. I'd rather have us explicitly pass that icon until we know when that "escape hatch" might need to be used

PR #482, opened 2020-02-26
speed up tsc, clean up dependencies, remove a ts-ignore

pierrebeitz:
1. ```
   ➜  time npx tsc
   npx tsc  16.86s user 0.80s system 168% cpu 10.460 total
   ➜  time npx tsc
   npx tsc  2.62s user 0.26s system 187% cpu 1.535 total
   #                                         ^^^^^
   ```

1. updated @react/types to a more appropriate version
2. allow for `import React from "react"` via allowSyntheticDefaultImports to
   make the tests pass
3. remove the @ts-ignore line and follow along the type errors. also fix
   some issues that come to sight then.

--------------

then my PR became red due to tslint. now there's a commit introducing sensible default (what we have in kommander and dcos-ui but without stuff that requires a lot of fixing here), running prettier on all files (formerly not everything was prettified because of "tslint-config-prettier" not being the last in the extend-block and somehow not everything being tested by the jest-test (this also is the reason why so many classes changed).


package.json
```diff
@@ -84,12 +81,14 @@
     "@storybook/addons": "5.0.5",
     "@storybook/react": "5.0.5",
     "@storybook/storybook-deployer": "2.8.1",
+    "@types/chartist": "^0.9.46",
     "@types/enzyme": "3.1.14",
     "@types/jest": "23.3.5",
     "@types/memoize-one": "3.1.1",
     "@types/node": "^10.12.26",
-    "@types/react": "16.4.16",
+    "@types/react": "16.9.23",
```
pierrebeitz: tradeoff: we're using types for newer versions of react than our peerDeps state...

Poltergeist: That is a problem, we discussed increasing the react peer dependency but I can't recall what we decided on @mperrotti might have a better memory than I do.

mperrotti: We talked about it in the last monthly frontend sync. We're ok with increasing the React peer dependency 👍

packages/textInput/components/TextInput.tsx
```diff
@@ -43,7 +43,7 @@ export interface TextInputProps extends React.HTMLProps<HTMLInputElement> {
   /**
    * Sets the contents of the input label. This can be a `string` or any `ReactNode`.
    */
-  inputLabel: React.ReactNode;
+  inputLabel?: React.ReactNode;
```
GeorgiSTodorov: Why do we make this optional?

pierrebeitz: because we often times don't set it. specially when `showInputLabel` is set to false. maybe we can change the API some day, so that becomes more clear.

GeorgiSTodorov: Okay, you convinced me. :D
@mperrotti , what do you think?

pierrebeitz: the more prevalent problem here is that we don't test the types of our tests. without this change tests threw type errors at me.

package.json
Poltergeist: Can we then add a newer peer dependency here as well.

pierrebeitz: done

Poltergeist: Thanks, I was about to do it ;-) Thanks for  being faster then me.

PR #484, opened 2020-02-28
Table v2

pierrebeitz:
this adds a new table component that strives for:
* performance - the former table implementations tend to render too often. this
  table is designed to not do much work when items are updated or users
  scroll. it has not virtualization yet though! i would not recommend rendering > 500 rows.
* simple API - have a look at the API. it's way less to learn.
* simple implementation - it's a fraction of the code that the legacy or v1
  version uses. it's dependency free!
* type safety (try e.g. specifying an initialSort without a respective column!)

Although there's a plan to provide 
* examples on how to use persistence along with this table
* examples on how to build a checkbox-table
* a helper for headercells with tooltips
* a test helper based on to easily select with row/col-coordinates (for testing)

i think this is in a state that's good enough to use it.

note that you still need to take care of what data you feed it. you'll need to alter entries immutably for things to work out nicely.


GeorgiSTodorov:
Tested with 500 rows, it's much better.

mperrotti:
I've begun reviewing these changes, but I have a hard stop in 30 minutes. I plan to complete a thorough review of this first thing tomorrow.

mperrotti:
It might be nice to put all of the commits before "feat: add table v2" in a separate PR.

pierrebeitz:
> It might be nice to put all of the commits before "feat: add table v2" in a separate PR.

that's a good idea. i did that and stated it in the PRs description. any ideas on how to improve the recognizability of such remarks in a PRs description? i think i'll add a ⚠️ next time. btw: that PR is merged, so this PR looks cleaner now.

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <Row key={"row-" + config.toId(el)} config={config as Config} el={el} />
+      ))}
+    </div>
+  );
+}
+
+const Row = React.memo<{ config: Config; el: unknown }>(({ config, el }) => (
+  <>
+    {config.columns.map(({ id, render }) => (
+      <div key={config.toId(el) + id} className={style.cell}>
```
mperrotti: We could improve assistive technology UX by setting `aria-role="gridcell"` and selecting associating each cell with a column header via `aria-describedby`

packages/tablev2/style.ts
```diff
@@ -0,0 +1,85 @@
...
+) => css`
+  display: grid;
+  grid-template-columns: ${gridTemplateFragments.join(" ")};
+  position: relative;
+  overflow: scroll;
+  ${firstColSticky && stickyFirstCol(gridTemplateFragments.length)}
+`;
+
+const stickyFirstCol = (n: number) => `
+  .${th}:first-child, .${cell}:nth-child(${n}n + 1) {
```
mperrotti: Newer versions of emotion recommend against interpolating class names: https://github.com/emotion-js/emotion/blob/ca95f385f7ce3da6d53de1a652b3b219f11434c4/packages/serialize/src/index.js#L248.

packages/tablev2/style.ts
```diff
@@ -0,0 +1,85 @@
...
+    position: sticky;
```
mperrotti: If the table parent element does not have a height set, the header cells will not be sticky because the content will never have to scroll.

packages/tablev2/style.ts
```diff
@@ -0,0 +1,85 @@
...
+export const headerCss = css`
+  box-sizing: border-box;
+  border-top: 1px solid ${dt.themeBorderHeavy};
+  border-bottom: 1px solid ${dt.themeBorderHeavy};
+`;
+
+export const table = (
+  gridTemplateFragments: string[],
+  firstColSticky: boolean = false
+) => css`
```
mperrotti: We might want to set the row height with `grid-auto-rows`. By default, the rows will stretch (or shrink) to fill the height of their table parent element.

packages/tablev2/style.ts
mperrotti: That being said, I really love this use of the `:nth-child` selector 😍 .

packages/tablev2/style.ts
```diff
@@ -0,0 +1,85 @@
...
+  display: grid;
+  grid-template-columns: ${gridTemplateFragments.join(" ")};
+  position: relative;
+  overflow: scroll;
+  ${firstColSticky && stickyFirstCol(gridTemplateFragments.length)}
```
mperrotti: Is there a reason you chose to apply the styles this way instead of using emotion's `cx` function in the table element's `className` prop?

Then, you could do:
```jsx
<div
  className={cx(
    tableStyle,
    className,
    {[stickyFirstCol(state.columns.length)] : config.firstColSticky}
  )
} {...divProps} />
```

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,168 @@
+import * as React from "react";
+import * as style from "./style";
+import { Draggable, Sorter } from "./Util";
+
+type Config<A = unknown, B = string, C extends B = B> = {
```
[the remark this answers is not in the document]
pierrebeitz: much wow! i'll try that tomorrow!

packages/tablev2/Table.tsx
pierrebeitz: thanks for making that research! we should totally do that!

packages/tablev2/Util.tsx
```diff
@@ -0,0 +1,90 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+export type SorterFactory<T> = <B extends string, A extends Record<B, T>>(
+  prop: B
+) => Sorter<A>;
+
+const string: SorterFactory<string> = prop => dir => (a, b) =>
+  (dir === "asc" ? 1 : -1) * a[prop].localeCompare(b[prop]);
+
+const number: SorterFactory<number> = prop => dir => (a, b) =>
```
pierrebeitz: @mperrotti today i tried integrating the table into dcos-ui. when writing custom sorters i was annoyed by having to check the direction which is either "asc" or "desc". i'm thinking of just passing a `1` instead of "asc" and a `-1` instead of "desc" so that the parameter can directly be multiplied:

```
const sortByNameWithGroupsAlwaysOnTop = dir => (a, b) =>
  a.__type === b.__type
    ? dir * strComp(a, b) // formerly: (dir === "asc" ? 1 : -1) * strComp(a, b)
    : +(a.__type === "group");
```

packages/tablev2/style.ts
pierrebeitz: that explains the trouble i had today! thanks a lot!

packages/tablev2/style.ts
pierrebeitz: no! your suggestion looks so much better.

packages/tablev2/Util.tsx
mperrotti: I can't think of anything better than `dir` right now either 🤔

I'm a _little_ concerned that `1` and `-1` might not be as clear as `asc` and `desc`. However, I'm fine with that change if you and other reviewers think it's a better developer experience.

pierrebeitz:
@mperrotti regarding your observations:

> The sticky column is a pain to work with, and I'm going to ask the design team if we can just let the all of the cells scroll horizontally. For example: When we use Table v2 to create a checkbox table, it will be tricky keep the checkbox column and the first column of data cells "sticky".

i initially thought that we might be able to put the checkbox right into the first column - along with the other data (e.g. the name).

> It might be useful to add a README that includes the following information:

i'll add that once i've incorporated your feedback and the lessons learned from integrating this into dcos-ui.


> We should consider having each row wrapped in it's own div

i think of this as a really tough tradeoff and switched sides on that discussion (taking place in my head) several times already.  
today i'd lean towards the the row-solutions that enables hover-styles and the assistive hint, as i think
* developers should be able to provide reasonable initial values most of the time
* we could enable initial auto-sizing by initially rendering invisibly without rows to then set the "optimal" widths that `max-content` figured out for us. i'm not suggesting that we SHOULD implement this. i'm just saying that we have options in case we go with the row-approach.

packages/tablev2/Util.tsx
GeorgiSTodorov: I prefer "asc" and "desc", but maybe we can accept both?

package.json
```diff
@@ -45,6 +45,7 @@
     "downshift": "3.2.10",
     "emotion": "9.2.12",
     "emotion-theming": "9.2.9",
+    "faker": "^4.1.0",
```
GeorgiSTodorov: What does faker do? We only use it in one place.

GeorgiSTodorov: I see, it's to create fake data.

packages/tablev2/Util.tsx
pierrebeitz: in `sortByNameWithGroupsAlwaysOnTop` how would you either have dir be `"asc" | "desc"` or `1 | -1`?
for the sake of clarity, i'd vow to not give consumers much freedom on how to use this.

package.json
pierrebeitz: we don't necessarily need that in ui-kit and i can remove that for sure before merging unless @mperrotti or @Poltergeist think it would be worthwhile to have that in.

i thought it might be cool to have lots of different data an then be forced to write content-agnostic tests.

packages/tablev2/Util.tsx
GeorgiSTodorov: If I had to choose one, I would choose "asc" and "desc".

package.json
GeorgiSTodorov: I think it's better to remove it before merging.

mperrotti: I think it might be nice to keep it in for testing purposes, but it's not totally necessary.

mperrotti:
@pierrebeitz

> i initially thought that we might be able to put the checkbox right into the first column - along with the other data (e.g. the name).

Ok, that sounds good to me. The only drawback is that the column header would be left-aligned with the checkbox, not with the first column's text. If the design team has an issue with that, we can get around it by adding extra left padding to that header column :/

I'm in agreement with you that we're not ready for the README, and that each row should have a parent div.

mperrotti:
Just a heads up - I'm working on addressing some of my own feedback. Right now I'm working on updating the markup structure so that each row is it's own element.

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,265 @@
...
+        stickyFirstCol={stickyFirstCol}
+        state={state}
+        update={update}
+      />
+
+      {data.sort(sort).map(el => (
+        <Row
+          key={"row-" + toId(el)}
+          // TODO: figure out how to pass a generic while using React.memo
+          // @ts-ignore
```
mperrotti: Mighty Typescript wizard @pierrebeitz - any ideas? :)

packages/tablev2/Table.tsx
```diff
@@ -91,8 +101,14 @@ function HeaderCell({ config, id, update, state }: HeaderCellProps) {
...
 
   return (
-    <div className={style.th} ref={cell} onClick={onClick} role="button">
-      <div className={sortClass}>{config.header}</div>
+    <div
+      className={style.headerCell}
+      id={id}
+      ref={cell}
+      onClick={onClick}
+      role="columnheader"
```
pierrebeitz: according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role we might want to add an `aria-sort` here, do we?

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const update = (x: Partial<State>) => {
+    setState({ ...state, ...x });
+    config.onStateChange?.({ ...state, ...x });
+  };
+  const col = config.columns.find(c => c.id === state.sortBy);
+  const sort = col?.sorter?.(state.order) || (() => 0);
+
+  // We currently need to cast Config to Config<unknown> as to a bug in typescript's handling of unknowns. We actually NARROW the interface, which is good.
+  return (
+    <div className={tableStyle} {...divProps}>
```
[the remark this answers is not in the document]
pierrebeitz: i actually thought that it might be nice to be able to override this. but i'm fine with what is currently implemented.

packages/tablev2/style.ts
pierrebeitz: thanks for fixing! 💜

packages/tablev2/Table.tsx
pierrebeitz: The issue is no more! types should look much more documenting now.

pierrebeitz:
Hi @mperrotti, thanks for driving this forward.

I polished stuff a bit (code- / type-wise) and cleaned up the discussions within this PR. If you like my commits, feel free to squash stuff now. I definitely approve of all the changes you added.

mperrotti:
@pierrebeitz - I'll have a look today. Thanks for doing some cleanup!

mperrotti:
@pierrebeitz - I rebased from master and pushed a commit that moves the column configuration into `<Column>` children.

When/if you have time, there are a few Typescript issues I'd like to pair on.

I've started testing the new table on this branch in Kommander: https://github.com/mesosphere/kommander/pull/new/mp/tableV2Test

pierrebeitz:
@mperrotti love it at first sight! will have a look at the TS-issues hopefully later this day. 

after having a look at the kommander-integration branch i realized that sorters are not used the way i intended them to be used (the direction is currently ignored). that raises the question of whether the API is too complex. i feel tempted to boil it down like this:

```
type Sorter = <A>(dir: "asc" | "desc") => (a: A, b: A) => 1 | 0 | -1
=> 
type Sorter = <A>(a: A, b: A, dir?: "asc" | "desc") => 1 | 0 | -1
```

we'll need access to the direction in DCOS-UI in order to keep `Groups` or "folders" on top on the services page, regardless of sorting up/down. i'd change the implementation to automatically multiply by `1` or `-1` depending on the currently set sorting order. the groups-case could then be solved by a sorter like this: 

```
const sortByNameWithGroupsOnTop = (a, b, dir) => 
  if (a.__type === b.__type) { // both are either groups or apps
    return a.name.localeCompare(b.name);
  }
  return (a.__type === "group" && dir === "asc") ? -Infinity : Intifinity
}
```
so we'd "add" or "substract" a lot to have groups always on top.

we formerly could do this:
```
const sortByNameWithGroupsOnTop = dir => (a, b) => {
  if (a.__type === b.__type) { // both are either groups or apps
    return (dir === "asc" ? 1 : -1) * a.name.localeCompare(b.name);
  }
  return a.__type === "group" ? 1 : -1; // groups are always "before" other things.
}
```

I think it's better to make the more common path more accessible. By that i mean that i like how you wrote the sorters in the kommander-branch to reflect an ascending order and not having to deal with the `dir` for 99% of the cases.

Sounds right?

mperrotti:
@pierrebeitz

I actually like the current API:
```
type Sorter = <A>(dir: "asc" | "desc") => (a: A, b: A) => 1 | 0 | -1
```.

Your groups/apps sorting examples make sense to me, and I agree with you that the second example is more explicit and straightforward

File: packages/tablev2/Column.tsx [hunk 12]
```diff
@@ -0,0 +1,44 @@
...
+
+export type ColumnProps<A, B = string> = {
+  /**
+   * An id given to that column. Needed for persistence.
+   */
+  id: B;
+  /**
+   * Renderer for the cells of that column.
+   */
+  render: (a: A) => React.ReactNode;
```
mperrotti: I think we should rename this to `renderer`. It could be confused with [render props](https://reactjs.org/docs/render-props.html)

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,340 @@
...
+import { cx } from "emotion";
+import * as style from "./style";
+import { typedMemo } from "../../typings/typedMemo";
+import { CheckboxCell, Draggable, HeaderCellContent } from "./Util";
+import { ColumnProps } from "./Column";
+import { ResetButton } from "../button";
+
+type ColumnChild<A, B> = React.ReactElement<ColumnProps<A, B>>;
+
+// TODO: make it possible to disable rows from being selected with a checkbox
```
mperrotti: I'm not sure if we need this for the MVP of Table V2. Reviewer/s: would you be comfortable with creating a follow-up Jira ticket for this? Or do you think it should be done now?

pierrebeitz: totally not a priority for dcos-ui.

packages/tablev2/PrimaryLinkCell.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from "react-emotion";
+import { textWeight } from "../shared/styles/styleUtils";
+import { linkReset } from "../shared/styles/styleUtils/resets/linkReset";
+import { themeTextColorPrimary } from "../design-tokens/build/js/designTokens";
+
+export const PrimaryLinkCell = styled("div")`
```
pierrebeitz: what's the rationale of forcing users of the table to use a `PrimaryLinkCell` in case they want to render a link. i think it would feel good if the table styled all links like this by default, without me having to wrap things in yet another component - because i think there's rarely a case where i want to maintain the default link styling within a table.

mperrotti: Only links in the first column are meant to have a heavier font weight.

Maybe instead of using a component, we could style anchor tags that are children of the cells. Something like:
```
[role="gridcell"] a {
    ${linkReset};
}

[role="gridcell"]:first-child a {
    ${textWeight("medium")};
}
```

What do you think?

pierrebeitz: i'd actually love that!

packages/tablev2/Util.tsx
```diff
@@ -144,11 +144,15 @@ export type SorterFactory<T> = <B extends string, A extends Record<B, T>>(
...
-const string: SorterFactory<string> = prop => (_, dir) => (a, b) =>
-  dir * a[prop].localeCompare(b[prop]);
+const string: SorterFactory<string | null | undefined> = prop => (_, dir) => (
+  a,
+  b
+) => dir * (a[prop] || "")!.localeCompare((b[prop] || "")!);
 
-const number: SorterFactory<number> = prop => (_, dir) => (a, b) =>
-  dir * (a[prop] - b[prop]);
+const number: SorterFactory<number | null | undefined> = prop => (_, dir) => (
```
pierrebeitz: i'm not sure about this change.

i'd find it more useful if the `null`s and `undefined`s would be always last (like the `sortGroupsAlwaysOnTop`-thing from DCOS-UI's ServiceTable). What do you think?

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * A callback invoked whenever the internal state of the table is changed.
+   * Have a look at the State type to see what's in there.
+   */
+  onStateChange?: (a: State) => void;
+
+  /**
+   * An object where the keys are the IDs of selected rows, and the value is a boolean
+   */
+  selectedRows?: {};
```
pierrebeitz: how often is kommander using this?
i initially thought of solving this in userland by providing examples/helpers.

asking as it adds quite some complexity and currently rerenders every cell when something in the table updates.

packages/tablev2/Util.tsx
mperrotti: I agree with you. How else might we change these functions so they accept properties from `data` that could be null or undefined?

packages/tablev2/Table.tsx
mperrotti: Kommander doesn't use this at all yet, but I think there are a few places in DC/OS that use this.

The re-render is definitely undesirable...maybe we could just figure out how to have selectable rows without baking it into the component, and add a Storybook example?

Curious to hear what you and other reviewers think about this.

mperrotti: @pierrebeitz - Are you sure it's re-rendering every cell? I just tested it, and I'm only seeing the first column and the header row re-render on update.

pierrebeitz: Oh, I'm sorry. I meant to say "every cell of the first column".

mperrotti:
@pierrebeitz - I accidentally force-pushed before pulling down your typed memo commit, `c17c1f3e86c65e96e742f1ebd025363fdd9421ff`. :(

Could you push those changes back up?

packages/tablev2/Column.tsx
GeorgiSTodorov: I think this makes sense.

packages/tablev2/tablev2.stories.tsx
```diff
@@ -77,6 +85,82 @@ storiesOf("Table V2", module)
...
+
+        const toggle = x =>
+          React.useCallback(() => dispatch({ type: "toggle", item: x }), [x]);
+
+        return (
+          <Table
+            data={items}
+            toId={el => el.id.toString()}
+            columns={[
+              // THIS IS THE COLUMN YOU WANT TO ADD ALONG WITH THE CEREMONY ABOVE.
```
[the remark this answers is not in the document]
pierrebeitz: oh yeah... we can of course choose to put that stuff right into the first column. with the same markup you had before. i like that!

packages/tablev2/tablev2.stories.tsx
```diff
@@ -77,6 +85,82 @@ storiesOf("Table V2", module)
...
+                    inputLabel="Toggle all rows"
+                    onChange={() =>
+                      dispatch({
+                        type: headerChecked ? "deselectAll" : "selectAll"
+                      })
+                    }
+                  />
+                ),
+                // the width of a checkbox + cell-padding ( + currently 8px to not make the drag handle hide parts of the checkbox. shall we make its bg transparent?)
+                initialWidth: "44px",
```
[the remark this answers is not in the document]
pierrebeitz: yeah, it would. also i totally don't like the math behind it currently, BUT your remark above resolves this as well. there's no need to specify a width, if we bake that into the first "real" column.

GeorgiSTodorov:
Looks great.
Some minor things I noticed:
- I think header cells need to be bold.
- In Chrome, there is an unscrollable scroll.
Screenshot from 2020-03-24 15-20-26
- In the checkbox table, whenever a row is updated, the checkbox on that row becomes unselected. Also, selecting all rows does not select the top row. Also, if we select the top row and then some row becomes unselected because it's updated, the top row does not become unselected. Is this on purpose?
Peek 2020-03-24 15-25
- I can resize the columns to be too small.
Peek 2020-03-24 15-27
- In the sortable table, the initial sort is in descending order, I think it should be in ascending order.
Screenshot from 2020-03-24 15-30-29
- I don't see anything special in the "selectable rows" story.
- In the "cells rendered w/ React components" story, resizing a column also moves it. Is this intentional?
Peek 2020-03-24 15-40

mperrotti:
@GeorgiSTodorov thanks for reviewing!


> ]I think header cells need to be bold.

They're supposed to be medium weight

---

> In Chrome, there is an unscrollable scroll.

Fixed

---

> In the checkbox table, whenever a row is updated, the checkbox on that row becomes unselected. Also, selecting all rows does not select the top row. Also, if we select the top row and then some row becomes unselected because it's updated, the top row does not become unselected. Is this on purpose?

This is not intentional. I'm going to remove this story for now because we've decided to take care of checkbox table functionality as a follow-up. Here is a Jira ticket: [D2IQ-65975](https://jira.d2iq.com/browse/D2IQ-65975)

---

> I can resize the columns to be too small.

I changed the minimum column width from 20px to 50px

---

> In the sortable table, the initial sort is in descending order, I think it should be in ascending order.

I agree. Fixed.

---

> I don't see anything special in the "selectable rows" story.

This story should have been removed when we took away the checkbox functionality.

---

> In the "cells rendered w/ React components" story, resizing a column also moves it. Is this intentional?

This is intentional, and it also happens in other stories (for example, "default").

GeorgiSTodorov:
@mperrotti , thanks for addressing my feedback.

GeorgiSTodorov:
- We don't show an ellipsis for some columns. Intentional?
Screenshot from 2020-03-25 10-36-12

- We don't have a browser tooltip like we did in the old table.

- In the story "without a sticky first column", we still have a border between the first and second column. Intentional?
Screenshot from 2020-03-25 10-40-35

packages/tablev2/Table.tsx
```diff
@@ -0,0 +1,306 @@
...
+) =>
+  cx(style.row(gridTemplateFragments), {
+    [style.rowWithStickyColumn]: stickyFirstCol
+  });
+
+type HeaderCellProps<Entry> = {
+  column: Column<Entry>;
+  selectableRowIds?: string[];
+  id: string;
+  isFirstColumn: boolean;
```
GeorgiSTodorov: Why do we need this?

pierrebeitz: we don't. it's a leftover from the selectableRow-functionality.

pierrebeitz:
> We don't show an ellipsis for some columns. Intentional?

yes, the initial idea was to go with the bare minimum in terms of implementation/performance costs. if one really wants to know what's in a truncated cell, they can resize the column. i mainly wanted to cut on that because i believe that the old table sometimes does not scroll as intended when the tooltip component "portals" the scroll event to outside of the table.

> We don't have a browser tooltip like we did in the old table.

i think this is answered, right?

> In the story "without a sticky first column", we still have a border between the first and second column. Intentional?

it looks like this in intentional for now and will be iterated upon: https://github.com/dcos-labs/ui-kit/pull/484/commits/4af8e64d59d4952923a30bf88e9597502b06b6c4

mperrotti:
@GeorgiSTodorov - I can't reproduce the text truncation bug in Chrome, FF, or Safari 🤔

I removed the the `isFirstColumn` prop

GeorgiSTodorov:
@mperrotti 

> I can't reproduce the text truncation bug in Chrome, FF, or Safar

I can in Firefox.

### Firefox
Снимка от 2020-03-27 10-04-59

### Chrome
Снимка от 2020-03-27 10-04-39

mperrotti:
@GeorgiSTodorov - that resize behavior definitely isn't great, but I think that's edge case-y enough to take care of as a follow up.

The only way a user would encounter this is if they go out of their way to do something weird.

@pierrebeitz - how would you feel about adding a `resizable` property to the column config? It would still default to `true`, but it could be set to `false` for cases like the dropdown column

package.json
brandonc, line 48: Could probably be a devDependency

packages/index.ts
```diff
@@ -95,6 +95,12 @@ export {
   NumberCell
 } from "./table";
 export { TableView, TableViewHeader, TableViewBody } from "./tableView";
+export {
+  Table as TableV2,
```
brandonc, line 99: I wonder if we should rename Table as TableV1 or Table_Deprecated and then do a major version release so we don't have to live with TableV2 as a name forever. Let clients rename if they want to keep the old table.

packages/tablev2/README.md
```diff
@@ -0,0 +1,25 @@
+# Table v2
+
+A v2 of the Table component was created to improve performance issues that were impossible to get around without eliminating core features and design specifications. For example: column sizing features, row hover styles, and horizontal scrolling styles.
+
+Our goal is to gradually replace Table v1 usage with Table v2, deprecate Table v1, and eventually remove Table v1 and it's dependencies from ui-kit.
```
brandonc, line 5: Another good reason so not ship this as "TableV2"

GeorgiSTodorov:
@brandonc 

> I think we should ship the component as Table and rename the old table (along with a major version bump) so we don't have to rename again. But I don't feel so strongly as to hold merge...

Wouldn't that be a breaking change?

pierrebeitz:
@GeorgiSTodorov 

> Wouldn't that be a breaking change?

yes it would. thus the suggestion to go with a major version bump.

packages/utilities/useIntersect.ts
```diff
@@ -0,0 +1,39 @@
...
+  threshold?: number[] | number;
+}): [
+  React.MutableRefObject<HTMLDivElement | null>,
+  IntersectionObserverEntry | undefined
+] => {
+  const [entry, updateEntry] = React.useState<IntersectionObserverEntry>();
+  const node = React.useRef<HTMLDivElement>(null);
+
+  const observer = React.useRef(
+    new window.IntersectionObserver(([entry]) => updateEntry(entry), {
```
pierrebeitz, line 20: this is pure magic! 😍 

you're a wizard!

packages/index.ts
pierrebeitz: i like that approach.

packages/tablev2/Util.tsx
```diff
@@ -0,0 +1,161 @@
...
+        />
+      </div>
+    </div>
+  );
+};
+
+///////////////////////////////////////////////////////////////////////////////
+//                            HEADER CELL CONTENT                            //
+///////////////////////////////////////////////////////////////////////////////
+export const HeaderCellContent = ({
```
pierrebeitz, line 80: @mperrotti could we follow this "composability over configuration" school of thought here to remove complexity from the table itself by handing users a `tooltipify`-function or a `TooltipCell`-component or something like that?

packages/tablev2/Util.tsx
```diff
@@ -0,0 +1,161 @@
...
+  a,
+  b
+) => {
+  // null/undefined sort after anything else
+  if (a[prop] == null) {
+    return 1;
+  }
+  if (b[prop] == null) {
+    return -1;
+  }
```
pierrebeitz, line 145: @brandonc thanks a lot!

if we don't compare strictly here, wouldn't `0`s be somewhere amongst the null/undefineds?

pierrebeitz: no, i was fooled by the falsiness concept once again 🤦‍♂

packages/index.ts
mperrotti: I'm on board with this 👍

packages/tablev2/Util.tsx
mperrotti: You're right, "composability over configuration" is the way to go 👍

pierrebeitz:
@mperrotti shall we squash most commits (preserving the interesting bits of the commit messages) and merge?

GeorgiSTodorov:
@pierrebeitz , @mperrotti , do we merge this?

pierrebeitz:
We do. If it was for me, we'd squash things first though.

mperrotti:
@pierrebeitz -yea, let's do some squashing. I can take care of that today

mperrotti:
I rebased the commits and it dismissed the 2 approvals.

🎉**This can be merged as soon as CI passes** 🎉

PR #488, opened 2020-03-04
Create design Guidelines directory

Poltergeist:


GeorgiSTodorov:
Don't we already have design guidelines for each component?

Poltergeist:
In this directory we will have more general guidelines, eventually.

PR #491, opened 2020-03-06
iterate on toasts

pierrebeitz:
because i wanted to use toasts in DCOS-UI and was interested in how stuff works, i played around with some code and thought it's worthwhile to share the results.

### DIFF
--- a/packages/toaster/components/Toast.tsx
+++ b/packages/toaster/components/Toast.tsx
@@ -13,12 +13,13 @@ import {
   flex,
   flexItem,
   margin,
-  padding,
-  tintSVG
+  padding
 } from "../../shared/styles/styleUtils";
 import Clickable from "../../clickable/components/clickable";
-import { green, red, yellow } from "../../design-tokens/build/js/designTokens";
+import * as dt from "../../design-tokens/build/js/designTokens";
 import { CloseIcon } from "../../shared/icons";
+import { Icon } from "../../icon";
+import { SystemIcons } from "../../icons/dist/system-icons-enum";
 
 export type ToastId = React.ReactText | undefined;
 
@@ -26,7 +27,7 @@ export interface ToastProps {
   title: React.ReactElement<HTMLElement> | string;
   description?: React.ReactElement<HTMLElement> | string;
   id: ToastId;
-  appearance?: "default" | "success" | "warning" | "danger";
+  appearance: "default" | "success" | "warning" | "danger";
   autodismiss?: boolean;
   onDismiss?: (event?: React.SyntheticEvent<HTMLElement>) => void;
   dismissToast?: (dismissedToastId: ToastId) => void;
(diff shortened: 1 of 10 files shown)


packages/toaster/components/Toast.tsx
```diff
@@ -26,21 +26,21 @@ export interface ToastProps {
   title: React.ReactElement<HTMLElement> | string;
   description?: React.ReactElement<HTMLElement> | string;
   id: ToastId;
-  appearance?: "default" | "success" | "warning" | "danger";
+  appearance: "default" | "success" | "warning" | "danger";
```
brandonc: ~Is this a breaking change?~

I see the default now

packages/table/stories/Table.stories.tsx
```diff
@@ -467,37 +467,37 @@ storiesOf("Table", module)
     >
       <Table data={items}>
         <Column
-          header={<HeaderCell lowerCase={true}>name</HeaderCell>}
+          header={<HeaderCell capitalize={false}>name</HeaderCell>}
```
GeorgiSTodorov: Thank you for this.

GeorgiSTodorov:
@pierrebeitz , the lint check is failing.

pierrebeitz:
thanks for having a look, @GeorgiSTodorov!

> Not related to this PR, but "with dismiss callback" doesn't seem to do anything.

i think there should be a message in the "actions"-panel when you click dismiss. is that not working?

GeorgiSTodorov:
> i think there should be a message in the "actions"-panel when you click dismiss. is that not working?

I don't remember if I checked there, I was expecting either an alert or a console.log.

packages/toaster/components/Toast.tsx
```diff
@@ -26,21 +26,21 @@ export interface ToastProps {
...
   autodismiss?: boolean;
   onDismiss?: (event?: React.SyntheticEvent<HTMLElement>) => void;
   dismissToast?: (dismissedToastId: ToastId) => void;
   primaryAction?: React.ReactElement<HTMLElement>;
   secondaryAction?: React.ReactElement<HTMLElement>;
 }
 
-const SuccessIcon = props => (
+const SuccessIcon = (
```
mperrotti: We could replace these with the Icon component. Toast components were created before we had the Icon component.

packages/toaster/components/Toast.tsx
```diff
@@ -49,13 +49,13 @@ const SuccessIcon = props => (
   </svg>
 );
 
-const DangerIcon = props => (
+const DangerIcon = (
```
mperrotti: This could also be replaced using the Icon component

packages/toaster/components/Toast.tsx
```diff
@@ -64,13 +64,13 @@ const DangerIcon = props => (
   </svg>
 );
 
-const WarningIcon = props => (
+const WarningIcon = (
```
mperrotti: This could also be replaced using the Icon component

packages/toaster/components/Toaster.tsx
```diff
@@ -5,22 +5,21 @@ import { ToastProps, ToastId } from "./Toast";
...
+
+interface ToasterProps {
+  children?: Child | Child[];
+}
+
+class Toaster extends React.PureComponent<
+  ToasterProps,
+  { toasts: Array<React.ReactElement<ToastProps>> }
+> {
+  public timeouts: NodeJS.Timeout[] = [];
```
mperrotti: Nice. I didn't know about NodeJS.Timeout

typings/react.d.ts
```diff
@@ -0,0 +1,7 @@
+namespace React {
+  interface ReactChildren {
+    // we sometimes want to handle typed children, which react does not support out of the box.
+    // instead of jumping through hoops, we introduce this parameterized version of toArray.
+    toArray<A>(children?: A | A[]): A[];
```
mperrotti: Thanks! This will be useful

packages/toaster/components/Toast.tsx
pierrebeitz: i wanted to do that but storybook was not open at that moment - so i just compared the svg-paths which did not match up and abandoned the idea.

thanks a lot for paying close attention here! 😍

pierrebeitz:
@mperrotti addressed your feedback. i'd hit the button, but am not sure about what the conventions in ui-kit are. would it be ok to self-merge if there's only been small changes and like 3 people approved before?

mperrotti:
@pierrebeitz - it's ok to self-merge :)

PR #493, opened 2020-03-06
D2IQ-64873: Stack component

mperrotti:
Closes [D2IQ-64873](https://jira.d2iq.com/browse/D2IQ-64873)

## Screenshot
Screen Shot 2020-03-06 at 5 44 37 PM

### DIFF
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -115,7 +115,13 @@ export { Legacy };
 export { CodeSnippet } from "./codesnippet";
 
 // Style utility components
-export { Container, Flex, FlexItem, GridList } from "./styleUtils/layout";
+export {
+  Container,
+  Flex,
+  FlexItem,
+  GridList,
+  Stack
+} from "./styleUtils/layout";
 export { Box, BorderedBox, SpacingBox } from "./styleUtils/modifiers";
 export {
   Text,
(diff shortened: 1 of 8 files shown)


GeorgiSTodorov:
If we have `<li>`s inside an element that is not ul or ol, then the dots are touching the left border. Is this okay?
Screenshot from 2020-03-09 10-10-14

pierrebeitz:
i'd approve code-wise.

most likely you have a compelling use case in kommander. naively i'd not see a lot of use cases for this or would even look at something called a `Stack` if i'd search for the functionality provided by this component.

most ui frameworks have something like a `List`. but that's usually leaning towards single-line items, which is not the use case for this, or is it?

from how i have worked with UI-Kit i would not look for this but rather roll my own "Stack" with the help of `SpacingBox`. But that might be my "functional" mindset that wants to reuse rather small tools.

this feedback is not blocking in any way.

mperrotti:
Thanks for the feedback @pierrebeitz. It's totally acceptable to use the `SpacingBox` component, but sometimes we don't want to have to wrap every child with a `SpacingBox`, we just want all the children to get the same space between them.

> most ui frameworks have something like a List. but that's usually leaning towards single-line items, which is not the use case for this, or is it?

We have our own `List` components. This is not the use case for this component

PR #494, opened 2020-03-09
Fix PrimaryButton with url color

GeorgiSTodorov:
Make sure to test in both Firefox and Chrome as some of the solutions that I tried worked only in Chrome.

### Before
Screenshot from 2020-03-09 14-53-50

### After
Screenshot from 2020-03-09 14-44-31

### DIFF
--- a/packages/button/components/ButtonBase.tsx
+++ b/packages/button/components/ButtonBase.tsx
@@ -23,10 +23,6 @@ import {
   processingTextStyle
 } from "../style";
 import { SharedLinkProps } from "../../link/types";
-import {
-  linkReset,
-  primaryLinkReset
-} from "../../shared/styles/styleUtils/resets/linkReset";
 
 export enum ButtonAppearances {
   Primary = "primary",
@@ -115,7 +111,6 @@ class ButtonBase extends React.PureComponent<ButtonBaseProps, {}> {
       textWeight("medium"),
       className,
       {
-        [appearance === "primary" ? primaryLinkReset : linkReset]: url,
         [fullWidthButton]: isFullWidth,
         [buttonInverse(appearance)]: isInverse,
         [getMutedButtonStyles(appearance)]: disabled || isProcessing,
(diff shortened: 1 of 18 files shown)


packages/button/stories/DefaultButton.stories.tsx
```diff
@@ -242,6 +242,46 @@ storiesOf("Buttons/Default", module)
       }
     }
   )
+  .add(
+    "PrimaryButton used as a link",
```
mperrotti: I don't think this story is necessary. If we want to show multiple button appearances being used as a link, I think we should do it in one story

packages/button/components/ButtonBase.tsx
```diff
@@ -112,7 +115,7 @@ class ButtonBase extends React.PureComponent<ButtonBaseProps, {}> {
       textWeight("medium"),
       className,
       {
-        [linkReset]: url,
+        [appearance === "primary" ? primaryLinkReset : linkReset]: url,
```
mperrotti: I figured out a simpler way to do this. I'll push up a commit.

packages/button/stories/DefaultButton.stories.tsx
GeorgiSTodorov: The problem is that PrimaryButton used as a link didn't have the expected color.

mperrotti:
Ignore my last commit, I screwed something up. I'll push a fix

packages/button/stories/DefaultButton.stories.tsx
mperrotti: I understand why you're showing the PrimaryButton with a `url` prop, but I think we should just use one story to show different button styles with `url`.

GeorgiSTodorov: Okay, I agree.

mperrotti: I pushed a commit to take care of Storybook updates.

PR #498, opened 2020-03-12
SidebarItem data cy

GeorgiSTodorov:
This attempts to fix two problems:
- Ugly border in Firefox for `SidebarItem`s with url-s.
### Before
Screenshot from 2020-03-12 14-10-13

- No data-cy for `SidebarItem`s.
### Before
Screenshot from 2020-03-12 14-11-55

### After
Screenshot from 2020-03-12 14-08-23

A third problem that I could not fix was that clicking on the left or right edge of a `SidebarItem` with url does nothing since it's the padding of the parent `<li>` element.
Peek 2020-03-12 13-50

### DIFF
--- a/packages/shared/styles/styleUtils/resets/linkReset.ts
+++ b/packages/shared/styles/styleUtils/resets/linkReset.ts
@@ -9,6 +9,6 @@ export const linkReset = css`
   }
 
   :focus {
-    outline:none;
+    outline: none;
   }
 `;
--- a/packages/shared/styles/styleUtils/resets/linkReset.ts
+++ b/packages/shared/styles/styleUtils/resets/linkReset.ts
@@ -7,8 +7,4 @@ export const linkReset = css`
   &:visited {
     color: inherit;
   }
-
-  :focus {
-    outline: none;
-  }
 `;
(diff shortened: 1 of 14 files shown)


pierrebeitz:
updated test again to make CI pass. i think we should not have 4 commits here. 1 or 2 should be enough :)

GeorgiSTodorov:
Thanks, @pierrebeitz .

GeorgiSTodorov:
```
ERR! docker container or on a CI, you need to pass the '--ci' flag to prevent opening a

ERR! browser by default.
```
This is weird. Has anyone seen this before?

packages/shared/styles/styleUtils/resets/linkReset.ts
```diff
@@ -7,4 +7,8 @@ export const linkReset = css`
   &:visited {
     color: inherit;
   }
+
+  :focus {
```
mperrotti: I'm strongly opposed to removing the focus outline on links. We should only hide it in specific cases where we have our own focus styles.

I'll push up a commit with my suggestion for how we should handle this for the sidebar items

GeorgiSTodorov:
Thanks, @mperrotti .

PR #502, opened 2020-03-19
Support more button props

mperrotti:
Instead of accepting all possible props for an HTML button element, I added a few that I think would be the most useful:
- aria-label
- onFocus
- onBlur

Relates to D2IQ-46033

---

Note: The first commit contains the button prop updates. There is a second commit in this PR that removes the placeholder icons we were using before we had an icon library. I noticed them while updating button stories.


packages/button/components/ButtonBase.tsx
```diff
@@ -37,6 +37,10 @@ export interface ButtonProps extends SharedLinkProps {
    * if the button triggers new content to appear (e.g.: modals and dropdowns)
    */
   ariaHaspopup?: boolean;
+  /**
+   * should be used if the button does not contain text children. For example: a button that is just an icon
+   */
+  ariaLabel?: string;
```
pierrebeitz: it looks like i can provide whatever aria value if i use kebap-case - like `aria-label="test"`.

that's mainly an observation. i also have mixed feelings about finding that prop in the interface while not having an explicit usage.

packages/toaster/components/Toast.tsx
```diff
@@ -139,7 +138,11 @@ class Toast extends React.PureComponent<ToastProps, {}> {
         )}
         <Clickable tabIndex={0} action={this.handleDismiss}>
           <span className={toastDismiss}>
-            <CloseIcon />
+            <Icon
+              shape={SystemIcons.Close}
+              color="inherit"
+              size={dt.iconSizeXs}
```
pierrebeitz: sometimes it looks like Xs should be the default size. 🤔

GeorgiSTodorov:
@mperrotti , do we merge this?

PR #503, opened 2020-03-25
D2IQ-64883: improve tooltip keyboard nav

mperrotti:
## Screenshots (focusing using the keyboard)
Before:
TooltipKeyboardNavDemo-before

After:
TooltipKeyboardNavDemo-after

### DIFF
--- a/packages/utilities/getFirstFocusableChildNode.ts
+++ b/packages/utilities/getFirstFocusableChildNode.ts
@@ -1,0 +1,29 @@
+// https://github.com/theKashey/focus-lock/blob/master/src/utils/tabbables.js
+const focusableNodeSelectors = [
+  "button:enabled:not([readonly])",
+  "select:enabled:not([readonly])",
+  "textarea:enabled:not([readonly])",
+  "input:enabled:not([readonly])",
+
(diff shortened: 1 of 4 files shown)


mperrotti:
@GeorgiSTodorov
Before these changes, you would have to tab twice to focus on a button or link trigger. See in the first GIF how a blue outline appears around the button, then the purple border appears around the button

packages/utilities/getFirstFocusableChildNode.ts
```diff
@@ -0,0 +1,29 @@
...
+  "area[href]",
+
+  "iframe",
+  "object",
+  "embed",
+
+  "[tabindex]",
+  "[contenteditable]",
+  "[autofocus]"
+];
```
brandonc: You can make this more efficient by just calling .join(",") here once instead of twice each time getFirstFocusableChildNode is called.

GeorgiSTodorov:
>  Before these changes, you would have to tab twice to focus on a button or link trigger. See in the first GIF how a blue outline appears around the button, then the purple border appears around the button

I see, you manually added a `PrimaryButton` to the story.

PR #504, opened 2020-03-26
D2IQ-61192: TextBlock component

mperrotti:
## Screenshot
Screen Shot 2020-03-26 at 12 57 02 PM

These changes also include:
- a fix for the local design tokens build. Before, only a .ts file was being generated when running `npm run build:design-tokens`. Now, tsc runs after the .ts file is generated, and the .js file is generated.
- global default text styles on the document root


packages/shared/styles/global.ts
```diff
@@ -8,10 +8,22 @@ import {
 } from "../../design-tokens/build/js/designTokens";
 import { normalize } from "./normalize";
 
+// this is used to remove top and bottom margins for <p> and <h1-6> tags
+//
+// removing the margin directly in the <Text> component styles makes it impossible to use
+// child selectors to change the margins.
+// e.g.: using the Stack or TextBlock components.
+export const staticClass_resetTextMargin = "rmMargins";
```
brandonc: rmTextMargins?

mperrotti: I like this suggestion. I'll make that small update and use admin permissions to merge ;)

PR #505, opened 2020-03-27
D2IQ-64871: make Storybook easier to navigate

mperrotti:
## Testing
Look at Storybook's sidebar navigation and let me know if you find it easier to navigate to whatever story you're looking for.

## Trade-offs
The categories (e.g.: Page structure, Layout, Forms) are not ordered in any way. When we upgrade to Storybook 5.2 or higher, we should be able to use the [sortStories](https://storybook.js.org/docs/configurations/options-parameter/#sorting-stories) option to put the categories into some logical order.

## Screenshot
Screen Shot 2020-03-27 at 3 12 21 PM

### DIFF
--- a/packages/card/stories/Card.stories.tsx
+++ b/packages/card/stories/Card.stories.tsx
@@ -2,12 +2,12 @@ import * as React from "react";
 import { storiesOf } from "@storybook/react";
 import { withReadme } from "storybook-readme";
 import { withKnobs, select } from "@storybook/addon-knobs";
-import { Card } from "../index";
+import { ButtonCard, Card, LinkCard } from "../index";
 import { SpaceSize } from "../../shared/styles/styleUtils/modifiers/modifierUtils";
 
 const readme = require("../README.md");
 
-storiesOf("Card", module)
+storiesOf("Layout|Card", module)
   .addDecorator(withReadme([readme]))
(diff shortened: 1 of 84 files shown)


packages/tableView/stories/TableView.stories.tsx
```diff
@@ -6,7 +6,7 @@ import DemoTable from "./helpers/DemoTable";
 
 const readme = require("../README.md");
 
-storiesOf("TableView", module)
+storiesOf("Page structure|TableView", module)
```
mperrotti: I was undecided about how to categorize this component. I could see it living in any of these three categories:
- Page structure
- Layout
- Data listing

packages/card/stories/Card.stories.tsx
```diff
@@ -7,7 +7,7 @@ import { SpaceSize } from "../../shared/styles/styleUtils/modifiers/modifierUtil
 
 const readme = require("../README.md");
 
-storiesOf("Card", module)
+storiesOf("Layout|Card", module)
```
mperrotti: I'm not sure if this really belongs in the "Layout" category, but I wasn't sure where else to put it.

GeorgiSTodorov:
I think some of the items can be moved, but I also think it's good the way it is.

PR #511, opened 2020-04-20
Adds Popover component (needed for D2IQ-65804)

mperrotti:
Relates to [D2IQ-65804](https://jira.d2iq.com/browse/D2IQ-65804)

## Testing
Take a look at the Popover stories under the "Overlays" heading

## Screenshots
Screen Shot 2020-04-20 at 6 16 24 PM

## Checklist for PR creator


GeorgiSTodorov:
This might be intended, but I cannot copy the content in the menu.

GeorgiSTodorov:
Also a minor thing, but the dimensions here are 2 pixels bigger than they should be.
Screenshot from 2020-04-21 10-16-46

packages/nonMenuDropdown/NonMenuDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+interface NonMenuDropdownProps
+  extends Omit<DropdownableProps, "open" | "dropdown" | "onClose"> {
+  /**
+   * A unique identifier to help screenreaders or UI testing tools associate the dropdown content with it's trigger button. If an ID prop is not provided, an arbitrary ID will be generated.
+   */
+  id?: string;
+  /**
+   * Whether the dropdown starts open
+   */
+  initialIsOpen?: boolean;
```
pierrebeitz, line 20: how about `isInitiallyOpen`?

packages/nonMenuDropdown/NonMenuDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /** The maximum height of the dropdown overlay, represented by number of pixels. */
+  maxHeight?: number;
```
pierrebeitz, line 22: might there ever be the case that a `vh` could come in handy? in that case a string might also be a valid choice.

danielone:
agree with @pierrebeitz  on the component naming
would recommend a positive naming vs. negative ;) 

Popover makes sense as different facets could launch it. 
https://material-ui.com/components/popover/
Check out how material design's also is related to modal
NonMenu naming feels confusing

danielone:
"descriptive" aspect of the button naming in the Do's section feels too vague
Same button naming standards apply ( like verb + objectType, objectType, action depending on usage)
In this case, we will use [icon] Apps

mperrotti:
Thank you all for the thoughtful feedback. I completely agree about the name, and have updated it to Popover.

All of the other feedback should be addressed too.

GeorgiSTodorov:
> Also, here the content is hidden.

> Also a minor thing, but the dimensions here are 2 pixels bigger than they should be.

These two are still the same for me.

packages/nonMenuDropdown/NonMenuDropdown.tsx
mperrotti: I was trying to keep it consistent with `DropdownMenu`, which uses `initialIsOpen`

packages/nonMenuDropdown/NonMenuDropdown.tsx
mperrotti: There probably is a case, but I can't think of one right now. If we find a case for accepting strings with CSS units, we should do a follow-up PR that adds that functionality to all Popover-like components

PR #514, opened 2020-04-27
chore: add a story showing that disabling cols is possible

pierrebeitz:
we had a bug in disabling columns before and created an SSCCE before fixing it.
here's a more readable version of that.

### DIFF
--- a/packages/tablev2/Table.tsx
+++ b/packages/tablev2/Table.tsx
@@ -47,6 +47,9 @@ export type Column<A, B = string> = {
   textAlign?: React.CSSProperties["textAlign"];
 };
 
+// we internally use a different type than we expose in the API.
+type InternalColumn<A, B = string> = Column<A, B> & { width: string };
+
 type TableProps<A, B = string, C extends B = B> = {
   /**
    * Table columns
@@ -81,15 +84,15 @@ type State = {
   order: "asc" | "desc";
   columns: Array<{
     id: string;
-    width: string;
+    width: string | null;
   }>;
 };
 
-const getCellClassName = (
-  gridTemplateFragments: string[],
+const rowClassName = <A extends unknown>(
+  columns: Array<InternalColumn<A>>,
   stickyFirstCol?: boolean
 ) =>
-  cx(style.row(gridTemplateFragments), {
+  cx(style.row(columns.map(c => c.width)), {
     [style.rowWithStickyColumn]: stickyFirstCol
   });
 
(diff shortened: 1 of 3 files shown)


mperrotti:
@pierrebeitz - the columns are visually removed, but the `grid-template-columns` property stays the same. This means when I remove a column, all of the column widths remain the same.

pierrebeitz:
@mperrotti uh, that's right! 💡🤔

do we want/need to handle that though? if developers show/hide stuff, we could encourage them to use `fr`s. if users show/hide columns they resized before, i think they will be fine with resizing again. i say that as i believe ui-kit should not try to automagically compute widths when changing column-visibility. but maybe you want to solve this differently (or have done this before - it just appears complex to me to implement that "correctly").

pierrebeitz:
after some background process realized what you actually meant, i think i'll provide an update today.

pierrebeitz:
should be fixed: 
table

packages/tablev2/Table.tsx
```diff
@@ -231,30 +223,29 @@ export function Table<Entry, ColIDs extends string, Sort extends ColIDs>({
   const sort =
     col?.sorter?.(state.order, state.order === "asc" ? 1 : -1) ?? (() => 0);
   const sortedData = data.sort(sort);
+  const internalColumns: InternalColumn<Entry>[] = columns.map(c =>
+    Object.assign(c, { width: getWidth(c, state.columns) })
```
pierrebeitz, line 227: this is the trick. we enhance the `Column` type by a width for internal use. that makes surrounding stuff less hard.

packages/tablev2/Table.tsx
```diff
@@ -185,20 +185,17 @@ function HeaderRow<Entry>({
...
+  const className = cx(rowClassName(columns, stickyFirstCol), style.headerRow);
   return (
     <div role="row" className={className} key="headerRow">
       {columns.map(toHeaderCell)}
     </div>
   );
 }
 
+const getWidth = (col: Column<any>, state: State["columns"]) =>
+  state.find(s => s.id === col.id)?.width || col.initialWidth || "1fr";
```
pierrebeitz, line 197: `getWidth` is now in charge of computing allTheWidths and their defaults.

mperrotti:
So sorry for taking so long to come back to this! I missed your updates.
I'll do another review and try to get this merged today :)

pierrebeitz:
corrected the failed linting... 🤦

mperrotti:
@pierrebeitz - I think snapshots need to be updated

GeorgiSTodorov:
@pierrebeitz , I updated the snapshots, I hope you don't mind.

pierrebeitz:
@GeorgiSTodorov you know i'm always delighted when someone adds to my PRs! 💜  
Thanks a bunch and sorry for not updating those in the first place.

PR #515, opened 2020-05-18
D2IQ-68105: removes styles that affect layout

mperrotti:
These changes also include styles that show an outline around avatars so they are visible when they are on a background that is the same color

Closes D2IQ-68105

### Before:
Layout quirks:
Screen Shot 2020-05-18 at 6 20 53 PM
Screen Shot 2020-05-18 at 6 21 11 PM

On top of the same background color:
Screen Shot 2020-05-18 at 6 28 37 PM

### After:
Layout quirks:
Screen Shot 2020-05-18 at 6 23 20 PM
Screen Shot 2020-05-18 at 6 26 37 PM

On top of the same background color:
Screen Shot 2020-05-18 at 6 27 59 PM

## Checklist for PR creator

### DIFF
--- a/packages/avatar/style.ts
+++ b/packages/avatar/style.ts
@@ -1,16 +1,24 @@
 import { css } from "emotion";
 import { AvatarSizes } from "./components/Avatar";
-import {
-  themeBgSecondary,
-  greyLightLighten5
-} from "../design-tokens/build/js/designTokens";
+import { themeBgSecondary } from "../design-tokens/build/js/designTokens";
 
 export const avatarContainer = css`
   background-color: ${themeBgSecondary};
-  border: 1px solid ${greyLightLighten5};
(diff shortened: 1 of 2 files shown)


danielone:
@mperrotti  icons look much better. kiali seems like the icon is a bit too big as icon touches edges compared to grafana? 

really nice touch to also address the issue about icons having same background color as our grey panels. 
concerns here: 
- this adds a border to some icons vs. not others, unless all have this grey border now? don't see that and not sure it would look good. 
- our background grey color here and the background of the icon shouldn't match in the first place even with a border. I don't think that will look good. I think we should rethink our default color to be something in between white and the grey panel, if possible.
As you prefer to handle: these concerns  don't need to delay this PR but then let's handle these concerns in another ticket please which I am happy to file.

PR #518, opened 2020-06-02
D2IQ-66357: repositions Dropdownable on scroll

mperrotti:
Closes D2IQ-66357

## Testing
- Components that use Dropdownable should update their position on scroll. The Dropdownable overlay should be hidden when the trigger is scrolled out of view.

## Trade-offs
PopperJS has been added as a dependency. It's pretty small, but it has a lot of very useful features.

## Screenshots
RepositionDropdownableOnScroll

## Checklist for PR creator


packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -86,17 +88,26 @@ class Dropdownable extends React.Component<DropdownableProps, State> {
 
   componentDidMount() {
     resizeEventManager.add(this.setPositionFromCurrentProps);
+    window.addEventListener("scroll", this.hideDropdownable, true);
+    // windowEventManager.add("scroll", this.hideDropdownable, true);
```
GeorgiSTodorov, line 92: I think this was forgotten. I am removing it.

pierrebeitz:
i don't feel too good about this change. 

i think i have not seen this behaviour in any other library for popovers yet (at least for popovers that i opened with a click).
the other thing is that we now have a "partially controlled component" - which is not the most straightforward thing to work with from my perspective.

can we maybe solve the original issue differently? i think issues like there are the reason why e.g. bootstrap ships with [popper.js](https://popper.js.org/).

mperrotti:
@pierrebeitz, I think you're right about the "partially controlled component" being undesirable. I'll look into alternative ways to solve this problem. I'm very open to any recommendations.

I've seen this "close the dropdown on scroll" behavior before, and I'd prefer we keep it. It helps us avoid accessing DOM node measurements, which causes a repaint.

mperrotti:
I haven't come up with anything I'm happy with for hiding on scroll. 😅 
I'm considering using Popper.js (probably via react-popper) to power Dropdownable and keep it performant.

mperrotti:
@pierrebeitz - it's referring to the "Expand/Collapse" button on the left

PR #523, opened 2020-06-22
D2IQ-67847: add pointer caret to Popover

mperrotti:
Relates to D2IQ-67847

## Testing

- Confirm that the Popover component has a caret pointing at the target (just like the Tooltip component)
- Ensure the style refactoring didn't break any Tooltip styles

## Screenshots
Screen Shot 2020-06-22 at 2 00 21 PM

## Checklist for PR creator


mperrotti:
@natmegs - this is the same behavior we have with the tooltip arrow, and I think we could make the caret a little "smarter" to avoid these situations.

Thanks for testing with a bunch of text to catch this!

mperrotti:
@natmegs - I can only reproduce that issue when the popover ends up being as wide as the viewport.

I think this is enough of an edge case to ignore, but I have a working solution that will position the caret at the horizontal center of the trigger. The only downside is that it adds 2 more props to `PopoverBox` and `TooltipContent`.

I'll push the commit if you or any other reviewers think it's worth it 👍

PR #527, opened 2020-07-10
UIKitProvider component

brandonc:
UIKitProvider is a new root level wrapper for ui-kit that will serve as a global context provider. The first use case is linkComponent, which allows consumers to specify a delegate for all links.

Under the hood, I added a component UnstyledLink that should be used instead of an anchor tag everywhere which will consume the underlying LinkComponentContext that is provided by UIKitProvider.

## Testing

Existing buttons should not be affected. The new story delegates to a link that overrides children and onClick handler.

## Trade-offs

This supercedes UIKitThemeProvider and will serve as a place to hang other global settings.

## Dependencies

n/a

## Screenshots

Screenshot_2020-07-10 Utils UIKitProvider - link delegate ⋅ Storybook

## Checklist for PR creator


packages/link/components/ResetLink.tsx
```diff
@@ -1,27 +1,17 @@
...
-    {...other}
-  >
+  <UnstyledLink className={cx(linkReset, className)} {...other}>
     {children}
-  </a>
+  </UnstyledLink>
 );
 
-export default ResetLink;
+export { ResetLink as default };
```
mperrotti, line 17: I haven't done default exports like this before. Are there any benefits? Or just a matter of personal preference?

Either way, I'm not opposed to it.

packages/utilities/interactionMode.ts
```diff
@@ -3,22 +3,31 @@
     https://blueprintjs.com/docs/#core/accessibility.focus-management
 */
 
+import classNames from "classnames";
```
mperrotti, line 6: I like these changes 👍

PR #528, opened 2020-07-13
D2IQ-69415: Accordion components

mperrotti:
Closes D2IQ-69415

## Testing
Play with the Storybook examples and try to break things.

## Trade-offs
Focus cannot yet be moved using your keyboard's arrow keys. This functionality is marked as "optional" in [the spec](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion), but might be a nice detail.

If any reviewers think we should add this functionality, I will create a follow-up ticket.

## Screenshots
Screen Shot 2020-07-15 at 5 20 43 PM

## Checklist for PR creator


danielone:
Feedback
1- in the screenshot, i get that for dispatch "console" we put the expanded content outside bar, but... hmmm... perhaps we shouldn't and have the border wrap around its contents so context is clearer and we can still have a horizontal line border below the title and controls? If this component would be used with other content/case, like some traditional fields or config map, it would look very strange for the content to sit outside. 

2- expand/collapse via keyboard would be a great addition for our developer/power users who work a lot with the keyboard anyway and be a very nice detail as you mention

3- Does the panel give layout for actions at the right automatically? i think this would be good so devs would not have to worry about. 

4- I think we have expandable/collapsible headings too in Ksphere? I know DC/OS for sure uses. We'll need to distinguish when to use which in the guidelines but ideally its an easy switch between both for the dev and the props are similar

mperrotti:
Thanks for taking a look @danielone.

> 1. ...have the border wrap around its contents so context is clearer and we can still have a horizontal line border below the title and controls...

Sure, we can change the design and do this. Just keep in mind that this would be a global change.

---

> 2. ...expand/collapse via keyboard would be a great addition...

You can expand and collapse using keyboard navigation. You just have to use the tab key to change which heading has focus, not up and down arrow keys. Everything I build in ui-kit is keyboard accessible :)

---

> 3. Does the panel give layout for actions at the right automatically?

No, because we don't have a design pattern for this yet.

---

4- I think we have expandable/collapsible headings too in Ksphere? I know DC/OS for sure uses. We'll need to distinguish when to use which in the guidelines but ideally its an easy switch between both for the dev and the props are similar

I think you're referring to the [Expandable](https://dcos-labs.github.io/ui-kit/?path=/story/actions-expandable--default) component. I don't think we have them in any Ksphere products yet, but we do in DC/OS. 100% agreed that we should document when to use an `Expandable` vs when to use `Accordion`.

danielone:
> Thanks for taking a look @danielone.
> 
> > 1. ...have the border wrap around its contents so context is clearer and we can still have a horizontal line border below the title and controls...
> 
> Sure, we can change the design and do this. Just keep in mind that this would be a global change.

Understood. This was the factor that drove me to raise for us to rethink the design for the dispatch "console" in such a component so that way dispatch "console" view can behave consistently with such a design change. Nan and I will share with you the latest mocks on this so we can look at the differences.

> 
> > 1. ...expand/collapse via keyboard would be a great addition...
> 
> You can expand and collapse using keyboard navigation. You just have to use the tab key to change which heading has focus, not up and down arrow keys. Everything I build in ui-kit is keyboard accessible :)

Ahh I see and great about the tab support!!. I assume tab would not skip over the contents of the header but go inside if it is open?  Hmmm.. would Tab+ up/down make sense vs. just the arrow key? Is there a keyboard standard for jumping to header to header / region to region? This way can apply same to the headers too perhaps and even cards/different sections? 

> 
> > 1. Does the panel give layout for actions at the right automatically?
> 
> No, because we don't have a design pattern for this yet.

SG. I guess if we use this elsewhere we can add this in at the right time you see fit for D2DS but Nan and I are starting to see a potential pattern here if you agree so as well. 

> 
> 4- I think we have expandable/collapsible headings too in Ksphere? I know DC/OS for sure uses. We'll need to distinguish when to use which in the guidelines but ideally its an easy switch between both for the dev and the props are similar
> 
> I think you're referring to the [Expandable](https://dcos-labs.github.io/ui-kit/?path=/story/actions-expandable--default) component. I don't think we have them in any Ksphere products yet, but we do in DC/OS. 100% agreed that we should document when to use an `Expandable` vs when to use `Accordion`.

SG. I'll look through Ksphere but i think the advanced setting section got descoped from Create Cluster. How come the caret is on the right side of the region please? I recall in DC/OS we have the caret on the left (similar to this component)

packages/accordion/README.md
```diff
@@ -0,0 +1,21 @@
+# Accordion
+
+The `Accordion` component is a collection of headings that can be clicked to hide and show related content.
+
+The `Accordion` is used to save vertical space when when rendering a lot of content, while still letting the user scan headings to help them find the specific content they're looking for.
```
TattdCodeMonkey, line 5: double `when` here

packages/accordion/components/Accordion.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface AccordionProps {
+  /**
+   * Whether multiple accordion items can be open at once
+   */
+  allowMultipleExpanded?: boolean;
+  /**
+   * human-readable selector used for writing tests
+   */
+  ["data-cy"]?: string;
```
TattdCodeMonkey, line 13: 👏  This is much needed and great to have. Thanks for explicitly supporting `data-cy` throughout the entire component tree.

PR #544, opened 2020-08-26
feat(Textarea): allow for `tooltipContent`

pierrebeitz:
`TextInput` nicely provides us with the ability to specify a `tooltipContent`
and then displays an appropriate tooltip right besides the label above the
input.

`Textarea` now can do this too! :tada:

i first copied over all the parts but then decided to start a utility. 
so we extracted `getLabelContent` and `getTooltipContent` from `TextInput` and refactored those a bit. 
by now i feel that a next step could be to make that an `input`-utility and put even more
shared code there. but that's something for another PR. :)

also somehow some changes regarding our data-cy-classes emerged as i found those
way too noisy:

```diff
-    const dataCy = [
-      "textInput",
-      ...(appearance && appearance !== InputAppearance.Standard
-        ? [`textInput.${appearance}`]
-        : [])
-    ].join(" ");
+    const dataCy = `textInput textInput.${appearance}`;
```

i can of course revert that change, if there's something known to rely on
`textInput.standard` being absent.

## Screenshots

Forms | Textarea - default ⋅ Storybook 2020-08-26 16-51-34

### DIFF
--- a/packages/utilities/label.tsx
+++ b/packages/utilities/label.tsx
@@ -1,0 +1,56 @@
+import * as React from "react";
+import { InputAppearance } from "../shared/types/inputAppearance";
+import { cx } from "emotion";
+import {
+  tintText,
+  visuallyHidden,
+  flex,
+  margin
+} from "../shared/styles/styleUtils";
+import { getLabelStyle } from "../shared/styles/formStyles";
+import {
+  themeError,
+  greyLightDarken2,
+  iconSizeXs
+} from "../design-tokens/build/js/designTokens";
+import Tooltip from "../tooltip/components/Tooltip";
+import { Icon } from "../icon";
+import { SystemIcons as SI } from "../icons/dist/system-icons-enum";
+
+const trigger = (
+  <Icon color={greyLightDarken2} shape={SI.CircleQuestion} size={iconSizeXs} />
+);
+const reqStar = <span className={cx(tintText(themeError))}> *</span>;
+
+export const renderLabel: React.FC<{
+  appearance?: string;
+  hidden?: boolean;
+  id: string;
+  label?: React.ReactNode;
+  required?: boolean;
+  tooltipContent?: React.ReactNode;
+}> = ({ appearance, hidden, id, label, required, tooltipContent }) => {
+  const hasError = appearance === InputAppearance.Error;
+  const labelClassName = hidden ? cx(visuallyHidden) : getLabelStyle(hasError);
+  const labelNode = (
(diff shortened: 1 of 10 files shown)


packages/utilities/label.tsx
```diff
@@ -0,0 +1,68 @@
...
+}) => {
+  const hasError = appearance === InputAppearance.Error;
+  const labelClassName = hidden ? cx(visuallyHidden) : getLabelStyle(hasError);
+  const labelNode = (
+    <label className={labelClassName} htmlFor={id} data-cy="textInput-label">
+      {label}
+      {required ? <span className={cx(tintText(themeError))}> *</span> : null}
+    </label>
+  );
+  if (!tooltipContent) {
```
pierrebeitz, line 49: we could also get rid of this conditional by always wrapping the result in `<div className={flex({ align: "center" })}>`. that apparently works but creates a lot of noise in the snapshots.

```diff
-  const labelNode = (
-    <label className={labelClassName} htmlFor={id} data-cy="textInput-label">
-      {label}
-      {required ? <span className={cx(tintText(themeError))}> *</span> : null}
-    </label>
-  );
-  if (!tooltipContent) {
-    return labelNode;
-  }
   return (
     <div className={flex({ align: "center" })}>
-      {labelNode}
+      <label className={labelClassName} htmlFor={id} data-cy="textInput-label">
+        {label}
+        {required ? <span className={cx(tintText(themeError))}> *</span> : null}
+      </label>
       {tooltipContent ? (
```

packages/utilities/label.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return labelNode;
+  }
+  return (
+    <div className={flex({ align: "center" })}>
```
pierrebeitz, line 53: i wonder why we need `align: "center"` in this use case it does not seem to be necessary.

mperrotti:
> by now i feel that a next step could be to make that an input-utility and put even more
shared code there. but that's something for another PR. :)

I agree! Could you create a Jira ticket for this?

mperrotti:
> also somehow some changes regarding our data-cy-classes emerged as i found those
way too noisy

There's nothing that would break in Kommander. However another app could have written a selector like this: `"[data-cy='textInput']"`. This would fail because it would need to be changed to something like `"[data-cy*='textInput']"`

packages/utilities/label.tsx
mperrotti: I don't remember why this was added, but removing it doesn't seem to have any affect. I'm fine with removing it :)

PR #551, opened 2020-09-04
D2IQ-68837: creates a new row in a fieldlist when tabbing off last input

mperrotti:
Closes D2IQ-68837

## Testing

- Go to any FieldList story
- Tab through the last row and confirm a new row appears when you tab from the last input to the last close button
- Tab through the new row _without typing values in the inputs_ and confirm that a new row is not added when you tab from the last input to the last close button

## Trade-offs

I'm using `onKeyUp` and checking for the "Tab" key to add a new row. I tried a few approaches using onBlur and onFocus, but they felt even more hacky

## Screenshots

https://share.getcloudapp.com/2Nuy9ZLX

## Checklist for PR creator

### DIFF
--- a/packages/formStructure/components/FieldList.tsx
+++ b/packages/formStructure/components/FieldList.tsx
@@ -1,10 +1,10 @@
 import * as React from "react";
+import { cx } from "emotion";
 import {
   FieldListColumnProps,
   FieldListColumnSeparator,
   FieldListColumnWidthProps
 } from "./FieldListColumn";
-import { cx } from "emotion";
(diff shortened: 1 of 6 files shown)


DanielMSchmidt (approved):

TattdCodeMonkey (approved):
Code changes LGTM and tested as expected.

I will say I don't particularly like the behavior though. IMO you should tab to the "Add" button and press enter and then focus the new row. which I think would of worked prior to this change. But that is more of a comment on the ticket itself.

PR #555, opened 2020-09-29
D2IQ-70897: adds StatelessAccordion component

mperrotti:
Relates to D2IQ-70897

## Testing

All the Accordion stories should still work as expected.
The StatelessAccordion story should work.

## Trade-offs

There's some shared code between Accordion and StatelessAccordion that could probably be DRY'd out.

## Screenshots

https://share.getcloudapp.com/qGuv1Dql

## Checklist for PR creator

### DIFF
--- a/packages/accordion/components/StatelessAccordion.tsx
+++ b/packages/accordion/components/StatelessAccordion.tsx
@@ -17,12 +17,6 @@ const Accordion: React.FC<AccordionProps> = ({
   expandedItems = [],
   onChange
 }) => {
-  //   React.useEffect(() => {
-  //     if (onChange) {
-  //       onChange(expandedItems);
-  //     }
-  //   }, [expandedItems]);
-  //   console.log("expandedItems", expandedItems);
   return (
     <AccordionProvider
       allowMultipleExpanded={allowMultipleExpanded}
--- a/packages/accordion/components/StatelessAccordion.tsx
+++ b/packages/accordion/components/StatelessAccordion.tsx
@@ -7,20 +7,26 @@ interface AccordionProps extends AccordionBaseProps {
   /**
    * An array of open accordion panel IDs
    */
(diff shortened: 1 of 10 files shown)


packages/accordion/components/StatelessAccordion.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const Accordion: React.FC<AccordionProps> = ({
+  allowMultipleExpanded,
+  "data-cy": dataCy,
+  children,
+  expandedItems = [],
+  onChange
+}) => {
+  //   React.useEffect(() => {
```
DanielMSchmidt, line 20: Here is a comment left?

packages/accordion/components/StatelessAccordion.tsx
```diff
@@ -0,0 +1,42 @@
...
+    </AccordionProvider>
+  );
+};
+
+Accordion.defaultProps = {
+  allowMultipleExpanded: false,
+  "data-cy": "accordion"
+};
+
+export default Accordion;
```
DanielMSchmidt, line 42: Just curious, why did you decide to go with a new component, vs. being able to use the accordion in a controlled and uncontrolled way (like one would use inputs)

packages/accordion/components/StatelessAccordion.tsx
mperrotti: 🤦 will fix

packages/accordion/components/StatelessAccordion.tsx
mperrotti: Because eliminating the prop `initialExpandedItems` would be a breaking change, and I didn't want to add new prop just to allow the Accordion to be used in a controlled way.

natmegs (approved):

brandonc (approved):

PR #561, opened 2020-10-19
D2IQ-69910: eliminates indentation of wrapping FlexItems when Flex has gutterSize set

mperrotti:
Closes D2IQ-69910

## Testing

Confirm there is no indentation of rows when `<Flex>` has `wrap="wrap"` and `gutterSize` props set.

## Screenshots
Before:
Screen Shot 2020-10-19 at 9 49 55 AM

After:
Screen Shot 2020-10-19 at 10 00 44 AM

## Checklist for PR creator

### DIFF
--- a/packages/shared/styles/styleUtils/layout/flexbox.ts
+++ b/packages/shared/styles/styleUtils/layout/flexbox.ts
@@ -132,29 +132,26 @@ export const applyFlexItemGutters = (direction, gutterSize) => {
     return;
   }
 
-  const columnPaddingTop = getResponsiveColumnStyles(
-    direction,
-    gutterSize,
-    "none"
-  );
-  const columnPaddingLeft = getResponsiveColumnStyles(
-    direction,
-    "none",
-    gutterSize
-  );
-  const paddingTop =
-    typeof columnPaddingTop === "object"
-      ? getGutterPaddingValues(columnPaddingTop, gutterSize)
-      : columnPaddingTop;
-  const paddingLeft =
-    typeof columnPaddingLeft === "object"
-      ? getGutterPaddingValues(columnPaddingLeft, gutterSize)
-      : columnPaddingLeft;
+  const rowGuttter = getResponsiveColumnStyles(direction, gutterSize, "none");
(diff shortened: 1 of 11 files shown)


natmegs (commented):
I saw this change in row spacing as I changed my screen width. Is that expected?
Screen Shot 2020-10-20 at 4 26 49 PM

mperrotti:
@natmegs - that was a Storybook mistake. Fixed.

packages/shared/styles/styleUtils/layout/flexbox.ts
```diff
@@ -132,29 +132,26 @@ export const applyFlexItemGutters = (direction, gutterSize) => {
...
-    > *:not(:first-child) {
-      ${getResponsiveSpacingStyle("padding-left", paddingLeft)};
-      ${getResponsiveSpacingStyle("padding-top", paddingTop)};
+    ${getResponsiveSpacingStyle("column-gap", columnGap)};
+    ${getResponsiveSpacingStyle("row-gap", rowGap)};
+
+    @supports not (column-gap: 1px) and (row-gap: 1px) {
+      > *:not(:last-child) {
+        ${getResponsiveSpacingStyle("padding-right", rowGap)};
+        ${getResponsiveSpacingStyle("padding-bottom", columnGap)};
```
natmegs, line 153: This seems switched to me (would have thought padding-right should be columnGap?) but not sure how to test. Likely I am interpreting incorrectly 😄

natmegs (approved):

packages/shared/styles/styleUtils/layout/flexbox.ts
mperrotti: Yea, I see how that wording is confusing.

[row-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap) refers to the space **between rows**
[column-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap) refers to the space **between columns**

natmegs (approved):

PR #563, opened 2020-10-19
Add `tag` prop to Flex/FlexItem + update markup in ProgressBar

mperrotti:
These changes will help eliminate some React errors


mperrotti:
I don't think this needs to be reviewed. Merging...

PR #574, opened 2020-11-25
D2IQ-65176: generate key for breadcrumb actions if not provided

natmegs:
Closes D2IQ-65176

## Testing

You should be able to remove keys on `breadcrumbElements` and `actions` (or just replace breadcrumbElements with array of strings) passed to PageHeader components in stories without getting React errors

## Trade-offs

`breadcrumbElements` already have keys generated here https://github.com/dcos-labs/ui-kit/blob/18ebc935b915e1354c7d8626335b26bffe09c44b/packages/breadcrumb/components/Breadcrumb.tsx#L35


DanielMSchmidt (approved):

packages/pageheader/components/PageHeader.tsx
```diff
@@ -44,11 +44,11 @@ class PageHeader extends React.PureComponent<PageHeaderProps, {}> {
               className={cx(listReset, flex({ align: "center" }))}
               data-cy="pageHeader-actions"
             >
-              {actions.map(action => {
+              {actions.map((action, i) => {
                 return (
                   <li
+                    key={action.key ? `${action.key}` : `action-${i}`}
```
pierrebeitz, line 50: we won't suffer from the problems described here, right? https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

pierrebeitz: ☝️ @natmegs @DanielMSchmidt?

DanielMSchmidt: I agree but also disagree. 
A) Normally they don't change
B) This is just a fallback, if we run into problems we can still go the better way and set them

pierrebeitz: A) true dat!
B) i see the point of making it easy to use this component. especially when we don't want to care about making up ids. just want to leave that "if we run into problems" does not resonate with me here, as this change effectively 1) makes it easy to break things 2) makes it easy to copy code that has non-obvious assumptions. 

i'm not saying that i want to revert it. just saying that i don't like components that don't have my back 😊

DanielMSchmidt: Yeah, I guess you are right 🤔 It's not really explicit that it can not change, it's just what we currently do... @natmegs what is you opinion?

PR #575, opened 2020-11-26
fix: add second generic to FieldListColumnProps definition

natmegs:
The FieldList and FieldListColumn components allow consumers to define any type of onChange handler for their field columns, and even allow configuring the `value` type via generic. Currently we are not able to fully take advantage of this `value` flexibility because the render props `onChange` handler assumes that the event will be an InputElement form event. To allow use with different types of inputs (like selects) and different types of form controls that don't use form events at all (like TextInputWithBadges), a second generic is added to the column definition.

## Testing

Should work as usual

### DIFF
--- a/packages/formStructure/components/FieldListColumn.tsx
+++ b/packages/formStructure/components/FieldListColumn.tsx
@@ -5,13 +5,13 @@ type DefaultProps = Pick<TextInputProps, "key" | "id" | "inputLabel">;
 
 type Value<T> = T;
 
-interface RenderProps<T> {
+interface RenderProps<T, E> {
   /** Whether field is disabled */
   disabled?: boolean;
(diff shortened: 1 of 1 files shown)


packages/formStructure/components/FieldListColumn.tsx
```diff
@@ -5,13 +5,13 @@ type DefaultProps = Pick<TextInputProps, "key" | "id" | "inputLabel">;
 
 type Value<T> = T;
 
-interface RenderProps<T> {
+interface RenderProps<T, E> {
```
pierrebeitz, line 8: should we add `React.FormEvent<HTMLInputElement>` here to make the transition easier?

brandonc (dismissed):

PR #579, opened 2021-01-22
D2IQ-73430 feat: slidetoggle

TattdCodeMonkey:
Added a new toggle component `SlideToggle` this is basically a checkbox
styled to be a slider with a check when checked.

I initially tried to implement this extending the ToggleComponent like
the CheckboxInput and RadioInputs do, but the styles were different
enough that I needed to make this its own component.

- Closes D2IQ-73430

## Testing

Open story and try it out.

## Screenshots

image

## Checklist for PR creator

### DIFF
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -134,6 +134,7 @@ export { UIKitThemeProvider } from "./themes";
 export { UIKitProvider } from "./uiKitProvider";
 export { CodeSnippet } from "./codesnippet";
 export { ClickToCopyButton } from "./clicktocopybutton";
+export { SlideToggle } from "./slideToggle";
 
 // Style utility components
 export {
--- a/packages/slideToggle/README.md
+++ b/packages/slideToggle/README.md
@@ -1,0 +1,3 @@
+# SlideToggle
+
+SlideToggles are used to toggle something on or off.
--- a/packages/slideToggle/components/SlideToggle.tsx
+++ b/packages/slideToggle/components/SlideToggle.tsx
@@ -1,0 +1,201 @@
+import * as React from "react";
+import { cx } from "emotion";
+import {
+  flex,
+  flexItem,
+  visuallyHidden,
+  display,
+  tintContent,
+  tintContentSecondary,
+  vAlignChildren
+} from "../../shared/styles/styleUtils";
+
+import {
+  themeError,
+  themeSuccess,
+  themeBgPrimary
+} from "../../design-tokens/build/js/designTokens";
+import { InputAppearance } from "../../shared/types/inputAppearance";
+import FormFieldWrapper from "../../shared/components/FormFieldWrapper";
+import { SystemIcons } from "../../icons/dist/system-icons-enum";
+import { Icon } from "../../icon";
+import {
+  checkContainer,
+  toggle,
+  toggleRound,
+  toggleContainer,
+  toggleInputApperances,
+  toggleInputLabel,
+  toggleInputFeedbackText
+} from "../style";
+
+export interface SlideToggleProps extends React.HTMLProps<HTMLInputElement> {
+  /**
+   * Sets the current appearance of the input component. This defaults to InputAppearance.Standard, but supports `InputAppearance.Error` & `InputAppearance.Success` appearances as well.
+   */
+  appearance?: InputAppearance;
+  /**
+   * Whether or not the input is checked
+   */
+  checked?: boolean;
+  /**
+   * Unique identifier used for the input element
+   */
+  id: string;
+  /**
+   * The text or node content that appears next to the input
+   */
+  inputLabel: React.ReactNode | string;
+  /**
+   * Defaults to `true`, but can be set to `false` to visibly hide the content passed to `inputLabel`. The `inputLabel` should still be set even when hidden for accessibility support.
+   */
+  showInputLabel?: boolean;
+  /**
+   * The value being toggled
+   */
+  value?: string;
+  /**
+   * How the text content vertically aligns with the input
+   */
+  vertAlign?: "center" | "top";
+  /**
+   * hintContent is text or a ReactNode that is displayed directly under the input with additional information about the expected input.
+   */
+  hintContent?: React.ReactNode;
+  /**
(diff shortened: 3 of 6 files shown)


packages/sliderInput/style.ts
```diff
@@ -0,0 +1,181 @@
...
+  themeBgDisabled,
+  themeBgPrimary,
+  themeBrandPrimary,
+  themeError,
+  themeSuccess,
+  themeTextColorDisabled,
+  white
+} from "../design-tokens/build/js/designTokens";
+
+const sliderInputHeight = 14;
```
natmegs, line 15: Toggle input height is 16, wonder if they should be the same?

packages/sliderInput/style.ts
```diff
@@ -0,0 +1,181 @@
...
+  border: solid;
+  border-width: ${sliderRoundBorder}px;
+  border-color: ${themeBorder};
+  bottom: 0;
+  cursor: pointer;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  transition: 0.4s;
```
natmegs, line 145: This is totally subjective so no action needed, but this feels a bit long to me? It looks like we commonly use 150-250ms transitions in other places

natmegs (commented):
This looks so good!!

packages/sliderInput/style.ts
TattdCodeMonkey: The size was taken directly from the invision design.

packages/sliderInput/style.ts
TattdCodeMonkey: This was taken directly from the CodePen I modeled this from so will change to be more inline with other components.

natmegs (dismissed):

packages/sliderInput/style.ts
danielone: @natmegs good eye

danielone: I can't review the interactive aspects but thanks @natmegs for reviewing such. trust your eye

packages/sliderInput/style.ts
danielone: @TattdCodeMonkey yeah best to keep same as mockup as control would be too big at same height but its box area should take up same space as other toggles so everything aligns when used together in a form layout.

TattdCodeMonkey: @danielone I don't think keeping it the same size as other toggles is possible. The checkbox is a square 16x16 and the radio button is a circle. This is an oval that is 14x28 and the width is needed to move the slider and display the check.

packages/sliderInput/README.md
```diff
@@ -0,0 +1,3 @@
+# SliderInput
```
pierrebeitz, line 1: 🔴 i'd totally vote against naming this a slider-input: https://duckduckgo.com/?q=slider+input&va=b&t=hc&iax=images&ia=images

how about "slide toggle" or such?

packages/sliderInput/components/SliderInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+    onBlur,
+    onFocus,
+    ...other
+  } = props;
+  const inputType = "Slider";
+  const inputDataCy = [
+    `${inputType}Input-input`,
+    ...(checked ? [`${inputType}Input-input.checked`] : []),
+    ...(appearance && appearance !== InputAppearance.Standard
+      ? [`${inputType}Input-input.${appearance}`]
```
pierrebeitz, line 94: i think at some point in time we decided not to make as big of a story out of those helper-classes and just went with not hiding the `Standard`-case as to reduce the noise.  
that's optional of course.

packages/sliderInput/components/SliderInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+      ? [`${inputType}Input.${appearance}`]
+      : [])
+  ].join(" ");
+  const [hasFocus, setHasFocus] = React.useState<boolean>(false);
+  const handleFocus = e => {
+    setHasFocus(true);
+
+    if (onFocus) {
+      onFocus(e);
+    }
```
pierrebeitz, line 110: WDYT about just `onFocus?.(e)`?

(totally optional)

packages/sliderInput/components/SliderInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+    <FormFieldWrapper id={id} errors={errors} hintContent={hintContent}>
+      {({ getValidationErrors, isValid, describedByIds, getHintContent }) => (
+        <div className={vAlignChildren} data-cy={parentDataCy}>
+          <label
+            className={cx(
+              flex({
+                align: vertAlign === "top" ? "flex-start" : "center"
+              }),
+              display("inline-flex"),
+              bruteForceKillLabelMargin // to override CNVS bottom margin on <label>
```
pierrebeitz, line 131: we still use CNVS? 😒

packages/sliderInput/style.ts
```diff
@@ -0,0 +1,181 @@
...
+  border: solid;
+  border-width: ${sliderRoundBorder}px;
+  border-color: ${themeBorder};
+  bottom: 0;
+  cursor: pointer;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  transition: 250ms;
```
pierrebeitz, line 145: nice. haven't seen that `position: absolute; tlbr: 0;` trick in a while. :)

pierrebeitz (changes requested):
works really well!

i have a concern regarding the naming though as from the components title i expected a completely different functionality :)

packages/sliderInput/components/SliderInput.tsx
TattdCodeMonkey: I'm actually not sure. this was very much copy 🍝  from `ToggleComponent`

TattdCodeMonkey: I removed this and didn't notice any changes, so I left it out for now.

pierrebeitz (dismissed):

TattdCodeMonkey:
rebased and force pushed to allow merging

PR #586, opened 2021-02-24
chore: remove deprecated table

pierrebeitz:
BREAKING CHANGE

we remove the Table_Deprecated component - along with react-virtualized which
seems to wheight in at around 30% of ui-kit's total size.

## Checklist for PR creator


package.json
```diff
@@ -31,7 +31,7 @@
     "setup:git-hooks": "./scripts/pre-install",
     "start": "NODE_ENV=development npm run storybook --root ./ ",
     "storybook": "start-storybook -p 6006",
-    "test": "jest",
+    "test": "npx jest",
```
pierrebeitz, line 34: needed to change this as else my pre-commit hook said 

```
sh: jest: command not found
```

maybe something with newer npm-versions? 🤷

TattdCodeMonkey (changes requested):
Changes look fine, but I think the commit message needs to be updated.

1. `chore` commits do not trigger a release so this should be made as a `feat` which is not intuitive, but is what it is.
2. the Body of the message probably should be:
```
BREAKING CHANGE: we remove the Table_Deprecated component - along with react-virtualized which
seems to wheight in at around 30% of ui-kit's total size.
```

I made a message similar to what you have currently in `kommander-ui` and semantic-release did not bump the major version. IDK why but `BREAKING CHANGE\n\n` was not handle the same as `BREAKING CHANGE: <message>`

pierrebeitz:
thanks a bunch @TattdCodeMonkey! updated the PR.

DanielMSchmidt (dismissed):

natmegs (dismissed):

TattdCodeMonkey (approved):

PR #602, opened 2021-07-21
feat: title attr on truncated Text

vacas5:
When a user sends the wrap prop with the arg of truncate, in the case
that the child is a string, output the child as the title HTML attribute
so that the user and screen readers will get the string in its entirety.

Closes D2IQ-77202

## Checklist for PR creator

### DIFF
--- a/packages/styleUtils/typography/components/Text.tsx
+++ b/packages/styleUtils/typography/components/Text.tsx
@@ -34,6 +34,12 @@ const Text = (props: TextProps) => {
     "data-cy": dataCy
   } = props;
   const TextTag = tag || defaultTag;
+  let title: string | undefined;
+
+  // if truncated, show title attr with complete text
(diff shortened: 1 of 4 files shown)


vacas5:
I am not so sure of the accessibility of this because it will force screen readers to read it twice, but I still think its the right solution.

packages/styleUtils/typography/components/Text.tsx
```diff
@@ -34,6 +34,12 @@ const Text = (props: TextProps) => {
     "data-cy": dataCy
   } = props;
   const TextTag = tag || defaultTag;
+  let title: string = "";
```
TattdCodeMonkey, line 37: I think this should default to `undefined` see the failing snapshots.

TattdCodeMonkey (changes requested):

PR #603, opened 2021-07-29
feat: add new airgapped icons

clintonmedbery:
Closes D2IQ-77204

## Testing

Go to Product Icons and see that they exist

## Screenshots

image

## Checklist for PR creator


packages/icons/src/icon-product/air-gapped.svg
```diff
@@ -0,0 +1,15 @@
...
+        <g id="icon/system/lock" transform="translate(7.000000, 6.000000)">
+            <mask id="mask-2" fill="white">
+                <use xlink:href="#path-1"></use>
+            </mask>
+            <use id="shape-primary" fill="#0C4C91" xlink:href="#path-1"></use>
+        </g>
+    </g>
+</svg>
```
vacas5, line 15: consider changing your IDE settings to automatically add EOL

vacas5 (commented):
Not sure I love the naming of this icon but it seems in line with the rest. Generally I prefer naming that describes what the icon is rather than how it's used. Material UI does it the way we have it and it drives me bonkers because it is so hard to find what you're looking for

packages/icons/src/icon-product/air-gapped-inverse.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vacas5, line 1: is this necessary? not seeing it in the other SVG files

packages/icons/src/icon-product/air-gapped-inverse.svg
```diff
@@ -0,0 +1,17 @@
...
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
vacas5, line 2: looks like most of the other icons use `24` rather than 16

packages/icons/src/icon-product/air-gapped-inverse.svg
```diff
@@ -0,0 +1,17 @@
...
+    <title>icon/system/air-gapped</title>
```
vacas5, line 3: not seeing any other svgs with title. Can remove

packages/icons/src/icon-product/air-gapped.svg
```diff
@@ -0,0 +1,15 @@
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
vacas5, line 1: see comments above

packages/icons/src/icon-product/air-gapped-inverse.svg
```diff
@@ -0,0 +1,17 @@
...
+        <g id="icon/system/lock" transform="translate(7.000000, 6.000000)">
+            <mask id="mask-2" fill="white">
```
vacas5, line 11: wondering about using a fill here. What happens if it's used on a gray background?

packages/icons/src/icon-product/air-gapped.svg
```diff
@@ -0,0 +1,15 @@
...
+        <g id="icon/system/lock" transform="translate(7.000000, 6.000000)">
+            <mask id="mask-2" fill="white">
```
vacas5, line 9: may need to set the fill dynamically

vacas5 (changes requested):

packages/icons/src/icon-product/air-gapped-inverse.svg
clintonmedbery: With the exception of removing the fill that matched our background color, I basically just used what design gave me but I can adjust.

clintonmedbery: Removed.

packages/icons/src/icon-product/air-gapped-inverse.svg
clintonmedbery: I think that is why we have two. One with colors and one that is white. TBH when I got them one was blue and the other was dark purple but they didn't match our color scheme so I made one white.

vacas5: yeah you are definitely following the existing convention here, so I'll table my concern

clintonmedbery:
Updated icon name but left the fill to fit in with other product icons. Also left the size since others had 16x16

packages/icons/src/icon-product/air-gapped.svg
mstruebing: Or even better: set up your IDE to respect the `.editorconfig`: https://github.com/dcos-labs/ui-kit/blob/master/.editorconfig

 https://editorconfig.org/

PR #606, opened 2021-08-09
feat: new promo banner background gradient

vacas5:
Introducing a new oceanBlue version of the background gradient for the
PromoBanner, which has a specific implementation with three stops.

I considered an alternative of allowing this component to take a custom background color.

Closes D2IQ-77630

## Dependencies

Have asked design to give me colors that are already in our design system rather than these three custom colors, so I will wait on them to merge this in, but the implementation can be reviewed.

## Screenshots

Screen Shot 2021-08-09 at 5 05 14 PM

## Checklist for PR creator

### DIFF
--- a/packages/promo/style.ts
+++ b/packages/promo/style.ts
@@ -1,7 +1,9 @@
 import { css } from "emotion";
 import { GradientColors, GradientStyle } from "./types";
 import {
+  blueLighten3,
   blueLighten4,
+  blueLighten5,
   spaceS,
   textBlockWidth,
   themeBgSecondary,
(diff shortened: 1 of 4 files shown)


packages/promo/style.ts
```diff
@@ -42,17 +44,30 @@ export const graphicContainer = css`
...
-      `
-    : "";
+export const getBackgroundGradient = (gradientStyle?: GradientStyle) => {
+  if (!gradientStyle) {
+    return "";
+  }
+
+  const stops = gradientStyles[gradientStyle];
+
+  if (gradientStyle === "oceanBlue") {
```
clintonmedbery, line 59: Could we base this on stops.length instead? Not sure how I feel about having oceanBlue just hardcoded here. Not sure it's a huge deal considering these are essentially themes, but that might make this more extensible.

clintonmedbery (commented):
Nice work. Just mentioned one issue.

packages/promo/style.ts
mstruebing: I'm not sure I agree here, this component can not be generally used as we only provide currently three options.
So if we add one other we probably want some other stops here.
On the other hand, if we would have always the same stops, we could make it very generic and calculate the single stop values from the total amount of the stops.

It's more of a question what we want here, generic component or some specific gradients.

vacas5: the stops for `oceanBlue` can't be mathematically deducted. This is a custom implementation thus, no logic. My original take had some advanced JS trying to calculate the %s based on # of stops but then I saw what design actually asked for.

mstruebing (approved):
I'm okay with that and as long as we only expose specific gradients it doesn't make a big difference to me.

vacas5:
at some point we might consider allowing this component to accept a custom background, but I think this is the shortest way home right now.

packages/promo/style.ts
clintonmedbery: Ok that makes sense.

PR #608, opened 2021-08-18
chore: reproduce field list column issue

mstruebing:
## Checklist for PR creator

- [ ] If any new components were added, there are exported from `packages/index.ts`
- [ ] If this PR is associated with a JIRA, it is mentioned in commit message footer ("Closes …")
- [ ] If this PR contains breaking changes, is stated in commit message body ("BREAKING CHANGE: …")
- [ ] Info for applicable sections above is provided

PR #609, opened 2021-08-18
fix: controlled FieldList row deletion

vacas5:
It seemed as though the onRemoveItem was never actually being called
based on the way that it was written, because the useEffect would not
fire, thus you'd end up with rendering issues. By tweaking a bit how
this function is called in the context, it appears to fix the issue.

I don't know why this callback function was written with the additional
closure, but I preserved it to maintain backwards compatibility.

Addresses D2IQ-77909

## Checklist for PR creator

### DIFF
--- a/packages/formStructure/components/FieldListContext.tsx
+++ b/packages/formStructure/components/FieldListContext.tsx
@@ -10,7 +10,7 @@ type FieldListContext = {
     rowId?: string | number,
     itemToAdd?: Record<string, any>
   ) => void;
-  removeListItem: (rowIndex: number) => () => void;
+  removeListItem: (rowIndex: number) => void;
 };
 
 export const Context = React.createContext<FieldListContext | null>(null);
@@ -46,12 +46,11 @@ export const FieldListProvider: React.FC<FieldListProviderProps> = ({
 
   const removeListItem = (rowIndex: number) => {
     if (onRemoveItem) {
-      return onRemoveItem(rowIndex);
+      const handleRemove = onRemoveItem(rowIndex);
+      handleRemove();
     }
 
-    return () => {
-      setData(data.filter((_, i) => rowIndex !== i));
-    };
+    setData(data.filter((_, i) => rowIndex !== i));
   };
 
   return (
--- a/packages/formStructure/components/FieldList.tsx
+++ b/packages/formStructure/components/FieldList.tsx
@@ -91,6 +91,10 @@ const FieldListRow: React.FC<FieldListRowProps> = React.memo(
       }
     };
 
+    const handleRowClick = () => {
+      fieldListContext?.removeListItem(rowIndex);
+    };
+
     return (
       <div className={getFieldRowGrid(columns, iconSizes[REMOVE_ICON_SIZE])}>
         {columns.map((col, i) => {
@@ -126,7 +130,7 @@ const FieldListRow: React.FC<FieldListRowProps> = React.memo(
         {/* this wrapper div is needed to fix a vertical centering bug in Firefox with <button> elements that are children of display: grid */}
         <div>
           <ResetButton
-            onClick={fieldListContext?.removeListItem(rowIndex)}
+            onClick={handleRowClick}
             disabled={isRowDisabled(rowIndex, disabledRows)}
             onKeyUp={addEmptyRow}
             data-cy="fieldList-removeButton"


packages/formStructure/components/FieldListContext.tsx
```diff
@@ -46,12 +46,11 @@ export const FieldListProvider: React.FC<FieldListProviderProps> = ({
 
   const removeListItem = (rowIndex: number) => {
     if (onRemoveItem) {
-      return onRemoveItem(rowIndex);
+      const handleRemove = onRemoveItem(rowIndex);
```
TattdCodeMonkey, line 49: I think this is a breaking change.

The previous functionality returned a curried function, calling the function directly changes the behavior. 

Additionally not returning and calling the callback and `setData` changes the expected behavior as well since as previously written if you set `onRemoveItem` the parent component was responsible for updating the data. This is how the component is used in several places.

TattdCodeMonkey (changes requested):

packages/formStructure/components/FieldListContext.tsx
vacas5: Check it out again, though. I only changed the handling within `FieldListContext`. `FieldList` still accepts a curried function. Is `FieldListContext` exposed as a component or just a part of `FieldList`?

vacas5: After looking again I'm pretty sure that FieldListContext is only used inside of FieldList, so we're free to refactor as long as the props on FieldList don't change.

clintonmedbery: FieldListAddButton uses it as well, but only the addListItem.

TattdCodeMonkey: 🤷‍♂️  tested this and seems to work and resolve the issue.

TattdCodeMonkey (approved):

packages/formStructure/components/FieldListContext.tsx
clintonmedbery: I am still trying to understand this so I might be off base but I don't think we are actually using removeItem as a curried function anywhere on kommander-ui so removing this would still work. If someone did want to access the SyntheticEvent within a function (which is why it is probably curried), this might limit them from doing so, so it is a breaking change.

You can test this by editing line: 
https://github.com/mesosphere/kommander-ui/blob/master/kommander/frontend/src/components/molecules/FormLabelRow.tsx#:~:text=removeItem%3A%20(affectedIndex%3A%20number)%20%3D%3E%20()%20%3D%3E%20void%3B

to be `removeItem: (affectedIndex: number) => (event: React.SyntheticEvent) => void;`

and lines: 
https://github.com/mesosphere/kommander-ui/blob/master/kommander/frontend/src/pages/projects/ProjectForm.tsx#:~:text=removeItem%3D%7B(rowIndex)%20%3D%3E%20()%20%3D%3E%20%7B,%7D%7D

to be 
` removeItem={(rowIndex) => (event) => {
              console.log("event", event);
              onChange({
                namespaceLabels: removeIndex(
                  formData.namespaceLabels,
                  rowIndex
                ),
              });
            }}`
and notice that event is now undefined with these changes. That being said, I am not sure we need to access the event and to me it might be a bit of overkill. Happy to mess with it more once these other tasks are done.

clintonmedbery: If I could get an example of a render failure and steps to recreate I could understand better.

mstruebing: @clintonmedbery here is an error: https://github.com/dcos-labs/ui-kit/pull/608

mstruebing (approved):
I tested it on kommander-ui and it is working there too.

If we care about a breaking change we could change the commit message and do a major version bump.

My feeling is that it would be good practice to be able to have access to the event, but I'm not to attached to it.

vacas5:
this is not a breaking change as written. Merging in so we can integrate into kommander-ui. We can handle a refactor with a breaking change later

vacas5:
some additional notes

there was FieldList which accepted an onRemove call back as a curried function (no idea why)

but then nested inside it had its own FieldListContext which then exposed a removeItem function, that was also curried. However this removeItem function was not properly firing the FieldList onRemove which it was supposed to, which meant upstream state wasn't changing and resulting in the weirdness.

when I refactored the removeItem function to not be curried, it worked as expected. Since that removeItem was only being used by the component itself, and not a part of the ui-kit library of components, it was a simple fix commit and not a breaking change. My guess is that this was a regression that was introduced when the FieldListContext was added to the component.

PR #610, opened 2021-08-27
D2IQ-77946 Update to Storybook settings - Badge Component updates

nataliepina:
A majority of this is to break down the prep for the upgrade to the latest v6+ update. This will serve as a POC to display what the new Component Story Format will look and function like.

The Badge component has been converted to a functional component, and the functionality remains the same. 

This is just the initial start of changes to come, feedback is welcome. 

## Testing

Checkout the branch and spin up Storybook. Be sure to click on the "Docs" tab in the top panel and drop down some `Code` examples.  View the badge component specifically. 

## Trade-offs

With this change, addons for "info" and "readme" will be removed both depreciated, and will be replaced with "addon-docs" which I have included here. In subsequent changes, I will be converting all stories to align with this change. For now, we will still have access to "knobs" however that will be replaced with "controls" in v6+. Overall more changes to come with the migration to Storybook 6+.  

An additional svg related dependency and configuration had to be added to gain the ability to render the brand logo for theming. 

## Screenshots

Screen Shot 2021-08-27 at 11 21 41 AM

## Checklist for PR creator


packages/badge/components/ColorCodedBadge.tsx
```diff
@@ -9,15 +9,16 @@ import { iconSizes } from "../../shared/styles/styleUtils/layout/iconSizes";
 interface ColorCodedBadgeProps {
   color?: React.CSSProperties["color"];
   iconShape?: SystemIcons;
+  children: JSX.Element | string;
 }
 
 const ICON_SIZE = "xxs";
 
-const ColorCodedBadge: React.FC<ColorCodedBadgeProps> = ({
+const ColorCodedBadge = ({
```
vacas5, line 17: I am not a TS expert, but I thought that the React.FC was helpful boilerplate for components. Was this something that we needed to change to support something else? Is it because children is not normally meant to be a string?

vacas5 (commented):
wondering if these could possibly be split into two PRs. Also, wondering about that package-lock.json. Maybe that's based on you using a much newer version of Node than with the file that was committed

packages/badge/tests/__snapshots__/badgeButton.test.tsx.snap
```diff
@@ -82,7 +82,7 @@ exports[`BadgeButton success 1`] = `
 
 exports[`BadgeButton warning 1`] = `
 <span
-  className="css-thsdwi"
+  className="css-1i0swvi"
```
vacas5, line 85: I ran into this recently. We may reconsider snapshots if this is going to happen over and over

packages/badge/components/badge.tsx
```diff
@@ -11,23 +11,17 @@ export type BadgeAppearance =
...
+const Badge: React.FC<BadgeProps> = ({
+  appearance = "default",
+  children
+}: BadgeProps) => {
+  return (
+    <span className={badge(appearance)} data-cy="badge">
+      {children}
+    </span>
+  );
+};
```
vacas5, line 26: so, as I understand it, to strictly recreate a Pure Component, we'll need to memoize it by wrapping it in React memo: https://stackoverflow.com/questions/43470659/declare-a-functional-component-as-pure

packages/badge/components/badgeButton.tsx
```diff
@@ -13,38 +13,31 @@ export interface BadgeButtonProps {
...
-      </Clickable>
-    );
-  }
-}
+  return (
+    <Clickable action={onClick} tabIndex={tabIndex} data-cy="badgeButton">
+      <span className={className}>{children}</span>
+    </Clickable>
+  );
+};
```
vacas5, line 41: same goes here. We'll have to memoize it.

vacas5: my guess is that the memoization is the whole reason these are classes (so he could leverage PureComponent)

packages/badge/index.ts
```diff
@@ -1,3 +1,3 @@
-export { default as Badge } from "./components/badge";
-export { default as BadgeButton } from "./components/badgeButton";
+export { default as Badge } from "./components/Badge";
+export { default as BadgeButton } from "./components/BadgeButton";
```
vacas5, line 2: did you mean to capitalize these? the files themselves are still lowercase

packages/textInput/components/TextInputWithBadges.tsx
```diff
@@ -19,7 +19,7 @@ import { Icon } from "../../icon";
 import Clickable from "../../clickable/components/clickable";
 import { TextInputWithIconProps } from "./TextInputWithIcon";
 import { TextInputWithIcon } from "..";
-import { BadgeAppearance } from "../../badge/components/badge";
+import { BadgeAppearance } from "../../badge/components/Badge";
```
vacas5, line 22: here too

packages/badge/tests/colorCodedBadge.test.tsx
```diff
@@ -10,12 +10,16 @@ expect.addSnapshotSerializer(serializer);
...
+      <ColorCodedBadge color="#f00f00">Color Coded Badge</ColorCodedBadge>
+    );
     expect(toJson(component)).toMatchSnapshot();
   });
   it("renders with an icon and color", () => {
     const component = shallow(
-      <ColorCodedBadge color="#f00f00" iconShape={SystemIcons.CircleCheck} />
+      <ColorCodedBadge color="#f00f00" iconShape={SystemIcons.CircleCheck}>
+        Color Coded Badge with icon
+      </ColorCodedBadge>
```
vacas5, line 22: does this change reflect a change in how the component works? If I read correctly, it always worked this way, you're just improving the test. Wanted to double check. Maybe change one test to take children and leave the other one without?

vacas5: on second thought, if you want to test that it renders the child correctly, I'd add a new test entirely and leave these alone. Would give us greater confidence in the refactor

vacas5 (commented):
I would look into if there's a way you could test the memoization part of PureComponent, and add that to the unit tests for these components, and then make sure that your refactor behaves the exact same way

packages/badge/components/ColorCodedBadge.tsx
nataliepina: This has been a recent discussion in the community about this and `React.FC` has been discouraged against. I have been trying to avoid it if possible. Due to how it implicitly takes children of type ReactNode, and issues with default props. [More Detail Here. ](https://github.com/facebook/create-react-app/pull/8177)

I have been following this handy [TypeScript Guide ](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/) which mentioned it as well.

vacas5: Okay great! way to lead us forward!

packages/badge/components/badgeButton.tsx
nataliepina: I was thinking about this. I am curious about everyone's opinion here in regard to the PureComponent or memo. It seems like it may be more of a benefit to have those added if there is complex business logic, the component is doing calculations on every render, or the component is displaying some janky behaviors. I am not sure if this would be the case with something like a Badge component. Although I am not sure on the other hand if the benefits outweigh to optimize anyway by adding the useMemo logic to all the components instead if we want functional components. Thoughts? 🤔

clintonmedbery (commented):
Looks good, I would take a look at the PureComponent stuff and try to see what works best.

packages/badge/index.ts
nataliepina: I had capitalized these files and something in the ui-kit appears to be reverting that change, digging into why that may be happening.

packages/badge/components/badge.tsx
```diff
@@ -11,23 +11,17 @@ export type BadgeAppearance =
...
-    return (
-      <span className={badge(appearance)} data-cy="badge">
-        {children}
-      </span>
-    );
-  }
+  children: React.ReactNode | string;
 }
 
+const Badge: React.FC<BadgeProps> = ({
```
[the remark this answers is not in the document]
nataliepina: Good call 😄

packages/badge/components/ColorCodedBadge.tsx
```diff
@@ -42,4 +43,4 @@ const ColorCodedBadge: React.FC<ColorCodedBadgeProps> = ({
   );
 };
 
-export default ColorCodedBadge;
+export default React.memo(ColorCodedBadge);
```
vacas5, line 46: sorry I know this is annoying. This one wasn't memoized before so doesn't need it here

packages/badge/index.ts
vacas5: think you may need to abandon this change. These files are still lower case

vacas5 (changes requested):

vacas5 (approved):

PR #611, opened 2021-08-30
D2IQ-77455 chore: added config-cycle icon

nataliepina:
Closes D2IQ-77455

This is the first step in transitioning the platform-service-support icon to the new system icon name - config-cycle. 

## Trade-offs

We will be keeping the PlatformServiceSupport icon in the ui-kit until we are safe to transition after updates to kommander-ui to utilize ConfigCycle. 

## Screenshots

Screen Shot 2021-08-30 at 9 41 59 AM

## Checklist for PR creator


vacas5 (changes requested):
does this need to be exported in any other files to be use-able?

Also, this should be a `feat` commit as it introduces something new

PR #618, opened 2021-09-17
Update to Emotion 11

nataliepina:
Closes D2IQ-73724

[Emotion 11 Migration Info](https://emotion.sh/docs/emotion-11)✨
The majority of the changes here are updating packages and imports.  
In our case:

```
@emotion/core → @emotion/react
emotion → @emotion/css
emotion-theming → moved into @emotion/react
jest-emotion → @emotion/jest
```

Additionally, some changes for the Theme type. The snapshot tests needed to be updated to pass. 

## Testing

Open up Storybook and ensure components are looking the same. 

## Checklist for PR creator


packages/appChrome/types.ts
```diff
@@ -1,4 +1,3 @@
-import { Theme } from "../themes/types/appTheme";
```
vacas5, line 1: was this used anywhere else? Shoudl the file be deleted?

packages/themes/types/appTheme.ts
```diff
@@ -1,6 +1,12 @@
+import { AppChromeTheme } from "../../appChrome/types";
+
 type Color = React.CSSProperties["color"];
+export type LocalTheme = AppChromeTheme & {
```
vacas5, line 4: single `&`? must be a TypeScript thing I'm unfamiliar with

packages/themes/types/appTheme.ts
```diff
@@ -44,4 +50,10 @@ export interface Theme {
     shadow?: Color;
     shadowInverted?: Color;
   };
+};
+
+declare module "@emotion/react" {
+  export interface Theme extends LocalTheme {
+    _unused: true; // cannot be empty
+  }
```
vacas5, line 58: could you explain a bit what's happening in this file?

vacas5 (commented):
a couple of questions, all looks well, just want to do a quick regression test

packages/themes/types/appTheme.ts
nataliepina: I was unfamiliar as well, it's an 'intersection type' so that it includes the types from AppChromeTheme and whatever is joined with the &.

packages/themes/types/appTheme.ts
nataliepina: So for the changes here, I referred to the previous PR related to the Emotion theme updates that looked nearly complete, and their solution seemed to work here. 
I do know with the Emotion 11 update the docs state to update this:
```
It’s now easier to provide a type for Theme. Instead of creating custom instances (like before) you can augment the builtin Theme interface like this:

declare module '@emotion/react' {
  export interface Theme {
    primaryColor: string
    secondaryColor: string
  }
}
```

But here we are extending the `LocalTheme` so it's formatted differently. I am assuming the `_unused: true` is due to potential TS issues 🤔 The "cannot be empty" comment included had me not wanting to remove that bit.

packages/appChrome/types.ts
nataliepina: Imports elsewhere will now be `import { LocalTheme as Theme } from "../types/appTheme";` with AppChromeTheme no longer extending Theme here it could be removed.

packages/themes/types/appTheme.ts
nataliepina: Here's a link to the other PR - https://github.com/dcos-labs/ui-kit/pull/578/files 
Let me know if you think I should find a different route for these changes instead.

mstruebing (approved):
LGTM 🚀

PR #619, opened 2021-09-20
Emotion 11 Update

nataliepina:
BREAKING CHANGE: new emotion version and export theme exports changed

Closes D2IQ-73724

## Checklist for PR creator


TattdCodeMonkey:
I think we should label this a `feat` commit just to be safe as we updated code and dep. 

`chore` should be for project changes that don't have direct code changes, like storybook updates etc.

vacas5:
I'd go with Rodney's suggestion here just to be safe

PR #622, opened 2021-09-24
Fix for Flexbox spacing issues with Emotion 11 updates

nataliepina:
This issue was unearthed with the Emotion 11 updates. There were a few places in Kommander UI on Emotion 11 that displayed slightly off spacing, due to the padding-right and padding-bottom being added in the `Flex` component. 

It is possible that the `@support` feature query was not actually being applied until everything became updated. The feature query should be used to apply padding when the browser does not support `row-gap` or `column-gap`. According to [caniuse](https://caniuse.com/?search=column-gap) it's mostly applicable for IE and common browser versions prior to ~2017. We may decide to remove this feature query in the future if it is confirmed that we no longer need to support IE. 

[More info on how Feature Query works.](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports)

## Testing

View Flexbox story 'With Buttons', which gives a sandbox replicating the issue and fix. 

## Screenshots

Seen in Kommander UI. 👇

Screen Shot 2021-09-24 at 11 11 55 AM

Before:

Screen Shot 2021-09-23 at 3 36 28 PM

After:

Screen Shot 2021-09-23 at 3 36 56 PM

## Checklist for PR creator

### DIFF
--- a/packages/codesnippet/tests/__snapshots__/CodeSnippet.test.tsx.snap
+++ b/packages/codesnippet/tests/__snapshots__/CodeSnippet.test.tsx.snap
@@ -52,9 +52,14 @@ exports[`CodeSnippet renders default 1`] = `
   width: auto;
 }
 
-@supports (column-gap: 1px) and (row-gap: 1px) {
+@supports not (column-gap: 1px) {
   .emotion-3>*:not(:last-child) {
     padding-right: 0;
+  }
+}
+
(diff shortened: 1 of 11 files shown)


vacas5 (approved):
Natalie and I paired up on this one yesterday. Awesome work recreating in here so we could confirm the fix 💪

PR #627, opened 2021-11-01
feat: update node

nataliepina:
Updated the version per - https://hub.docker.com/layers/node/library/node/16.13.0-alpine3.14/images/sha256-19956c794ae17fdc7115124ca3caf2df84c7c5e73c968016697239ca5c245532?context=explore

Snapshots had to be updated here.

## Checklist for PR creator


vacas5 (commented):
Should we just bite the bullet and go all the way to 16? node lts is now 16 (I think that's recent). Also, I think we should lock the dockerfile to a specific node version the same way we have it locked here: https://github.com/dcos-labs/ui-kit/blob/master/Dockerfile#L1

I am okay with either changing the dockerfile to point to 14.17.6 or changing both to 16

nataliepina:
> Should we just bite the bullet and go all the way to 16? node lts is now 16 (I think that's recent). Also, I think we should lock the dockerfile to a specific node version the same way we have it locked here: https://github.com/dcos-labs/ui-kit/blob/master/Dockerfile#L1
> 
> I am okay with either changing the dockerfile to point to 14.17.6 or changing both to 16

Sounds like a plan @vacas5, updated! 🎉

vacas5 (approved):

PR #629, opened 2021-11-04
fix: for console errors found in kommander

nataliepina:
Many of these errors were due to the pass-through props on the `TextInputWithBadges`.

I would like to fix the lifecycle methods more so when we are able to have time to incrementally update and convert UI Kit components. 

## Testing

Check console errors while Storybook is running.

## Checklist for PR creator

### DIFF
--- a/packages/textInput/components/TextInputWithBadges.tsx
+++ b/packages/textInput/components/TextInputWithBadges.tsx
@@ -77,6 +77,7 @@ export class TextInputWithBadges extends TextInputWithIcon<
       onBadgeChange,
       downshiftReset,
       addBadgeOnBlur,
+      badgeAppearance,
       ...inputProps
     } = baseProps as TextInputWithBadgesProps;
     inputProps.onKeyDown = this.handleKeyDown;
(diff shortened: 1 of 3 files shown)


packages/textInput/components/TextInputWithBadges.tsx
```diff
@@ -119,33 +121,37 @@ export class TextInputWithBadges extends TextInputWithIcon<
...
-                              <Icon shape={SystemIcons.Close} size="xxs" />
-                            </span>
-                          </Clickable>
-                        </FlexItem>
-                      </Flex>
-                    </Badge>
-                  </span>
-                ))}
+                this.props.badges.map(badge => {
+                  const { badgeAppearance = "primary" } = this.props;
```
vacas5, line 125: we're already destructuring `this.props` above. Should we just do all of these there? unsure why this would make a difference

nataliepina: Thought I was still seeing the error with it there while troubleshooting, but it looks likes what I've done elsewhere resolves that. Thanks!

vacas5 (approved):

PR #631, opened 2021-11-08
Update tabs and add story controls

nataliepina:
Closes D2IQ-62145

Updates:
- The spacing between tabs has been increased from 16px -> 24px. This aligns with the Spacing Standards in the Design System https://mesosphere.invisionapp.com/console/Kommander-GA-cknny74gbi6ql01y3t8vc22o0/cknny9h71i8fy01y3z0z129of/play
- While I was in there I decided to give the component an update to a FC, functionality remains the same after testing within Storybook.
- The Story has been updated to the new template format, there is now the option to use controls to change the direction prop.

## Testing

Test the Tabs Story. 📖 

##### Before:
Screen Shot 2021-11-08 at 7 05 58 AM

##### After:
Screen Shot 2021-11-08 at 7 05 45 AM

## Checklist for PR creator

### DIFF
--- a/packages/tabs/stories/Tabs.stories.tsx
+++ b/packages/tabs/stories/Tabs.stories.tsx
@@ -1,49 +1,62 @@
 import * as React from "react";
 import { TabTitle, Tabs, TabItem } from "../index";
-import { TabDirection } from "../components/Tabs";
+import { TabsProps } from "../components/Tabs";
+import { Story, Meta } from "@storybook/react";
 
-class Example extends React.Component<
-  { direction?: TabDirection },
-  Partial<{ selectedIndex: number }>
-> {
-  state = { selectedIndex: 0 };
-  handleSelect = selectedIndex => {
-    this.setState({ selectedIndex });
-  };
-  render() {
-    const { selectedIndex } = this.state;
-    return (
-      <Tabs
-        selectedIndex={selectedIndex}
-        onSelect={this.handleSelect}
-        direction={this.props.direction}
-      >
-        <TabItem>
-          <TabTitle>Tab 1 Name</TabTitle>
-          <div>First tab Content</div>
-        </TabItem>
-        <TabItem>
-          <TabTitle>Tab 2 Name</TabTitle>
-          Second Tab Content
-        </TabItem>
-      </Tabs>
-    );
-  }
-}
+const TabsStory = ({ direction }: TabsProps) => {
+  const [selectedIndex, setSelectedIndex] = React.useState(0);
+
+  return (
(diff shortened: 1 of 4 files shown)


mstruebing (approved):
Good work :+1

packages/tabs/stories/Tabs.stories.tsx
```diff
@@ -1,49 +1,65 @@
...
+} as Meta;
+
+const Template: Story<TabsProps> = args => <TabsStory {...args} />;
 
-export const Default = () => <Example />;
+export const Horizontal = Template.bind({});
+Horizontal.args = { direction: "horiz" };
 
-export const Vertical = () => <Example direction="vert" />;
+export const Vertical = Template.bind({});
```
mstruebing, line 54: Would it make sense to remove one of them because it can now be controlled from the controls anyway?

nataliepina: I'm glad you pointed this out, so yes :)

mstruebing (dismissed):
We should probably get rid of that merge commit, but other than that it's good 🎉

vacas5 (changes requested):
This needs to be a BREAKING CHANGE, unless we want to consider it a bug fix. It's probably a `fix` with a BREAKING CHANGE. Should anyone use this with layout, we'd want them to know to expect it to look doubly different.

mstruebing (dismissed):
I'm not sure this really needs to be a BREAKING CHANGE as nothing really breaks and works differently than before, it's only a spacing alignment that is specified in the design system and we haven't implemented it correctly, just my two cents.

If we want this to be a breaking change I'm fine with that, therefore approving.

nataliepina:
> I'm not sure this really needs to be a BREAKING CHANGE as nothing really breaks and works differently than before, it's only a spacing alignment that is specified in the design system and we haven't implemented it correctly, just my two cents.
> 
> If we want this to be a breaking change I'm fine with that, therefore approving.

@vacas5  final consensus here? Good for me to make a mental note for the future either way for all the best times to use `BREAKING CHANGE`

vacas5:
Let me just verify my assumptions. I could be using the default implementation of this component, and then with this change the padding will double. That will likely mean that I will have to change something. Though, if it only makes this component different, I could see where it might not be a breaking change: https://medium.com/eightshapes-llc/visual-breaking-change-in-design-systems-1e9109fac9c4

vacas5 (approved):
as long as something couldn't actually break as a result of this padding change, I think we're okay.

clintonmedbery (approved):

PR #638, opened 2021-11-18
fix(package.json): remove node dependency from package.json

seialkali:
Removing node from package.json because it shouldn't need to be listed as a dependency, and also it's causing issues when trying to run `npm i` on an M1 Mac

## Testing

Run `npm i` to check you can run the installs successfully. Maybe click on some of the components in the storybook to check that everything seems to be running ok.

## Checklist for PR creator

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,6 @@
         "fs-extra": "10.0.0",
         "jest": "23.6.0",
         "jest-runner-prettier": "0.2.6",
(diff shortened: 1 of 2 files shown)


mstruebing (approved):
👍  Works also on non M1-machines

nataliepina (approved):
Awesome, thanks for fixing this up! Building Storybook took half the time, we may have unearthed the source of the slowness.

PR #639, opened 2021-11-22
feat: improve grid features and alignment options

nataliepina:
I have done this in a way that I do not believe will break any of the styles and `GridList` usage in Kommander. Columns can still be explicitly set, have responsive options, but now will have a backup style that is a fluid grid. GridList items can be centered with a new optional prop, which provides some additional benefits such as the containers shrinking to fit their content instead of stretching. 
I've taken the opportunity here to update the story and create some useful Controls which can be used to compare to the old functionality of the GridList here - https://dcos-labs.github.io/ui-kit/?path=/story/layout-gridlist--column-count
I am still using BREAKING CHANGE to be safe, mostly due to the change in how to use some of the `GridList` props.

BREAKING CHANGE
Before this fix columnCount was required, new behavior includes defaults for when it is not set to become a fluid grid that sets the amount of rows and columns based on available space.
Before you could not center grid items, now the optional centerItems prop can be used.
Other features include a data-cy prop, and className for further customizations.

Closes D2IQ-81475

## Testing

Test out the `GridList` controls. 

## Checklist for PR creator

### DIFF
--- a/packages/styleUtils/layout/gridList/components/GridList.tsx
+++ b/packages/styleUtils/layout/gridList/components/GridList.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { css } from "@emotion/css";
+import { css, cx } from "@emotion/css";
 import { gridColumnTemplate } from "../style";
 import { SpaceSize } from "../../../../shared/styles/styleUtils/modifiers/modifierUtils";
 import {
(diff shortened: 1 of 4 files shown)


packages/styleUtils/layout/gridList/components/GridList.tsx
```diff
@@ -8,46 +8,62 @@ import {
...
+  className?: string;
+
+  /**
+   * Human-readable selector used for writing tests.
+   */
+  ["data-cy"]?: string;
+  /**
+   * Centers grid children by implementing `place-items: center;` Optimal for grids that could have a state with only 1 grid child.
+   */
+  centerItems?: boolean;
```
vacas5, line 40: need a bit to chew on this. There's a vertical and a horizontal centering to consider, and this one property would by default set both. it is pretty different than the current default behavior.

packages/styleUtils/layout/gridList/style.ts
```diff
@@ -7,7 +7,9 @@ export const gridColumnTemplate = columnCount => {
           acc[breakpoint] = `repeat(${columnCount[breakpoint]}, 1fr)`;
           return acc;
         }, {})
-      : `repeat(${columnCount}, minmax(0, 1fr))`;
+      : columnCount
+      ? `repeat(${columnCount}, minmax(0, 1fr))`
+      : `repeat(auto-fill, minmax(min(320px, 100%), 1fr));`;
```
vacas5, line 12: could this have backwards compatibility issues?

vacas5 (commented):
🤔  <- me rn

packages/styleUtils/layout/gridList/style.ts
nataliepina: My thought was that this here protects us from breaking anything where the columnCount was explicitly set to a single number before. If we don't set any columnCount, it will default to the last part which should work in a lot of scenarios.

packages/styleUtils/layout/gridList/components/GridList.tsx
nataliepina: 🤔 By default, this will not be true. I was hoping to simplify centering by using place-items as a one-liner to toggle on or off. Another option here might be to give more options for the user to enter to customize place-items, or break it out into justify-content and align-items.

vacas5: yeah I guess you are right, we could always break this out into two props

packages/styleUtils/layout/gridList/style.ts
vacas5: okay cool!

vacas5 (approved):

mstruebing (approved):
I manually tested the changes in kommander-ui and nothing seems broken 👍

nataliepina:
> I manually tested the changes in kommander-ui and nothing seems broken 👍

Thank you for the extra testing Max 💯

PR #641, opened 2021-11-24
fix: old npm version set in devDependencies

nataliepina:
## Checklist for PR creator


mstruebing (approved):
Good find, I think this shouldn't be in there in the first place 👍

PR #645, opened 2021-11-29
chore: update react related dependencies

nataliepina:
Separated out these react-related dependencies, which required some snapshot updates. Components are still working as expected and all tests are passing. 

## Checklist for PR creator


package.json
```diff
@@ -54,12 +54,12 @@
     "react-chartist": "0.14.4",
     "react-click-outside": "3.0.1",
     "react-delegate-component": "1.0.0",
-    "react-draggable": "3.3.2",
```
mstruebing, line 57: Are we even using this dependency?

nataliepina: I was unsure about this one, didn't see it actively used here but I saw it in Kommander and wasn't sure if it's safe to remove. Thoughts?

mstruebing: In kommander-ui I can see it only in `package-lock.json` and a markdown file. So I think it is not really used and should be safe to be removed.

nataliepina: Done! Thanks @mstruebing

mstruebing (approved):
🎉 Good stuff

seialkali (approved):
Looks good 🕺

PR #652, opened 2021-12-03
add controls for typography and style utils

nataliepina:
Converted stories under Typograph and Style Utils. 

## Checklist for PR creator


clintonmedbery (approved):
😎 Nice. I am just wondering. I see some have a setObject control. Is it possible to provide these with some options or context?

nataliepina:
> 😎 Nice. I am just wondering. I see some have a setObject control. Is it possible to provide these with some options or context?

Which stories were you seeing that on? There should be more context for each prop under the Docs tab. I've been trying to limit those controls to eliminate confusion on how to use, Storybook is "smart" and can pull in most controls on its own based on the props so that may be happening and I missed it. I need to do some more research to see if there is a way to add some sort of documentation with the control.

clintonmedbery:
Box -> Default. Ok I see a dropdown with an example in the options. That is neat.

packages/styleUtils/typography/stories/text.stories.tsx
```diff
@@ -1,25 +1,92 @@
...
+  DangerText,
+  InteractiveText,
+  MonospaceText,
+  SmallText,
+  SuccessText,
+  Text,
+  WarningText
+} from "../";
+
+const colors = {
```
seialkali, line 20: This should be an array

packages/styleUtils/typography/stories/text.stories.tsx
```diff
@@ -31,114 +98,36 @@ export const Default = () => (
...
-
-export const Color = () => {
-  const colors = {
-    themeTextColorPrimary,
-    themeTextColorSecondary,
-    themeTextColorDisabled,
-    blue
-  };
-  const color = select("color", colors, themeTextColorPrimary);
+export const _SuccessText = args => (
```
seialkali, line 113: All of the different types of text have the same styles. Should they be different?

packages/styleUtils/typography/stories/textBlock.stories.tsx
```diff
@@ -47,9 +48,11 @@ export const Default = () => (
...
 
-export const UsingTypographyComponents = () => (
+export const Default = Template.bind({});
+
+export const UsingTypographyComponents = args => (
   <Container>
-    <TextBlock>
+    <TextBlock {...args}>
       <div>
         <HeadingText1>Lorem Ipsum</HeadingText1>
```
seialkali, line 57: It might be useful to say which typography components are being used within this the TextBlock story, in case the user isn't aware of what's considered a Typography components.

seialkali (commented):
This looks really good! I made a few suggestions for how I think some of these stories can be improved. I also found a few options objects that will need to be changed into arrays. 

On a separate note, I noticed a few issues that I'm going to make tickets for. They're not really related to what you're working on here, so no point fixing them as part of this PR:

- I noticed in the `BorderedBox` and `Box` components that selecting certain tags causes the component to blow up, for instance `link` and `input`. It might be worth removing those as options from the prop type in the components themselves.

- I’m getting an intermittent console error in these `CodeSnippet` stories: `Can't perform a React state update on an unmounted component`. I think this is coming from the `setTimeout` not being cleared in `ClickToCopy.tsx`.

packages/styleUtils/typography/stories/text.stories.tsx
nataliepina: They should be different, they're different components with their own appearance. Will fix!

packages/styleUtils/typography/stories/textBlock.stories.tsx
nataliepina: This is a good idea! I will add this to improve the story using UI Kit components. Users can also see every component used when they click the "Show Code" on the Docs page to see how each story was written.

seialkali (approved):

PR #653, opened 2021-12-03
add controls for navigation and layout

nataliepina:
Converted stories under Navigation and Layout. 

## Checklist for PR creator


clintonmedbery (approved):
Nice work! Any thoughts on removing the knobs panel if we don't use it?

nataliepina:
> Nice work! Any thoughts on removing the knobs panel if we don't use it?

I still have one more PR coming to remove knobs for some of the other components. Then I will remove it from existence 😄

PR #655, opened 2021-12-06
chore: update github actions

nataliepina:
## Checklist for PR creator

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 # This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-name: ui-kit actions
+name: UI Kit Actions
 
 on:
   push:
@@ -25,6 +25,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: "npm"
-      - run: npm ci
-      - run: npm test
-      - run: npm run lint
+      - name: Continuous Integration
+        run: npm ci
+      - name: Install Dependencies
+        run: npm install
+      - name: Lint
+        run: npm run lint
+      - name: Build
+        run: npm run dist
+      - name: Unit Tests
+        run: npm test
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -25,6 +25,9 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: "npm"
+      - name: Install dependencies
+        run: npm install
       - run: npm ci
+      - run: npm run dist
```
vacas5, line 31: I think we should have names for all of these that match the old names.

Also, order matters. Run lint first, because it takes like 5 seconds but if it fails, we shouldn't do any of the other stuff

Jenkinsfile
```diff
@@ -113,4 +104,4 @@ pipeline {
       }
     }
   }
-}
+}
```
vacas5, line 107: I think you should remove the other steps too!

vacas5 (changes requested):

Jenkinsfile
nataliepina: It breaks the integration test if I remove some of the other steps. If we can get that integration action merged in first I can give it another go.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,13 @@ jobs:
...
           cache: "npm"
-      - run: npm ci
-      - run: npm test
-      - run: npm run lint
+      - name: Install Dependencies
+        run: npm install
+      - name: Lint
+        run: npm run lint
+      - name: Continuous Integration
+        run: npm ci
```
mstruebing, line 33: npm ci basically installs all dependencies. So you can just run `npm ci` in step one instead of `npm install`.
https://docs.npmjs.com/cli/v8/commands/npm-ci

nataliepina: Wasn't sure where to place that one. Thanks!

mstruebing (dismissed):

mstruebing (commented):
Way better than that jenkins stuff! 💯

mstruebing (approved):

vacas5 (approved):

PR #659, opened 2021-12-10
Github Action - Deployment

nataliepina:
We are transferring from Jenkins to Github Actions. This is to set up the deployment action, which is transferred from how the deployment is handled in the jenkinsfile. We first need to install npm and then we need to deploy Storybook, the user is confirmed by authentication with `GH_TOKEN: ${{ secrets.GH_TOKEN }}`.  This process should deploy our storybook to the live site configured with Github pages. 

## Testing

Review the code. We will ultimately need to test this with a merge into master. 

## Checklist for PR creator


clintonmedbery (dismissed):
This is hard to test because it only runs on a push to master. Maybe we can merge it and see what happens 🤷

clintonmedbery:
Let's not merge on Friday afternoon in case the deployment does go wrong 😅

mstruebing (changes requested):
I think we need to use the `setup-node` action as we are executing node/npm commands similar to her: https://github.com/dcos-labs/ui-kit/blob/master/.github/workflows/ci.yml#L23-L27

clintonmedbery (approved):

.github/workflows/ci.yml
```diff
@@ -26,19 +23,51 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: "npm"
       - name: Continuous Integration
-        run: npm ci
-      - name: Install Dependencies
+        run: npm ci 🔄
+      - name: Install Dependencies 🔧
         run: npm install
```
mstruebing, line 28: We do not want to run `install` and `ci` here, what we want is probably only a `npm ci`.
https://docs.npmjs.com/cli/v8/commands/npm-ci

vacas5 (commented):
I think we can split out actions so they can run concurrently. Should we consider pulling this one out? Not sure if there is a way to make them dependent of each other

vacas5 (approved):

mstruebing (approved):
👍

PR #660, opened 2021-12-10
chore: github action for semantic release

nataliepina:
Created by following the outline here -> https://github.com/marketplace/actions/action-for-semantic-release

## Testing
May be difficult to test as this is run when merging into master. 

## Checklist for PR creator

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,3 +35,10 @@ jobs:
         run: npm run dist
       - name: Unit Tests
         run: npm test
+      - name: Semantic Release
+        uses: cycjimmy/semantic-release-action@v2.6.0
+        with:
+          semantic_version: 18
+        env:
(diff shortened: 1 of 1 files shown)


mstruebing (changes requested):
We should make sure to restrict pushes to `master` for everyone including admins if we do it like this.

This workflow doesn't lint, test, or transpiles any code so if we would push to `master` we may publish falsy code, also we do not have the transpiled code from `dist` in that workflow.

I think the easiest way would be to create a single workflow like:

```
- install
- lint
- build
- test
- release (when master)
```

nataliepina:
> 

So do you believe that this should still be a separate

> 

@mstruebing  Are you suggesting to include this in the other workflow that includes the lint and unit tests? Or to include those steps within this file?

mstruebing:
I personally would include it in the other workflow so that we do not need to execute some of these steps twice.

mstruebing (approved):

PR #664, opened 2021-12-13
refactor: typography controls updates

seialkali:
Suggestions for changes/improvements to typography controls. Also includes the addition of a `controlConstants.ts` file to store reusable consts for the stories. 

**Changes of note:**

- Usage of `defaultValue` has been removed from the `argTypes`, and included in the args for the individual stories because it's being depreciated in Storybook 7.0
- I've exclude controls from certain stories where they were irrelevant or caused issues with other necessary controls. For instance:
  - `vertAlignChildren` has been removed from the `DisplayInline` story because once you choose an option for that control it changes the display to `flex` and doesn't allow it to be changed back to `inline` or any other option. Also, the `vertAlignChildren` is irrelevant to the `DisplayInline` story anyway, as you can't use flex styles with `display: inline`.
  - Removed all of the controls except `children` from `CodeSnippet` because all other usages of the component are covered in other stories. I thought it might be worth having a story for the component without any props passed in, but I'm happy to change it back if anyone thinks the default story should still show all the controls.
- Added height and children to Bordered Box & Box (where relevant) to allow user to test out the `vertAlignChildren` control.
- Spread in `args` after any props which are passed into the component, as this allows the those prop values to be changed via the controls. When the args are passed in first, the controls for any props passed in after won't work.
  - **Works:** <Box tag="section" {...args}>
  - **Doesn't work:** <Box {...args} tag="section">

 
## Checklist for PR creator

### DIFF
--- a/packages/storybookHelpers/controlContants.ts
+++ b/packages/storybookHelpers/controlContants.ts
@@ -1,0 +1,60 @@
+export const cssDisplayPropertyValues = [
+  "inline",
+  "block",
+  "contents",
+  "flex",
+  "grid",
+  "inline-block",
+  "inline-flex",
+  "inline-grid",
+  "inline-table",
+  "list-item",
+  "run-in",
+  "table",
+  "table-caption",
+  "table-column-group",
+  "table-header-group",
+  "table-footer-group",
+  "table-row-group",
+  "table-cell",
+  "table-column",
+  "table-row",
+  "none",
+  "initial",
+  "inherit"
+];
+
+export const textAlignValues = [
+  "center",
+  "end",
+  "inherit",
+  "initial",
+  "justify",
+  "left",
+  "right",
+  "start",
+  "unset"
+];
+
+export const vertAlignValues = ["top", "center", "bottom"];
+
+export const spacingSizeValues = [
+  "xxs",
+  "xs",
+  "s",
+  "m",
+  "l",
+  "xl",
+  "xxl",
+  "none"
+];
+
+export const sideValues = [
+  "all",
+  "top",
+  "right",
+  "bottom",
+  "left",
+  "horiz",
+  "vert"
+];
--- a/packages/styleUtils/modifiers/stories/Box.stories.tsx
+++ b/packages/styleUtils/modifiers/stories/Box.stories.tsx
@@ -1,43 +1,41 @@
 import * as React from "react";
-import { pickReadableTextColor } from "../../../shared/styles/color";
 import {
-  greyLightLighten5,
-  red,
-  yellow,
-  green,
-  blue,
-  purple,
   greyLight,
   themeBrandPrimary
 } from "../../../design-tokens/build/js/designTokens";
 import Box, { BoxProps } from "../components/Box";
 import { css } from "@emotion/css";
 import { Story, Meta } from "@storybook/react";
+import {
+  cssDisplayPropertyValues,
+  textAlignValues,
+  vertAlignValues
+} from "../../../storybookHelpers/controlContants";
 
 export default {
   title: "Style Utilities/Box",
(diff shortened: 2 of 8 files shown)


packages/styleUtils/modifiers/stories/Box.stories.tsx
```diff
@@ -57,88 +48,91 @@ export default {
...
-  </Box>
-);
+const setHeight = css`
+  > * {
+    height: 200px;
+  }
+`;
+
+const Template: Story<BoxProps> = args => {
+  const addBorder = css`border: 2px solid ${themeBrandPrimary}`
```
nataliepina, line 58: 🥜  Might be best to move this border outside of the template. Or you can pass in the emotion css inline` <div className={css``border: 2px solid ${themeBrandPrimary}``}>`

packages/storybookHelpers/controlContants.ts
```diff
@@ -0,0 +1,43 @@
...
+  "right",
+  "start", 
+  "unset"
+];
+
+export const vertAlignValues = ["top", "center", "bottom"];
+
+export const spacingSizeValues = ["xxs", "xs", "s", "m", "l", "xl", "xxl", "none"];
+
+export const sideValues = ["all", "top", "right", "bottom", "left", "horiz", "vert"];
```
nataliepina, line 43: Nice! This should be helpful in the future.

packages/styleUtils/typography/stories/headingText.stories.tsx
```diff
@@ -41,14 +31,17 @@ const Template: Story<HeadingTextProps> = args => (
 );
 
 export const DefaultHeadingText1 = Template.bind({});
+DefaultHeadingText1.args = { align: "inherit" };
```
nataliepina, line 34: TIL that `defaultValue` will be depreciated with v7. I've read the docs over some more, what we could do here instead to keep this drier is to pass `args` below the `argTypes` in the configuration above. 
Ex:
```
  args: {
    display: "inline",
  }
```
Now display inline will be applied to all of the stories so we don't have to pass it in individually.

nataliepina: Going to have to revisit all of the defaultValue uses throughout in the near future, at least we have some time until v7 is stabilized. 😅

packages/styleUtils/typography/stories/text.stories.tsx
```diff
@@ -16,23 +10,7 @@ import {
...
-  "justify",
-  "left",
-  "match-parent",
-  "right",
-  "start"
-];
+import { textAlignValues } from "../../../storybookHelpers/controlContants";
 
 export default {
   title: "Typography/Text",
```
nataliepina, line 16: It appears that the control for text color is an object so we could improve these controls by including the `{type: color}` here as well

packages/styleUtils/typography/stories/headingText.stories.tsx
seialkali: Good spot! I actually read that in MIGRATION.md, but misinterpreted it 🤦‍♀️

packages/styleUtils/typography/stories/text.stories.tsx
seialkali: Good shout. Do you think it's worth disabling the `color` control for the stories where a color is set by the component, for instance `Warning` and `Success`?

nataliepina: Yes, good point, was going to mention that as well. Should disable color for Warning/Success to suggest that it shouldn't really be changed.

nataliepina (approved):
👏

PR #665, opened 2021-12-14
refactor: refactor args for stories

nataliepina:
Refactoring some other stories to include variables from the controls constants files, removed all uses of `defaultValue` to prep for the v7 release of storybook where it is depreciated - (Thank you @seialkali). 
Aligned our stories with some other optimizations for the controls where I was able to. 

I also improved some of the labelings of components so that the values we select will be more accurate to the ones you would use with the component in an application. For example, we will want to pass in "SystemIcon.ArrowRight" instead of "arrow-right" to a component and the control should reflect this now by utilizing "mapping" within the story args configuration. 

## Checklist for PR creator


packages/card/stories/Card.stories.tsx
```diff
@@ -1,20 +1,18 @@
...
 import { Story, Meta } from "@storybook/react";
-
-const sizes = ["xxs", "xs", "s", "m", "l", "xl", "xxl"];
+import { spacingSizeValues } from "../../storybookHelpers/controlConstants";
 
 export default {
   title: "Layout/Card",
   component: Card,
   subcomponents: { ButtonCard, LinkCard },
   argTypes: {
```
seialkali, line 10: I noticed there was a story for a Card with an `AspectRatio`, but the control for AspectRatio automatically defaults to an object. This seems to be the Storybook control type for arrays. I wasn't able to figure out how to get the control working so you could specify an aspect ratio. Do you know how it would be done, and if not should we disable the control?

packages/badge/stories/colorcodedbadge.stories.tsx
```diff
@@ -10,8 +10,11 @@ import {
...
   purple
 } from "../../design-tokens/build/js/designTokens";
-import { SystemIcons } from "../../icons/dist/system-icons-enum";
 import { ColorCodedBadgeProps } from "../components/ColorCodedBadge";
+import {
+  systemIconLabels,
+  systemIcons
+} from "../../storybookHelpers/controlConstants";
 
 const colors = {
```
seialkali, line 19: Should we change the this over to the color control type?

packages/storybookHelpers/controlConstants.ts
```diff
@@ -0,0 +1,105 @@
+import { Direction } from "../dropdownable/components/Dropdownable";
```
seialkali, line 1: Thanks for catching the file name misspelling and fixing it 🙈

packages/card/stories/Card.stories.tsx
nataliepina: Good point. Now that I know about the mapping I may be able to do something about this to make this control more usable. Otherwise, I may disable it, it's a prop that probably is not used often.

seialkali (approved):
Looks good! Thanks for taking the time to make all those tweaks.

PR #670, opened 2022-01-06
feat: add card header for images

nataliepina:
Closes D2IQ-82107

## Screenshots
Screen Shot 2022-01-06 at 1 43 43 PM

## Checklist for PR creator

### DIFF
--- a/packages/card/components/Card.tsx
+++ b/packages/card/components/Card.tsx
@@ -1,8 +1,14 @@
 import * as React from "react";
-import { cardBase } from "../style";
+import { cardBase, cardHeaderImage, headerHeight } from "../style";
 import { cx } from "@emotion/css";
 import { preserveAspectRatio, padding } from "../../shared/styles/styleUtils";
 import { SpaceSize } from "../../shared/styles/styleUtils/modifiers/modifierUtils";
+
+export enum HeaderSizes {
+  S = "s",
+  M = "m",
+  L = "l"
+}
 
 export interface CardProps extends React.HTMLProps<HTMLDivElement> {
   /**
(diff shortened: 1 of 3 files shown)


packages/card/components/Card.tsx
```diff
@@ -14,6 +14,13 @@ export interface CardProps extends React.HTMLProps<HTMLDivElement> {
    */
   paddingSize?: SpaceSize;
   children?: React.ReactNode | string;
+  /**
+   * The header will be set to the top section of the card optionally. Set a header image along with corresponding alt text.
+   */
+  header?: {
+    headerImg?: string;
```
clintonmedbery, line 21: ❓ - if the header is optional, should these be mandatory? Thinking out loud more than anything.

packages/card/style.ts
```diff
@@ -19,6 +20,16 @@ export const cardBase = css`
   }
 `;
 
+export const cardHeaderImage = css`
+  > img {
+    width: 100%;
+    height: 30%;
+    max-height: 100px;
```
clintonmedbery, line 27: ❓ - Would we want this to be more flexible in the future?

clintonmedbery (approved):
Looks great! I mentioned some things to think about, but no deal breakers.

packages/card/style.ts
nataliepina: This I was debating over as well. Do you think a user would prefer to be able to set the height of the image or have it taken care of for them?

packages/card/components/Card.tsx
```diff
@@ -14,6 +20,14 @@ export interface CardProps extends React.HTMLProps<HTMLDivElement> {
    */
   paddingSize?: SpaceSize;
   children?: React.ReactNode | string;
+  /**
+   * The card header will be set to the top section of the card optionally. Set a header image along with corresponding alt text. Size can be set to "s", "m", or "l" to increase the header height.
+   */
+  header?: {
+    headerImg: string;
+    headerImgAltText?: string;
```
mstruebing, line 28: Shouldn't we make this mandatory if the `headerImg` also is mandatory?

nataliepina: IMO no, sometimes an image might be more of a background or decorative image that doesn't require additional context for screen readers. When a screen reader comes across null alt text, it will completely skip over the image, without announcing its presence.

mstruebing: Thanks for clarifying, I wasn't sure how this will be used so I just assumed an image has a meaning. 👍

mstruebing (approved):

PR #672, opened 2022-01-11
Storybook Landing Page

nataliepina:
Changes Explained:
- To have the welcome page render before the other stories it was required to be added into main.js first. This is our first MDX file. 
- I discovered a fix required for the card header when an image with width and height was passed in. The previous style passed down a height of 100% to all child divs, which would cause the image to take over the entire card. This change affected related snapshots. 
- Welcome page has been broken down into components as it is easier to work with than to work with jsx inside of a .mdx file. 
- Updated storybook while working to resolve issues. 
- Previously we were not able to import static assets (PNGs, SVGs), I have configured the `png.d.ts` and `svg.d.ts` files to allow for this which correct related TS errors. 

## Testing
Start up Storybook, you should be taken to this new landing page.  🎉 

## Trade-offs

Jest changes were required to fix broken tests after the new png import implementation. Webpack and Jest together require this setup to understand processing static assets. 

## Screenshots
Screen Shot 2022-01-12 at 8 45 06 PM

Screen Shot 2022-01-12 at 8 45 25 PM

## Checklist for PR creator


packages/card/components/Card.tsx
```diff
@@ -70,7 +70,9 @@ class Card<P extends CardProps, S extends {}> extends React.PureComponent<
             <img src={header.headerImg} alt={header.headerImgAltText ?? ""} />
           </div>
         )}
-        <div className={padding("all", paddingSize)}>{children}</div>
+        <div className={cx(padding("all", paddingSize), cardContent)}>
```
vacas5, line 73: this may be a breaking change. Or perhaps we can call it a bug fix

packages/docs/components/WelcomeBanner.tsx
```diff
@@ -0,0 +1,35 @@
...
+            standards, and guidelines to use to create a unified user experience
+            at D2iQ.
+          </Text>
+        </TextWrapper>
+      </SpacingBox>
+    </BannerContainer>
+  );
+};
+
+export default WelcomeBanner;
```
vacas5, line 35: are these docs components not meant to be consumed by library users?

vacas5 (approved):
just some questions but no blockers

packages/card/components/Card.tsx
nataliepina: I put that bit under the first fix commit. I don't believe it should be a breaking change, as it's retaining the previous style of height 100% but making sure that it does not get applied to the card header div I added. I wasn't able to see this issue until passing in an image with a set height and width from placehold.it

packages/docs/components/WelcomeBanner.tsx
nataliepina: They should only be needed to export into UI Kit docs. There was an issue related to the PNG module importing that was only resolved by including the exports, as well as the inclusion in the packages > index.ts file. I am not sure of a workaround for this yet.

mstruebing (approved):
Very cool 🎉  I don't understand the existence of `__mocks__` but it looks good and works 👍

.storybook/webpack.config.js
```diff
@@ -11,5 +11,18 @@ module.exports = ({ config }) => {
   config.resolve.alias["core-js"] = path.dirname(require.resolve("core-js"));
   config.module.rules[0].use[0].options.sourceType = "unambiguous";
 
+  // Taken from stackoverflow
+  // https://stackoverflow.com/a/59639736/3058839
```
seialkali, line 15: Thanks for adding links for more info here and in the Jest config. That's really helpful!

seialkali (approved):
Looks great 😍

PR #676, opened 2022-01-20
chore: fixed versioning and classnames peer dep

nataliepina:
Updating the dependencies to be set as a fixed version for safety purposes. 
`classnames` is a peer dependency that needed to be added. 

## Checklist for PR creator


mstruebing (changes requested):
Good to have our direct dependencies pinned here as well but I think we do not want to pin the peer-dependencies to a specific version but rather want to have a version range like before so that we do not necessarily need multiple versions of react when using this component library.

> One piece of advice: peer dependency requirements, unlike those for regular dependencies, should be lenient. You should not lock your peer dependencies down to specific patch versions.

https://nodejs.org/es/blog/npm/peer-dependencies/

nataliepina:
> Good to have our direct dependencies pinned here as well but I think we do not want to pin the peer-dependencies to a specific version but rather want to have a version range like before so that we do not necessarily need multiple versions of react when using this component library.
> 
> > One piece of advice: peer dependency requirements, unlike those for regular dependencies, should be lenient. You should not lock your peer dependencies down to specific patch versions.
> 
> https://nodejs.org/es/blog/npm/peer-dependencies/

This is informative, thanks Max!

mstruebing (approved):
:

PR #680, opened 2022-01-31
fix: unknown prop warnings

seialkali:
Add fixes for components which are throwing the unknown-prop warning.
https://reactjs.org/warnings/unknown-prop.html

Closes D2IQ-83375

## Testing

To test this you will need to run: `npm run dist` and then replace the ui-kit dist folder within the ui-kit package in the frontend folder of kommander-ui with the newly created dist folder from ui-kit.

You can run the following command in kommander-ui to quickly replace the dist folder. Make sure to add in the path of your local ui-kit dist folder: 

`rm -r ./kommander/frontend/node_modules/@dcos/ui-kit/dist/ && cp -r PATH_OF_UI_KIT_DIST_FOLDER ./kommander/frontend/node_modules/@dcos/ui-kit/ `

Then run `npm run start`

Visit the following pages/sections and follow the instructions. Make sure you don't see the Unknown Prop Warning:

> Warning: React does not recognize the [PROP] prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase [PROP] instead. If you accidentally passed it from a parent component, remove it from the DOM element.

***Getting Started***
 * Click on Add Infrastructure Provider button

***Clusters***
 * Click Add Cluster button, select Attach Cluster, in the Attach Cluster form, when you click on either option for "No additional networking restrictions" or "Cluster has networking restrictions"

***Create Konvoy Cluster Form***
 * Clicking on one of the recommended labels under the Add Labels section

***Create Workspace Form***
 * Clicking on Show Advanced

***Projects Page***
 * Clicking on Create Project button

***Project Detail Page***
 * Clicking on Edit Project
 * Under Clusters tab, clicking on the 3 dots for a cluster within the table then clicking on Edit
 * Under Continuous Deployment tab, clicking the Add GitOps Source button
## Checklist for PR creator


packages/typeahead/components/Typeahead.tsx
```diff
@@ -181,7 +181,7 @@ class Typeahead extends React.PureComponent<TypeaheadProps, TypeaheadState> {
                         );
                       },
                       value: value === undefined ? inputValue : value,
-                      downshiftReset: other.reset,
+                      ...(downshiftReset && { downshiftReset: other.reset }),
```
vacas5, line 184: will this attempt to spread `false` if downshiftReset is undefined? Maybe you want `...(downshiftReset ? { downshiftReset: other.reset } : {})`?

Not sure, if you tested and it's okay, then 👍🏻

seialkali: This should be fine. The spread operator takes lower precedence than the `&&` operator. If the conditional returns false then the line will be reduced to `...(false)`, which shouldn't cause any issues because the spread operator ignores values without a property. [See this article for more info](https://medium.com/@slamflipstrom/conditional-object-properties-using-spread-in-javascript-714e0a12f496). 

However, I've just realised that this isn't going to work the way I need it to. It's a bit complicated and wordy to type out. I'm going to sleep on it and see if I can find another solution tomorrow.

packages/selectInput/components/SelectInput.tsx
```diff
@@ -99,6 +99,7 @@ class SelectInput extends React.PureComponent<
       showInputLabel,
       inputLabel,
       tooltipContent,
+      hintContent,
```
mstruebing, line 102: 💡  In line #116 we could replace 

```
        hintContent={this.props.hintContent}
```

with

```
        hintContent={thintContent}
```

because of this now.

packages/textInput/components/TextInput.tsx
```diff
@@ -134,6 +134,7 @@ export class TextInput<P extends TextInputProps> extends React.Component<P> {
       value,
       showInputLabel,
       appearance,
+      tooltipContent,
```
mstruebing, line 137: ❓ I'm not sure, but could there be a way to make sure to only pass valid HTML to the `<input>`-element?

Something like: 

```
<input {...rest.filter(prop => isValidNativeInputProp(prop) />
```

But I'm not sure if something like this exists or is possible, but this would feel a lot better than to destructure all props we do not want to pass.

mstruebing (commented):
Finally, we get to fix these, thank you! 🥇

vacas5 (approved):

packages/textInput/components/TextInput.tsx
mstruebing: We talked about it and decided to not do it for now. If that pops up a lot again, we can look into if we want to implement something like this.

nataliepina (approved):
Awesome work finding and fixing this bug 🐛

PR #681, opened 2022-02-01
feat: allow text wrapping for tables

nataliepina:
Closes D2IQ-84285

If you set `contentNoWrap` false, the content of a column will allow for text wrapping instead of full truncation. 

Text will still "truncate" with the ellipsis after wrapping if it extends into overflow, as we wouldn't want something long (that cannot break) such as an email address to overlap other table cells or become cut off. We still want the user to understand that there is more text there with the ellipsis in that case.

## Testing

Compare the Storybook story for Content Wrapping and ensure previous behavior on other Table stories.

## Screenshots
Screen Shot 2022-02-01 at 3 22 59 PM

## Checklist for PR creator

### DIFF
--- a/packages/tablev2/Table.tsx
+++ b/packages/tablev2/Table.tsx
@@ -45,6 +45,11 @@ export type Column<A, B = string> = {
    * Text alignment for cells in the column
    */
   textAlign?: React.CSSProperties["textAlign"];
+
+  /**
+   * The contentNoWrap prop sets the `whitespace` to `nowrap` which prevents content from wrapping within the cell. The default is set to true to preserve truncation backwards compatibility.
+   */
+  contentNoWrap?: boolean;
 };
 
 // we internally use a different type than we expose in the API.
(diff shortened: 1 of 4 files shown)


packages/tablev2/Table.tsx
```diff
@@ -272,20 +277,23 @@ const Row = React.memo(function Row<A>({
...
+            className={cx(style.cell(textAlign), {
+              [style.makeSpaceForSortIndicator]:
+                Boolean(sorter) && textAlign === "right",
+              [style.rowScrollShadow]: showScrollShadow,
+              [style.nowrap]: !contentWrap
+            })}
+            role="gridcell"
+          >
+            {render(el)}
+          </div>
```
vacas5, line 294: hmm... I don't love how this isn't semantic HTML. oh well.

packages/tablev2/Table.tsx
```diff
@@ -45,6 +45,11 @@ export type Column<A, B = string> = {
    * Text alignment for cells in the column
    */
   textAlign?: React.CSSProperties["textAlign"];
+
+  /**
+   * Text wrapping allows column content to wrap around instead of becoming cut off and truncated. The default is set to false to preserve truncation with backwards compatibility.
+   */
+  contentWrap?: boolean;
```
vacas5, line 52: because the default table behavior is wrap, and the existence of a boolean actually enables `nowrap`, you might consider inverting this prop. Making it `contentNoWrap` and defaulting it to `true` for backwards compatibility. either way, glad to have it.

vacas5 (approved):

seialkali (approved):
Looks great! Good work 🎉

PR #682, opened 2022-02-03
fix: unknown prop warning - openInNewTab

seialkali:
Destructure the `openInNewTab` prop in `ButtonBase.tsx` so that it doesn't get passed down to the `button` DOM element which doesn't have that attribute.

## Testing

There is some discussion about destructuring props vs filtering out invalid props. This is a quick fix until we make a final decision on how we want to handle these situations.

## Checklist for PR creator


mstruebing (approved):
👍

PR #701, opened 2022-04-04
feat: fullscreenview banner component

nataliepina:
This change should allow for banners on fullscreen view pages and modals. 
Additionally, the style changes may help to resolve the double scrollbar issue discovered in the UI. 

## Screenshots
Screen Shot 2022-04-04 at 2 56 52 PM

## Checklist for PR creator

### DIFF
--- a/packages/fullscreenView/components/FullscreenView.tsx
+++ b/packages/fullscreenView/components/FullscreenView.tsx
@@ -20,6 +20,8 @@ interface FullscreenViewProps {
   isContentFlush?: boolean;
   /** Custom header content component. ⚠️Use rarely and with caution⚠️ */
   headerComponent?: React.ReactNode;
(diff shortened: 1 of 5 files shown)


mstruebing (approved):
I think one problem with this would be that we would need to pass the `bannerComponent` on every single fullscreen page.

An idea could be that we would create a custom component in `kommander-ui` based on this  change and pass the `bannerComponent` to the `FullscreenView` based on a context.

Something like this:
 ```js
<Context>
// ...some route magic
  <CustomFullscreenView  title="ABC" />
</Context>


function CustomFullscreenView {
  {bannerComponent} = useContext(BannerContext)

  return (
     <FullscreenView bannerCompoent={bannerComponent} {...props}>{children}</FullscreenView>
   )
}
```

PR #542, opened 2020-08-25
D2IQ-70983: constrain dimensions of dropdown element passed to Dropdownable to the dimensions of the viewport

mperrotti:
## Testing
Try opening a dropdown menu powered by dropdownable (Typeahead, DropdownMenu) that has too many items to fit in the viewport.
The dropdown menu should be scrollable

Bonus: test this in Kommander using the workspace selection dropdown populated with a bunch of workspaces

## Screenshots
https://share.getcloudapp.com/eDuBl7BJ

## Checklist for PR creator

- [ ] If any new components were added, there are exported from `packages/index.ts`
- [x] If this PR is associated with a JIRA, it is mentioned in commit message footer ("Closes …")
- [ ] If this PR contains breaking changes, is stated in commit message body ("BREAKING CHANGE: …")
- [x] Info for applicable sections above is provided

PR #703, opened 2022-04-11
D2IQ-87203: react-popper console errors

mstruebing:
## Testing

You would need to build rebuild `ui-kit` and manually copy the `dist` folder to `kommander`, I do it like this from `ui-kit`:

```bash
npm run dist && \
 rm -rf ~/projects/work/kommander-ui/kommander/frontend/node_modules/@dcos/ui-kit/dist/packages && \
 cp -r dist/packages ~/projects/work/kommander-ui/kommander/frontend/node_modules/@dcos/ui-kit/dist/
```

Make sure to update the paths to yours.

http://localhost:1234/workspace/kommander-workspace/access-control/roles/new and select a resource type.

## Trade-offs

I can not see any difference without the `getFlipModifier` applied, anybody knows in which situation we need it?

## Checklist for PR creator

### DIFF
--- a/packages/dropdownable/components/Dropdownable.tsx
+++ b/packages/dropdownable/components/Dropdownable.tsx
@@ -1,6 +1,6 @@
 import React, { CSSProperties } from "react";
 import { usePopper } from "react-popper";
-import { ModifierPhases } from "@popperjs/core";
+import { ModifierPhases, Modifier } from "@popperjs/core";
 import maxSize from "popper-max-size-modifier";
 import Overlay from "../../shared/components/Overlay";
 import DropdownContents from "./DropdownContents";
@@ -45,7 +45,7 @@ const getPreferredDirection = (
 
 const getFlipModifier = (preferredDirections?: Direction | Direction[]) => {
   if (!preferredDirections || !preferredDirections.length) {
-    return {};
+    return null;
   }
   if (preferredDirections.length > 1) {
     return {
@@ -55,7 +55,7 @@ const getFlipModifier = (preferredDirections?: Direction | Direction[]) => {
       }
     };
   }
-  return {};
+  return null;
 };
 
 const applyMaxSize = {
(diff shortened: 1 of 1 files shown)


seialkali:
I've done some digging and I've worked out what the `getFlipModifier` function is for. [`Flip` is a built-in Popper.js modifier](https://popper.js.org/docs/v2/modifiers/flip/) which `can change the placement of a popper when it's scheduled to overflow a given boundary`. Popper.js also allows you to add custom modifiers or to override the options of a built-in modifier. What `getFlipModifier` is doing is overriding the default options for the Flip modifier, to allow for the directions passed in via the `preferredDirections` prop to be used as the `fallbackPlacements` instead. This basically means that if you set a `preferredDirections` prop, it's not going to work for the situations where Flip comes into effect.

vacas5:
Is there an opportunity for us to remove a dependency here? 

Secondly

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -90,7 +75,7 @@ const Dropdownable: React.FC<DropdownableProps> = ({
     React.useState<HTMLDivElement | null>(null);
   const { styles, attributes } = usePopper(referenceElement, popperElement, {
     placement: getPreferredDirection(preferredDirections),
-    modifiers: [maxSize, applyMaxSize, getFlipModifier(preferredDirections)]
```
vacas5, line 93: I think this is just deprecating functionality which means it would be a BREAKING CHANGE.

nataliepina:
Much of the changes related to what is being removed can be traced back to here - https://github.com/dcos-labs/ui-kit/pull/542/files

packages/dropdownable/components/Dropdownable.tsx
mstruebing: I fixed it in another way now.

nataliepina (approved):

packages/dropdownable/components/Dropdownable.tsx
```diff
@@ -90,7 +90,12 @@ const Dropdownable: React.FC<DropdownableProps> = ({
...
+    modifiers: [
+      maxSize,
+      applyMaxSize,
+      getFlipModifier(preferredDirections)
+      // we need valid modifiers here otherwise we get a bunch of console errors
+    ].filter(Boolean) as Array<Partial<Modifier<string, object>>>
```
seialkali, line 98: What's this filter for?

seialkali (approved):
Nice work figuring this one out! 🎉

PR #704, opened 2022-04-14
Removing Deprecated SFC References

nataliepina:
SFC is deprecated. If we want to update to React 18 we will need to first address this. I wanted to dedicate a PR to focus solely on the transition away from Stateless Functional Component.

## Checklist for PR creator

### DIFF
--- a/packages/dropdownMenu/components/DropdownMenu.tsx
+++ b/packages/dropdownMenu/components/DropdownMenu.tsx
@@ -55,7 +55,7 @@ export interface DropdownMenuProps {
   disablePortal?: boolean;
 }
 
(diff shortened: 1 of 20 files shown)


vacas5 (commented):
didn't we also decide that we don't want to use the type React.FC? Maybe we should just swap all of these out for no type?

nataliepina:
> didn't we also decide that we don't want to use the type React.FC? Maybe we should just swap all of these out for no type?

I'm going to update these for the ones that we can be sure will not cause breaking changes or require additional components to be brought into this update. I will address the other FC references when creating future component updates such as adding the id and data-cy to other components.

vacas5 (approved):

clintonmedbery (approved):
Nice!

PR #707, opened 2022-04-22
fix: change type for children prop in the Accordian component

seialkali:
The component was getting a Typescript error when used with only a single ReactNode passed in as a child, so
the type needed to be updated to allow for for both a single ReactNode and an array of ReactNodes

## Checklist for PR creator


mstruebing (approved):
Great find 👍

nataliepina (approved):

PR #708, opened 2022-04-28
Fix for tabIndex values with updated react-focus-lock versioning

nataliepina:
## Screenshots
With this update, the div with tabIndex="1" is removed from the snapshot. This is a good indicator that it will be our fix in the UI based on the a11y violation that tabIndex values should not be greater than 0. 
Screen Shot 2022-04-27 at 12 22 12 PM

## Checklist for PR creator


vacas5 (approved):
love a dependency upgrade that removes unnecessary dom elements

PR #712, opened 2022-05-03
fix: badgebutton children and data-cy props

nataliepina:
This is a fix for the `children` props on `BadgeButton` which should allow a string or any elements we pass in. This can be covered with the use of `React.ReactNode` which is consistent with other `children` props in ui-kit. Took the opportunity here to allow for a customizable data-cy prop as well. 

## Checklist for PR creator


mstruebing (approved):
🎉

PR #713, opened 2022-05-05
ESLint Rule Review and Cleanup

nataliepina:
We wanted to timebox these changes to ~3 days.  I was able to remove many unnecessary ESLint rules, especially the ones that were there but turned off.

I've updated the eslint-related dependencies here and added some ignore files for eslint and prettier. 

## Testing

- npm run test
- npm run lint 

## Trade-offs

There are some rules I am unable to remove due to the vast amount of changes that would be required, for example
"@typescript-eslint/no-explicit-any": "off".

There were a few rules that I attempted to remove that broke a lot of tests surrounding types with the use of 'this'. If we are able to modernize and refactor class components we can revisit these rules for a second round of rule cleanup which should resolve a lot of what I saw. 

## Checklist for PR creator


vacas5 (approved):
Seems like positive improvements

TattdCodeMonkey (approved):

PR #714, opened 2022-05-06
chore: update faker dependency

clintonmedbery:
Closes D2IQ-88823

## Testing

Go to tables and pagination in storybook. Make sure there is fake data there.

## Trade-offs

We were using something called createCard. It would create a "user" with name, email etc. This is deprecated so the name might not be similar to the email in our table. For example with 5 the name might be Bob Ross and the email would be bob.ross@gmail.com. This does not seem to exist in 6 but frankly it is fake data so who cares.

https://github.com/faker-js/faker/issues/528

## Dependencies

Faker 6
https://github.com/faker-js/faker

## Screenshots

image

## Checklist for PR creator


nataliepina (approved):
Sick 🤙  Thanks for updating this!

vacas5 (approved):
yeah for instance my name is Russ but my email is `straightedgehardcore4lyfe@hotmail.com` so maybe this is even better fake data

PR #715, opened 2022-05-10
Theming defaults for AppChrome

nataliepina:
Currently, in the UI we are using `ThemeProvider` in 2 locations to override variables for the AppChrome sidebar and header. We can remove these if we create updates to the AppChrome to match these values. This could even help to remove the `@emotion/react ` depedency altogether, as it's only used for `ThemeProvider`. 
The previous defaults here seemed to match the old DC/OS theming with the dark header and sidebar. This is a good opportunity to align with the current design of our product using ui-kit.

We are able to change the fallback design token for `bgAppHeader` as it is only used by AppChrome and we would be overriding this value anyway in the UI. There was some auto-formatting on this fallbacks.json file also. 

I've refactored the stories to better match the changes, with some extra cleanup. 

All components have been refactored to move away from React.FC and the `children` props are explicitly set. It was not necessary for `AppChrome` to be a class component so that was converted to a memoized FC.

Added a TODO and I will revisit the two values that are set to hex codes that we do not have in our Design System. I have brought this up with product design and we will address this in the design token audit with our transition to Figma. 

Overall, these changes should help us to tidy up some outdated and unnecessary code in the UI. 

## Testing

Startup Storybook and view the AppChrome stories. 

## Checklist for PR creator


packages/design-tokens/properties/theme/fallbacks.json
```diff
@@ -1,84 +1,85 @@
...
+      "bgScrim": {
+        "base": { "value": "rgba(0, 0, 0, 0.2)" }
+      },
+      "bgSelected": {
+        "base": { "value": "#7D58FF" },
+        "inverted": { "value": "#7D58FF" }
+      },
+      "bgAppHeader": {
+        "base": { "value": "#FFFFFF" },
+        "inverted": { "value": "#1B2029" }
```
nataliepina, line 53: This is the only value that I have changed. I've matched what it was being set to in the UI. I also added an inverse variable, the inverse of our primary brand color (white) is #1B2029.

vacas5:
FYI this is probably a feature: and is a BREAKING CHANGE

vacas5 (commented):
I have some ideas how to make it a non-breaking change, but I am not sure how fallbacks are used. Changing a design token would be a breaking change either way.

packages/appChrome/components/AppChrome.tsx
```diff
@@ -9,54 +9,49 @@ import {
...
-    );
-  }
-}
+      </div>
+    </ThemeProvider>
+  );
+};
 
-export default AppChrome;
+export default React.memo(AppChrome);
```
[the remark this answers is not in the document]
nataliepina: Is it a breaking change though? I asked myself this. I am changing a fallback value but these values seem to reference DC/OS color variables, and that value is only used for AppChrome, so the fallback value isn't even being used at all with the override in the UI. 
The style.ts file handles this sort of logic, for example, this is how it handles the background color:
`  const bgColor =
    theme.headerBackgroundColor || getCSSVarValue(themeBgAppHeader);`

packages/appChrome/components/Sidebar.tsx
```diff
@@ -43,12 +41,7 @@ const StyledSidebar: React.FC<{ isOpen?: boolean }> = ({
   );
 };
 
-const Sidebar: React.FC<SidebarProps> = ({
-  children,
-  isOpen,
-  onClose,
-  onOpen
-}) => {
+const Sidebar = ({ children, isOpen, onClose, onOpen }: SidebarProps) => {
```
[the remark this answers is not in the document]
nataliepina: This will require the ability to also set a` hoverColor` and an `activeColor`, I've included all 3 of these as props here.

packages/appChrome/components/AppChrome.tsx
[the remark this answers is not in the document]
nataliepina: I will leave the theme to take care of this as we already have the specific theming variable. There will be more design token changes to come in the near future with the audit.

packages/appChrome/stories/AppChrome.stories.tsx
```diff
@@ -19,7 +19,7 @@ export default {
   component: AppChrome
 } as Meta;
```
seialkali, line 20: Do we want Storybook users to be able to use the controls for the AppChrome props? If not, should we disable them? At the minute they're not very intuitive, and also, clicking on the `Set object` button under `App Chrome Bare` gives an error.

packages/appChrome/stories/AppChrome.stories.tsx
```diff
@@ -84,7 +91,35 @@ export const SidebarWithItemsWithOnClick = args => (
...
+  </Sidebar>
+);
+
+export const SidebarWithCustomSidebarColors = args => (
+  <Sidebar
+    isOpen={true}
+    bgColor={"#000"}
+    hoverColor={"#988ee4"}
+    activeColor={"#8375eb"}
+    {...args}
```
seialkali, line 121: Should we try to set up this story so that the user can change these colors with the Storybook controls?

packages/appChrome/components/AppChrome.tsx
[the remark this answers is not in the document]
seialkali: With the future design token changes will these also be breaking changes? If so, would it make sense to do them all at once?

packages/appChrome/stories/AppChrome.stories.tsx
nataliepina: Yeah, agreed that it's not very helpful when you have to pass in an object of some sort. It's also difficult to have controls for nested aspects which is why I haven't included the color picker for the section below. I'm going to disable controls for this story.

packages/appChrome/components/AppChrome.tsx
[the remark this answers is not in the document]
nataliepina: This one directly relates to the AppChrome component so I would like to change it here ideally. I think this will help with the issues we've seen trying to use npm link. I saw these issues when updating Emotion in the UI as well. This color is only used with this component.

vacas5 (approved):

packages/appChrome/components/AppChrome.tsx
```diff
@@ -8,55 +8,52 @@ import {
...
```
clintonmedbery, line 59: I feel like memoizing the whole app might have some repercussions. Why do we need to do this? It might be worth running the React DevTools Profiler with and without memo and seeing the difference in performance.

clintonmedbery (commented):
This looks good but I wonder if React.memo on the whole app is the right call or not.

packages/appChrome/components/AppChrome.tsx
nataliepina: The reason I've done this is that when I've refactored another component that was a `PureComponent `it was suggested that I should memoize it to retain that functionality.

nataliepina: Removing this after further review, since we are not doing any expensive calculations with this component there will not be a big difference in performance.

PR #716, opened 2022-05-12
chore: added dev watch to ui-kit

clintonmedbery:
## Testing

Run `npm i`.
Run `npm run dev`.
Make some changes and see if they appear.

## Dependencies

https://github.com/gilamran/tsc-watch

## Checklist for PR creator

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "pretest": "npm run build:design-tokens && npm run build:svg",
     "predist": "npm run clean && npm run build:design-tokens && npm run build:svg",
     "dist": "tsc -p tsconfig.dist.json",
+    "dist:watch": "tsc --watch -p tsconfig.json",
     "lint": "eslint --ext .tsx,.ts packages/",
     "format": "prettier --write './**/*.{js,ts,tsx,html,md}'",
     "setup:git-hooks": "./scripts/pre-install",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -24,5 +24,15 @@
     "./typings/*",
     "./.storybook/docs/components/*.tsx"
   ],
+  "watchOptions": {
+    // Use native file system events for files and directories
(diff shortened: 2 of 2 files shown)


nataliepina (approved):
Lovely

package.json
```diff
@@ -34,7 +34,8 @@
     "storybook": "start-storybook -p 6006",
     "test": "npx jest",
     "test:integration": "./scripts/test-integration",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "dev": "tsc-watch -p tsconfig.dist.json"
```
vacas5, line 38: this is very similar to `dist`. Should we rname `dist:watch`?

vacas5:
this is actually only tangentially related to the `npm link` issue, so, probably no need to reference the issue here.

package.json
mstruebing: I would match it with the build script name.
So in this case I would agree with Russ and name it `dist:watch` - that way it's easier to remember.

We should keep in mind that if we change svgs or update our design tokens we need to execute some additional steps (`npm run build:design-tokens && npm run build:SVG`) but I think that is okay and we should not run it everytime a watch event appears as this would take much longer to build.

clintonmedbery:
Before we merge, @mstruebing was telling me about https://www.typescriptlang.org/docs/handbook/configuring-watch.html and that sounds like it might be better. I am going to take a look.

tsconfig.json
```diff
@@ -24,5 +24,15 @@
     "./typings/*",
     "./.storybook/docs/components/*.tsx"
   ],
+  "watchOptions": {
```
clintonmedbery, line 27: This are some of the "boilerplate" commands, but they seemed to make sense to use them.

vacas5 (approved):

PR #717, opened 2022-05-16
fix: deprecate use of downshift

vacas5:
In an attempted upgrade of downshift I noticed a bug in their types.
Looking into a possible contribution, I then noticed that they have not
merged a PR in about a year. I am not sure if the project is currently
maintained, as all PRs were failing CI.

Instead I am attempting to remove downshift as a dependency altogether,
which should give users more granular control of accessibility concerns.

BREAKING CHANGE:
I will no longer return all of the Downshift props from the onSelect
callback in DropdownMenu.

## Checklist for PR creator


nataliepina (changes requested):
Screen Shot 2022-05-17 at 8 39 35 AM
Comparing the changes and it looks like there are some missing flex props that should be passed down to a `div` that is a child of `PopoverListItem` in the `DropdownMenu`. These look to have come from along with Downshift. 

Missing flex props can be seen here:
Screen Shot 2022-05-17 at 8 58 11 AM

tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "moduleResolution": "node",
     "skipLibCheck": true,
     "tsBuildInfoFile": ".tsbuildinfo",
-    "lib": ["es6", "dom", "es2017"]
```
seialkali, line 20: Nice. Just out of curiosity, did adding this fix any issues we were having with TS?

packages/popover/components/PopoverListItemIcon.tsx
```diff
@@ -8,7 +8,7 @@ interface PopoverListItemProps extends IconProps {
   position?: "start" | "end";
 }
 
-const PopoverListItemIcon: React.SFC<PopoverListItemProps> = props => (
+const PopoverListItemIcon = (props: PopoverListItemProps) => (
```
nataliepina, line 11: 👏

vacas5:
I will more than likely close this, but I'll try and grab some of the nicer parts into another PR

vacas5:
For now, I've gotten the upgrade to work. We can possibly revisit this if we find that downshift is indeed abandoned

PR #721, opened 2022-05-20
fix: reintroduce dropped typeahead stories

vacas5:
During a refactor a few of these examples got dropped and stopped
functioning. Adding them back in to help with regression testing of the
downshift upgrade.

Additionally I discovered that the complex list items example doesn't
work even when adding back in args to the default. This apparently has
not worked since some time in 2019, as of the following commit https://github.com/dcos-labs/ui-kit/commit/1c01ea9bbbaf49047f5098fa80f0ab694f02d8fa

Thus, I removed it completely rather than trying to untangle it. We
evidently have not needed it.

## Checklist for PR creator


packages/typeahead/stories/helpers/MultiselectTypeahead.tsx
```diff
@@ -1,69 +1,46 @@
...
-class MultiselectTypeahead extends React.PureComponent<
-  MultiselectTypeaheadProps,
-  MultiselectTypeaheadState
-> {
-  constructor(props) {
-    super(props);
-
-    this.selectHandler = this.selectHandler.bind(this);
+const MultiSelectTypeahead = ({ children }: MultiselectTypeaheadProps) => {
+  const [selectedItems, setSelectedItems] = React.useState<string[]>([]);
```
nataliepina, line 20: Great work updating this component too.

nataliepina (approved):
Great find and clean up here. Everything looks good to me.

PR #722, opened 2022-05-20
chore: upgrade downshift

vacas5:
not sure exactly how this test was meant to work prior to this but it
seems like the user behavior is what we want, it just skips over the
disabled item and allows you to select the first available.

Otherwise some type updates to get it to work. Could use some visual
regression testing on these components.

Closes D2IQ-88824

## Checklist for PR creator

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "chartist": "0.11.0",
     "classnames": "2.3.1",
     "copy-to-clipboard": "3.3.1",
-    "downshift": "3.2.10",
+    "downshift": "6.1.7",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "chartist": "0.11.0",
     "classnames": "2.3.1",
     "copy-to-clipboard": "3.3.1",
-    "downshift": "3.2.10",
+    "downshift": "6.1.7",
```
nataliepina, line 55: 3 major versions 😳

PR #723, opened 2022-05-22
refactor: dropdown components

vacas5:
During the downshift drama I refactored this component so that the child
components did more than simply just pass along children. Thought this
would make it easier to discover where particular styles are coming
from, etc.

## Checklist for PR creator

### DIFF
--- a/packages/dropdownMenu/components/DropdownMenu.tsx
+++ b/packages/dropdownMenu/components/DropdownMenu.tsx
@@ -1,10 +1,8 @@
 import * as React from "react";
-import { cx } from "@emotion/css";
 import Downshift, { ControllerStateAndHelpers } from "downshift";
 import { Dropdownable } from "../../dropdownable";
-import { border, buttonReset, display } from "../../shared/styles/styleUtils";
+import { buttonReset, display } from "../../shared/styles/styleUtils";
 import PopoverBox from "../../popover/components/PopoverBox";
-import PopoverListItem from "../../popover/components/PopoverListItem";
 import { DropdownSectionProps } from "./DropdownSection";
 import { DropdownMenuItemProps } from "./DropdownMenuItem";
 import { Direction } from "../../dropdownable/components/Dropdownable";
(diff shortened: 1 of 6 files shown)


vacas5:
additionally, it would make these components possible to use in isolation, if one wanted to

vacas5:
this may not be backwards compatible if users are using these subcomponents outside the context of a dropdown menu 😢

packages/dropdownMenu/components/DropdownSection.tsx
```diff
@@ -1,12 +1,26 @@
 import * as React from "react";
+import { cx } from "@emotion/css";
 import { DropdownMenuItemProps } from "./DropdownMenuItem";
+import { border } from "../../shared/styles/styleUtils";
 
 export interface DropdownSectionProps {
+  sectionIndex?: number;
```
mstruebing, line 7: ❓ Can we make this required?
Otherwise, it will become `key="dropdown-undefined"` very easily.

packages/dropdownMenu/components/DropdownMenu.tsx
```diff
@@ -134,52 +91,70 @@ const DropdownMenu = (props: DropdownMenuProps) => {
...
+                          })
+                        });
+                      })
+                    });
+                  })}
+                </PopoverBox>
+              }
+              disablePortal={disablePortal}
+            >
+              {React.isValidElement(trigger) ? (
```
mstruebing, line 143: I've never knew this function exists 🤓

packages/dropdownMenu/components/DropdownMenuItem.tsx
```diff
@@ -9,12 +10,49 @@ export interface DropdownMenuItemProps {
...
+  ...rest
+}: DropdownMenuItemProps) => {
+  return (
+    <PopoverListItem
+      listLength={listLength || 0}
+      isActive={isActive}
+      appearance={appearance}
+      index={index || 0}
+      disabled={disabled}
+      {...rest}
```
mstruebing, line 51: 💡 If we would set default arguments for `listLength` and `index` we could do:
`<PopoverListItem ...rest listlength={listlength} index={index}  />`

We would not need to spread them, except for the children as we do not do anything with these props.

mstruebing (commented):
💬 This nested `React.cloneElement` looks a bit scary but is probably necessary.

packages/dropdownMenu/components/DropdownSection.tsx
vacas5: not sure we can require it and not create TS errors all over our app based on how it is used now. I don't love the original implementation. Let me see if I can figure out a way to explain it. Essentially, the way this was written, DropdownMenu would extend the DropdownSection and DropdownMenuItem components with additional styling, but only if they were used as children of `DropdownMenu`. As such, my changes here, I realize now, are not backwards compatible.

vacas5: wrt to the `key` prop though, thanks for pointing this out because now I see it is not necessary at all

packages/dropdownMenu/components/DropdownMenuItem.tsx
mstruebing: But not in this component as far as I can see?

This component could look like this:


```
const DropdownMenuItem = ({
  listLength = 0,
  index = 0,
  children,
  ...rest
}: DropdownMenuItemProps) => {
  return (
    <PopoverListItem
      listLength={listLength}
      index={index}
      {...rest}
    >
      <>{children}</>
    </PopoverListItem>
  );
};
```

packages/dropdownMenu/components/DropdownSection.tsx
mstruebing: > wrt to the key prop though, thanks for pointing this out because now I see it is not necessary at all

I actually wanted to suggest that but I thought I may not know enough how that would work together with these clones - great that it can be removed.

packages/dropdownMenu/components/DropdownMenuItem.tsx
```diff
@@ -9,12 +10,49 @@ export interface DropdownMenuItemProps {
...
+   */
+  index?: number;
+  /**
+   * If the current item is in use
+   */
+  isActive?: boolean;
+  /**
+   * Children for PopoverListItem
+   */
   children: React.ReactNode;
```
nataliepina, line 32: ❓  Do you think it would be helpful to add a configurable `data-cy` here? It looks like there is only one set for the Popover items.

vacas5:
I need to add a BREAKING CHANGE here and maybe make it part of a release branch

vacas5:
I stand by some of this but it's a bit stale now.

PR #724, opened 2022-05-23
chore: removed max size deprecated dependency

clintonmedbery:
This feature was not working and I am not sure replacing it is worth the effort.

Expected behaviour:
When you have a pop up, it is not covered up by the bottom of the window. It is supposed to fit to the bottom of the screen.

So this feature does not work. The problem is usePopper raises the div outside of its parent div to the top of the DOM. As such there is no parent div so max-width and max-height do not work. They have another library called useFloating, but that _outputs_ refs instead of taking them, so it would take some refactoring. I say we just rip this out for two reason.

1. This was never working so no one would miss it.
2. There is already some behavior for when this occurs. The pop up flips to the other side of the parent element.

If we wanted to revisit this, I say we make a new story saying replace usePopper with the floating-ui library.
## Testing

None, it doesn't work.

## Trade-offs

See above. 

## Dependencies

The deprecated `popper-max-size-modifier` has been removed.

## Checklist for PR creator

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "downshift": "6.1.7",
         "eslint-plugin-prettier": "4.0.0",
         "memoize-one": "6.0.0",
-        "popper-max-size-modifier": "0.2.0",
         "react-chartist": "0.14.4",
         "react-click-outside": "3.0.1",
         "react-delegate-component": "1.0.0",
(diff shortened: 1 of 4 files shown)


vacas5:
@clintonmedbery can you use a git bisect to determine when this component stopped working? If it's been a couple years, I think we should deprecate it.

nataliepina:
You can fix the snapshots with `npm test -- -u`

clintonmedbery:
I went back to the original commit and couldn't get this to work properly. Paired with @nataliepina on that too. We also added functionality to flip the popover so maybe no one noticed?

vacas5:
if it didn't work in the OG commit then I think we should just remove the entire component, no?

clintonmedbery:
The component works, it's just the maxSize functions, which handles the popover size in certain situations as when it leaves the viewport.

mstruebing (approved):
✔️

seialkali (approved):
Good find!

nataliepina (approved):
There is a typo in your title/commit message if you want to update that from maz to max. Otherwise looking good, I've looked at this with you again and I think we're fine to remove this.

PR #725, opened 2022-05-24
Container component updates with custom data-cy and className props

nataliepina:
Overall some clean-up and new props for the Container component. 

We should be able to pass in a custom `data-cy` attribute to the Container component. We should also be able to extend the container styles and customize them with emotion. I noticed this through attempting to create a styled(Container) with Emotion, there are errors as classNames is not currently an available prop on this component. Having these new props also made the storybook controls usable.

This should ensure backward compatibility with the fallback values, there were no snapshot updates as well.

## Checklist for PR creator


packages/styleUtils/layout/container/components/Container.tsx
```diff
@@ -1,12 +1,19 @@
 import * as React from "react";
+import { cx } from "@emotion/css";
 import { container } from "../style";
 
 export interface ContainerProps {
   children: React.ReactNode;
+  ["data-cy"]?: string;
```
seialkali, line 7: What is this syntax called? I've not seen this before and can't seem to find anything about it in Google.

nataliepina: I'm not sure what the term is for this, I had been following the previous conventions in the ui-kit for data-cy props as string values to keep it consistent. I was under the assumption it was for TS to handle the option string prop with value but I've tested now it and it works regardless. Might have been an issue that was resolved along the way with TS.

nataliepina: I'm going to update it here to remove the brackets and I'll have to address the consistency separately. Thanks for pointing this out.

vacas5: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names

vacas5: oh wait, this is TS we're looking at

mstruebing (approved):
👍

PR #726, opened 2022-05-27
feat: add gap property to flex component

seialkali:
Adding the gap property to the Flex component so the gutter size for columns and rows can be set with a single value.

Closes D2IQ-90211

## Trade-offs

I've added the style for `gap` just below where `column-gap` and `row-gap` are set so that if `gap` is set, it will override the other values. I'm not sure if this is the best solution, but it was certainly the easiest.

## Checklist for PR creator

### DIFF
--- a/packages/messagePanel/tests/__snapshots__/messagePanel.test.tsx.snap
+++ b/packages/messagePanel/tests/__snapshots__/messagePanel.test.tsx.snap
@@ -117,7 +117,7 @@ exports[`MessagePanel renders with primary and secondary actions 1`] = `
         Body content
       </div>
       <div
-        class="css-bhf0my"
+        class="css-e35dnx"
         data-cy="flex"
       >
         <div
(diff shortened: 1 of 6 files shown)


mstruebing (commented):
❓ Is it possible to change the implementation so that they override the other way around?
I think if `row-gap` or `column-gap` is set it should override `gap` as they are more specific.

seialkali:
> ❓ Is it possible to change the implementation so that they override the other way around? I think if `row-gap` or `column-gap` is set it should override `gap` as they are more specific.

@mstruebing the reason I've done it this way is because you can't set `row-gap` or `row-column` separately, you can only set one or the other using both the `gutterSize` and `direction` props, so I thought it would be better to have `gap` be the override because if you've set `gap` then your intention is probably to set the gap for both column and row.

mstruebing (approved):

nataliepina (approved):

PR #727, opened 2022-05-31
Color prop for SectionLoadingIndicator

nataliepina:
Ability to set a custom color for the SectionLoadingIndicator.
Loading indicators story improvements included.

These changes have backward compatibility, the fallback will still be the gray color, and there will not be a breaking change caused by this addition.

## Trade-offs
I don't know if it's just on my end, but the storybook color picker control is having a slow-to-respond behavior.

## Screenshots
Initial Screenshots:
May-31-2022 14-00-11
Screen Shot 2022-05-31 at 2 33 19 PM
Screen Shot 2022-05-31 at 2 33 08 PM

Updated Screenshot:
Screen Shot 2022-06-01 at 9 33 16 AM

## Checklist for PR creator


packages/loadingIndicator/components/SectionLoadingIndicator.tsx
```diff
@@ -1,13 +1,23 @@
...
+  color?: string;
+}
+
+const SectionLoadingIndicator = ({ color }: SectionLoadingIndicatorProps) => {
+  return (
+    <div className={sectionLoadingWrapper}>
+      <div
+        data-cy="sectionLoadingIndicator"
+        className={css(sectionLoadingIndicator, {
+          backgroundColor: color
```
vacas5, line 16: I think we should default this to whatever it was before, or perhaps only set this if color is defined so it works exactly as before if not sent

packages/loadingIndicator/stories/loadingIndicator.stories.tsx
```diff
@@ -25,13 +33,13 @@ export const Template: Story = args => (
 
 export const _SectionLoadingIndicator = Template.bind({});
 
-export const _InlineLoadingIndicator = args => (
+export const _InlineLoadingIndicator = () => (
```
vacas5, line 36: can I ask why this change?

vacas5 (commented):
🤔

seialkali (approved):
Looks good! The only thing I noticed is that the color control in the storybook was really laggy when I tried to change the color. I thought the page crashed, but it was only frozen temporarily. I'm not sure if it's just my machine or if it's an issue with the app performance.

packages/loadingIndicator/components/SectionLoadingIndicator.tsx
nataliepina: It should still default to the same color, it will only override if a color prop is passed in. The snapshots haven't changed.

packages/loadingIndicator/stories/loadingIndicator.stories.tsx
nataliepina: It will display the color picker for the InlineLoadingIndicator, which you cannot set a color for this component. I'm not sure if it's best to break this out into its own story entirely or not. There are no props to pass into this component so it felt better to disable the controls by removing args as an option.

packages/loadingIndicator/components/SectionLoadingIndicator.tsx
nataliepina: The style in `sectionLoadingIndicator` still handles the background color as previously, and it will override this with the custom background color if it's passed in. I could pass in a fallback as an extra measure if we'd like.

nataliepina:
> Looks good! The only thing I noticed is that the color control in the storybook was really laggy when I tried to change the color. I thought the page crashed, but it was only frozen temporarily. I'm not sure if it's just my machine or if it's an issue with the app performance.
> 
> Edit: Just seen what you wrote under trade-offs. I'm having the same issue.

Yeah, I'm going to have to try to research this one some more. I'm not sure why it's only delayed with this component.

nataliepina:
I've spent some time looking into this and made some additional adjustments. 
The root cause of the slowness seems to be related to the `PageHeader` component in conjunction with the color picker, I am not sure exactly why this is. I do not feel the `PageHeader` is quintessential to display this component so I've removed it from the story. 
I've broken out the loading indicators into their own stories to better differentiate the two and to avoid confusing controls.

seialkali:
@nataliepina That's fixed it for me! Great work figuring out what the issue was!

nataliepina:
@vacas5  Do you mind taking a second look with the new updates?

vacas5 (approved):

PR #733, opened 2022-06-06
feat: a11y regression testing for storybook

nataliepina:
Closes D2IQ-90239

Following the guidelines from Storybook to set up a11y regression testing with Storybook 6.5. 

- https://storybook.js.org/blog/automate-accessibility-tests-with-storybook/
- https://storybook.js.org/docs/react/writing-tests/test-runner
- https://storybook.js.org/docs/react/writing-tests/test-runner#set-up-ci-to-run-tests

## Testing

Run storybook in one terminal window and run `npm run test-storybook` in another. 

## Checklist for PR creator


package.json
```diff
@@ -36,6 +36,8 @@
     "test": "npx jest",
     "test:integration": "./scripts/test-integration",
     "test:watch": "jest --watch",
+    "test-storybook": "test-storybook",
+    "test-storybook:watch": "test-storybook --watch",
```
vacas5, line 40: I don't think we need this one actually. You can run this locally by doing `npm run test-storybook -- --watch`

package.json
```diff
@@ -36,6 +36,8 @@
     "test": "npx jest",
     "test:integration": "./scripts/test-integration",
     "test:watch": "jest --watch",
```
vacas5, line 38: this is probably an unnecessary command here, for the same reason as my comment below

vacas5 (approved):
cool!

.github/workflows/storybook-tests.yml
```diff
@@ -0,0 +1,26 @@
+name: Storybook Tests
+on: deployment_status
```
vacas5, line 2: when do you want this to run?

package.json
nataliepina: Will remove both 👍

.github/workflows/storybook-tests.yml
nataliepina: I originally went with the method to run against the deployed storybook, but I've looked at the docs again and am going to update this to the non-deployed method. I mainly just want to integrate these checks into CI.

package.json
```diff
@@ -64,7 +65,8 @@
     "react-popper": "2.2.5",
     "react-tabs": "3.2.3",
     "react-transition-group": "4.4.2",
-    "relative-luminance": "2.0.1"
+    "relative-luminance": "2.0.1",
+    "wait-on": "6.0.1"
```
vacas5, line 69: these new ones are likely devDepenencies

nataliepina: Still trying to figure out the CI here for this command:
`    - name: Serve Storybook and Run Tests 📖
        run: |
          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
            "npx http-server storybook-static --port 6006 --silent" \
            "npx wait-on tcp:6006 && npm run test-storybook"`

clintonmedbery (commented):
I think this looks great! Just one question.

package.json
```diff
@@ -35,7 +35,7 @@
     "storybook": "start-storybook -p 6006",
     "test": "npx jest",
     "test:integration": "./scripts/test-integration",
-    "test:watch": "jest --watch",
+    "test-storybook": "test-storybook",
```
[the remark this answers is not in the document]
nataliepina: Russ suggested we could run `npm run test-storybook -- --watch` and the same for test:watch locally instead.

seialkali (approved):
Nice!

package.json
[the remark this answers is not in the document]
vacas5: yeah to me those convenience methods are overkill but that may be just an opinion. Having it there resulted in Natalie thinking that adding a `test-storybook:watch` would also be necessary, because she's less familiar with jest and how it works.

It's convenient for sure but there is cognitive load and maintenance burden on everything we write. This may be inspiring a blog post.

.github/workflows/storybook-tests.yml
```diff
@@ -0,0 +1,29 @@
+name: Storybook Tests
```
vacas5, line 1: since we have so many types of tests, can we call this Accessibility tests? or something more appropriate?

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           cache: "npm"
 
       - name: Install Dependencies 🔧
-        run: npm ci
+        run: npm install
```
mstruebing, line 28: What's the reason we switch to `install` instead of `ci` here?

PR #736, opened 2022-06-23
fix: remove linechart and chartist deps

nataliepina:
I'd like to remove the LineChart component for these reasons:
- I have continued to see problems with chartist and react-chartist with conflicting dependencies. I'd like to remove these dependencies altogether. 
- We are not using this component in the UI. 
- I will re-implement a line chart in the future with more modern and supported library choices if there is a use case

## Checklist for PR creator


mstruebing (approved):
I'm in favor of this change.
This is an excellent example of what happens over time when we do not put any effort into this project.

clintonmedbery (approved):
Nice job. Good looking out.

PR #742, opened 2022-07-18
chore: update tablemark and switch module system

mstruebing:
## Testing

* Run `npm run dist`
* Run `npm run start`

## Trade-offs

I needed to switch the module system for the updated tablemark library to work.
It seems to work but I'm not so deep into the node module system specifics.

## Checklist for PR creator


seialkali (approved):
I'm not really clued up on the module system, but I've pulled this down and run storybook without any issues

nataliepina:
We should revisit if Tablemark is really doing much for us anymore or it can be removed.

nataliepina (approved):
Good work updating the module system throughout here.

mstruebing:
> We should revisit if Tablemark is really doing much for us anymore or if it can be removed.

We definitely should, but I'm afraid that will do not save us from messing with the different module systems in the long run, I've seen more and more libraries updated to ES-Modules recently.

PR #745, opened 2022-08-01
refactor: link components

nataliepina:
Converting link components to functional components with updated types and additional cleanup.

Refactoring the components to move away from class components and better prop types to prepare for React 18.

Refactored components include:
- FocusStyleManager
- UnstyledLink
- Link
- UIKItProvider
- UIKitThemeProvider
- customLink

Some of these components impact each other which is why they are grouped together in these changes.

## Testing

Regression testing on the stories for the components listed above. 

Button components should have the same functionality, they should still have an outline or underline with keyboard navigation. 

## Checklist for PR creator

### DIFF
--- a/packages/link/components/UnstyledLink.tsx
+++ b/packages/link/components/UnstyledLink.tsx
@@ -1,32 +1,24 @@
 import React from "react";
-import { ExpandedLinkProps } from "../types";
 import { LinkComponentContext } from "../../uiKitProvider/link/context";
+import { ExpandedLinkProps } from "../types";
 
-class UnstyledLink extends React.Component<ExpandedLinkProps> {
-  static contextType = LinkComponentContext;
-  render() {
-    const LinkComponent = this.context;
+const UnstyledLink = (props: ExpandedLinkProps) => {
+  const { href, url, target, openInNewTab, children, ...rest } = props;
 
-    if (LinkComponent) {
-      return <LinkComponent {...this.props} />;
-    }
-
-    const { href, url, target, openInNewTab, children, ...rest } = this.props;
-    const rel = target === "_blank" || openInNewTab ? "noopener" : undefined;
-    const blankTargetOrDefault = !target && openInNewTab ? "_blank" : target;
-
-    return (
-      <a
-        href={href || url}
-        target={blankTargetOrDefault}
-        // https://web.dev/external-anchors-use-rel-noopener/
-        rel={rel}
-        {...rest}
-      >
-        {children}
-      </a>
-    );
+  // Context is used for `UIKitProvider` link delegation
+  const LinkComponent = React.useContext(LinkComponentContext);
+  if (LinkComponent) {
+    return <LinkComponent {...props} />;
   }
-}
 
-export { UnstyledLink as default };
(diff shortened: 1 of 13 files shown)


packages/link/components/UnstyledLink.tsx
```diff
@@ -1,32 +1,23 @@
 import React from "react";
 import { ExpandedLinkProps } from "../types";
-import { LinkComponentContext } from "../../uiKitProvider/link/context";
 
-class UnstyledLink extends React.Component<ExpandedLinkProps> {
-  static contextType = LinkComponentContext;
-  render() {
-    const LinkComponent = this.context;
+const UnstyledLink = (props: ExpandedLinkProps) => {
+  const { href, url, target, openInNewTab, context, children, ...rest } = props;
```
vacas5, line 5: context won't be a prop here.

packages/link/components/UnstyledLink.tsx
```diff
@@ -1,32 +1,23 @@
...
 
-    if (LinkComponent) {
-      return <LinkComponent {...this.props} />;
-    }
+  // Context is used for `UIKitProvider` link delegation
+  const LinkComponent = context;
```
vacas5, line 8: try `const LinkComponent = React.useContext(LinkComponentContext)` here and remove context from props.

packages/link/types.ts
```diff
@@ -2,13 +2,16 @@ import React from "react";
...
    */
   url?: string;
   /**
-   * if the `url` prop is set, this will open that link in a new tab
+   * If the `url` prop is set, this will open that link in a new tab.
    */
   openInNewTab?: boolean;
+  children?: React.ReactNode | React.ReactNode[];
+  className?: string | undefined;
+  context?: LinkComponent | undefined;
```
vacas5, line 14: you'll want to remove this to maintain the existing functionality

vacas5 (changes requested):
Not exactly how you want to refactor the context from the class component

packages/link/components/UnstyledLink.tsx
nataliepina: Thanks for taking a look at this. Looks like this does the trick here 👍🏻

nataliepina:
Screen Shot 2022-09-12 at 2 03 15 PM

I've made some additional updates to try to fix the issues with the link delegate story under the `UIKitProvider`. 
Unless there is something else to address the undefined 'className'.
I have been following the trail of errors. It looks like the next thing to update and convert might be `interactionMode.ts`, may be able to turn this into a hook. 
The issue is around the FocusStyleManager set on the Button component which the link delegate uses. I suspect it might have something to do with this being a class and the use of `this` and overall class-based logic.

vacas5 (approved):
looks like some other things changed with this PR too. I'll need a bit in the morning to look at it, but the other changes look good

nataliepina:
@vacas5 
Sadly the UIKitProvider / Link Delegate story has the error above still.

Still WIP because of this, but if anyone wants to poke around and try to fix it they are very much welcome to!

vacas5 (approved):

PR #754, opened 2022-08-16
Remove Cypress Integration Tests

nataliepina:
# Description

Historically the Cypress integration tests were set up and intended to be used with the addition of new components. I'd love for this to be considered again for the future of testing UI Kit. However, the only integration test we currently have with Cypress is to make sure that Storybook opens.
We are a major version behind on Cypress, and it is another dependency to keep up with and add to the size of the project without being fully utilized. The hope here is to remove Cypress to clean up the project and to reconsider introducing a fresh configuration for Cypress in the future.

## Testing
Are you able to npm install and start the project?
Do you see any references to the cypress integration tests that still need to be removed?

## Checklist


vacas5 (approved):
I think so

mstruebing (approved):
I've noticed that it's in there lately and wondered why it's there - glad it's gone now, thank you 🙏

PR #755, opened 2022-08-18
build: migrate ts-loader to babel-loader

nataliepina:
Closes D2IQ-92433

# Description
By default, Storybook is configured to work with Babel - https://storybook.js.org/docs/react/configure/babel
Babel’s ecosystem has evolved for supporting typescript in recent years. Benefits of fully migrating to Babel include better tree-shaking resulting in smaller bundle size compared to using ts-loader with parts of babel. Babel is overall faster than ts-loader. In many Storybook Webpack 5 migration guides, I am seeing babel used instead of ts-loader. So I've decided to try out the migration here in the hopes of decreasing our bundle size. I've also included a script to run for checking types.

By default, babel-loader will store the cache data in node_modules/. cache/babel-loader.

I believe the renovate.json file is a remnant of the Jenkins configuration as it has much older versions of dependencies and the commit history points to 4 years ago. 

https://jira.d2iq.com/browse/D2IQ-92433

## Testing
Break a prop type and run the new types test:
`npm run test:types`

## Trade-offs
There are a few things we cannot do until we update Typescript to v4+ such as including `--noEmit` in the types check. 

## Checklist


tsconfig.json
```diff
@@ -17,7 +16,10 @@
     "moduleResolution": "node",
     "skipLibCheck": true,
     "tsBuildInfoFile": ".tsbuildinfo",
-    "lib": ["es6", "dom", "es2017"]
+    "lib": ["es6", "dom", "es2017"],
+    "incremental": true,
```
mstruebing, line 20: This is such a small but great improvement! 👍

package.json
```diff
@@ -36,7 +36,8 @@
     "test": "npx jest",
     "test-storybook": "test-storybook",
     "chromatic": "chromatic",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "test:types": "tsc -p tsconfig.json"
```
mstruebing, line 40: Do we want to add this to our `pre-commit` hooks and Github-Actions?

nataliepina: Yes, we should! I'll include it here. 🎉

tsconfig.json
seialkali: Could I ask what these changes are doing? I'm not very familiar with tsconfig options.

mstruebing: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag

It makes things faster.

seialkali: Ahh ok, it was already in the config, just moved to this line. Does the placement in the config make a difference?

nataliepina: The placement does not make a difference here.

mstruebing (approved):

PR #756, opened 2022-08-20
refactor: form controls to functions

vacas5:
Refactoring some form PureComponents to be functions instead. One thing
that tripped me up was that the refactor of SelectInput threw warnings
if appearance was a required prop. Not sure how this wasn't throwing
    warnings before. Either way though, because it is defaulted, making
    it an optional prop is still backwards compatible.

# Description

## Checklist

### DIFF
--- a/packages/segmentedControl/components/SegmentedControlButton.tsx
+++ b/packages/segmentedControl/components/SegmentedControlButton.tsx
@@ -36,61 +36,53 @@ export interface SegmentedControlButtonProps {
    * Content that appears in a Tooltip when a users hovers the button
    */
   tooltipContent?: React.ReactNode;
+  children?: React.ReactNode;
 }
 
-class SegmentedControlButton extends React.PureComponent<SegmentedControlButtonProps> {
-  placeholderId = nextId("segmentedControlButton-");
-
-  render() {
-    const {
-      id = this.placeholderId,
-      isActive,
-      onChange,
-      name,
-      value,
-      tooltipContent,
-      children
-    } = this.props;
-    const segmentChildren = tooltipContent ? (
-      <Tooltip
-        id={`${id}-tooltip`}
-        trigger={<div className={segmentedControlButtonInner}>{children}</div>}
-      >
-        {tooltipContent}
-      </Tooltip>
-    ) : (
-      <div className={segmentedControlButtonInner}>{children}</div>
-    );
-
+const SegmentedControlButton = React.memo(
+  ({
+    id = nextId("segmentedControlButton-"),
+    isActive,
+    onChange,
+    name,
+    value,
+    tooltipContent,
+    children
+  }: SegmentedControlButtonProps) => {
     return (
-      <React.Fragment>
-        {/* eslint-disable jsx-a11y/role-has-required-aria-props */
-        /* This rule is asking for a `aria-checked` attribute on the input element.
-        We're already setting `checked`, so this is redundant and unnecessary.
-        */}
-        <FocusStyleManager focusEnabledClass={staticKeyboardFocusClassname}>
-          <label
-            className={cx(segmentedControlButton, {
-              [segmentedControlButtonActive]: this.props.isActive
-            })}
-            data-cy="segmentedControlButton"
-            htmlFor={id}
-          >
-            <input
-              className={visuallyHidden}
-              id={id}
-              type="radio"
-              name={name}
-              value={value}
-              checked={isActive}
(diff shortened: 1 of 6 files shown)


packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -63,15 +63,15 @@ class SegmentedControlButton extends React.PureComponent<SegmentedControlButtonP
     );
 
     return (
-      <React.Fragment>
+      <>
```
mstruebing, line 66: Do we even need this fragment?

seialkali (changes requested):
I've tested out the Segmented Control on storybook and only the last story seems to be working. All the others will only select the first option.

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -36,21 +36,21 @@ export interface SegmentedControlButtonProps {
    * Content that appears in a Tooltip when a users hovers the button
    */
   tooltipContent?: React.ReactNode;
+  children?: React.ReactNode;
 }
 
-class SegmentedControlButton extends React.PureComponent<SegmentedControlButtonProps> {
-  placeholderId = nextId("segmentedControlButton-");
+const placeholderId = nextId("segmentedControlButton-");
```
seialkali, line 42: I think this needs to be defined inside the component in order for the segmented control to work properly. See suggested change below.

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -36,21 +36,21 @@ export interface SegmentedControlButtonProps {
...
-      isActive,
-      onChange,
-      name,
-      value,
-      tooltipContent,
-      children
-    } = this.props;
+const SegmentedControlButton = React.memo(
+  ({
+    id = placeholderId,
```
seialkali, line 46:
```suggestion
    id = nextId("segmentedControlButton-"),
```

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -36,21 +36,21 @@ export interface SegmentedControlButtonProps {
...
-    const {
-      id = this.placeholderId,
-      isActive,
-      onChange,
-      name,
-      value,
-      tooltipContent,
-      children
-    } = this.props;
+const SegmentedControlButton = React.memo(
```
seialkali, line 44: How come you've added a React.memo here?

nataliepina: React.memo is meant to be an alternative when refactoring a class component using [React.PureComponent](https://reactjs.org/docs/react-api.html#reactpurecomponent) to a functional component with hooks.

packages/segmentedControl/components/SegmentedControlButton.tsx
```diff
@@ -36,21 +36,21 @@ export interface SegmentedControlButtonProps {
    * Content that appears in a Tooltip when a users hovers the button
    */
   tooltipContent?: React.ReactNode;
+  children?: React.ReactNode;
```
nataliepina, line 39: Great 👍🏻  Each one of these is one step closer to React 18 in full.

vacas5:
I'll make those changes, thanks yall

packages/segmentedControl/components/SegmentedControlButton.tsx
clintonmedbery: I am not sure I like the idea of just trying to memoize components across the board. There is a performance hit to memoization. That being said, I can find articles [that say that you should](https://attardi.org/why-we-memo-all-the-things/), and articles [that say that you shouldn't](https://dmitripavlutin.com/use-react-memo-wisely/). In my mind, we know there is a performance hit to memoizing a component and it might not be a good idea to do it for these smaller components.

vacas5 (commented):
@seialkali your suggested change resulted in less changed snapshots. Thanks for finding that

packages/segmentedControl/components/SegmentedControlButton.tsx
vacas5: agree that not everything should be auto-memoed. However in this case, have to maintain memoization to maintain the existing functionality

seialkali (approved):
Looks good to me!

nataliepina (approved):

clintonmedbery (approved):

PR #767, opened 2022-09-13
chore: update typescript

clintonmedbery:
Closes D2IQ-90920
# Description

Updated Typescript to latest. Had to fix some TS related issues. Made some comments on those and made two commits as one is for updating snapshots.

- https://jira.d2iq.com/browse/D2IQ-90920

## Testing

Really just click around and make sure things still work. The element with the biggest changes was dropdown menu.

## Checklist

### DIFF
--- a/packages/popover/components/Popover.tsx
+++ b/packages/popover/components/Popover.tsx
@@ -151,7 +151,7 @@ const Popover: React.StatelessComponent<PopoverProps> = ({
         overlayRoot={overlayRoot}
       >
         {React.isValidElement(trigger) ? (
-          React.cloneElement(trigger, {
+          React.cloneElement(trigger as React.ReactElement, {
             ...sharedTriggerProps,
             tabIndex: 0
           })
--- a/packages/popover/components/PopoverListItem.tsx
+++ b/packages/popover/components/PopoverListItem.tsx
@@ -80,11 +80,10 @@ const PopoverListItem = (props: PopoverListItemProps) => {
           itemContent: children
         };
       }
-      const isItemGraphic = (
-        child
-      ): child is typeof PopoverListItemIcon | typeof PopoverListItemAvatar =>
-        (React.isValidElement(child) && child.type === PopoverListItemIcon) ||
-        child.type === PopoverListItemAvatar;
+      const isItemGraphic = (child): boolean =>
+        React.isValidElement(child) &&
+        (child.type === PopoverListItemIcon ||
+          child.type === PopoverListItemAvatar);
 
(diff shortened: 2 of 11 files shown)


packages/popover/components/PopoverListItem.tsx
```diff
@@ -80,11 +80,10 @@ const PopoverListItem = (props: PopoverListItemProps) => {
           itemContent: children
         };
       }
-      const isItemGraphic = (
-        child
-      ): child is typeof PopoverListItemIcon | typeof PopoverListItemAvatar =>
-        (React.isValidElement(child) && child.type === PopoverListItemIcon) ||
-        child.type === PopoverListItemAvatar;
+      const isItemGraphic = (child): boolean =>
```
clintonmedbery, line 83: This took me longer to fix than I would like to admit. The problem was that here we are basically turning the child into a type that did not have props on it. Since a react element at run time has props and this functional component does not, it does not work. An even better idea would be to refactor this to explicitly take an element for the start and end position, but that seemed out of scope.

packages/textInput/components/TextInput.tsx
```diff
@@ -151,15 +151,16 @@ export class TextInput<P extends TextInputProps> extends React.Component<P> {
...
       ) => void;
     }
-
+    const additionalProps = {
+      ...{ ...inputElementProps, onChange, value, type: this.props.type }
+    };
     return (
       <input
         className={cx(inputReset, ...additionalClasses)}
-        type={this.props.type}
```
clintonmedbery, line 158: This didn't like that type was on the inputElementProps and being passed in again explicitly.

packages/popover/components/Popover.tsx
```diff
@@ -151,7 +151,7 @@ const Popover: React.StatelessComponent<PopoverProps> = ({
         overlayRoot={overlayRoot}
       >
         {React.isValidElement(trigger) ? (
-          React.cloneElement(trigger, {
+          React.cloneElement(trigger as React.ReactElement, {
```
mstruebing, line 154: I'm not so sure about the internal react types, but would it also achieve the same if we set the type of trigger to `React.ReactElement` at the props level of this component? 
I think `ReactElement`?

mstruebing (approved):
Only one small question/nitpick, nothing serious.

packages/popover/components/Popover.tsx
clintonmedbery: This produced the same error. I guess it sees the trigger as any? Some more info: https://stackoverflow.com/a/42261933/3058839

vacas5 (approved):

PR #773, opened 2022-09-20
fix: pagination as links bug, fullscreen props

vacas5:
This began as a refactor of several components away from classes. While doing so I was able to remove the Delegate library that was throwing react errors because it was very out of date. Additonally I found a bug with the way proptypes were defined for headerComponent in two components, and also a bug with the pagination when trying to use the pager buttons as links.

Was not able to refactor some tests away from enzyme with this as I'd hoped because `create` does not seem to work with jest mock functions. This may be why Clint used two patterns in his PR.

# Description

## Testing

Validate the behavior is the same in prod storybook for DonutChart, Dropdown, FullScreenView, Modal (fullscreen)

Validate that the pagination that uses links actually do something in the browser.

## Checklist


packages/fullscreenView/components/FullscreenView.tsx
```diff
@@ -16,32 +14,34 @@ export interface FullscreenViewProps {
...
   subtitle?: React.ReactNode;
-  /** Whether we automatically add padding to the body of the modal. */
+  /** Function that gets called when the modal is closed */
+  onClose: (event?: React.SyntheticEvent<HTMLElement>) => void;
+}
+export interface FullscreenViewProps extends HeaderProps {
   isContentFlush?: boolean;
   /** Custom header content component. ⚠️Use rarely and with caution⚠️ */
-  headerComponent?: React.ReactNode;
+  headerComponent?: React.JSXElementConstructor<HeaderProps>;
```
seialkali, line 23: I'm not really familiar with `React.JSXElementConstructor`, but could this be a breaking change? Say if someone was using a component for the header that didn't have `HeaderProps`?

vacas5: So, from my understanding `React.ReactNode` was the incorrect type altogether for what users were meant to be passing in (and what it did with `headerComponent`). What we want is a React component declaration: some function that will return JSX. That component will only ever get HeaderProps, but it's possible that someone would be using it with a component that took other stuff. So, I could see where this might be too strict.

Our other option is `React.JSXElementConstructor<any>`. What else could I possibly do?

seialkali: I'm actually getting a TS error trying to use the `headerComponent` prop in kommander-ui (with and w/o your changes). Maybe we could take a look in our 1:1.

vacas5: We looked and resolved the TS error in kommander-ui. Discussed and opted for an any here instead just to make sure we don't introduce a breaking change.

package.json
```diff
@@ -53,7 +53,6 @@
     "copy-to-clipboard": "3.3.2",
     "downshift": "6.1.9",
     "react-click-outside": "3.0.1",
-    "react-delegate-component": "1.0.0",
```
mstruebing, line 56: 🎉

packages/fullscreenView/components/FullscreenView.tsx
```diff
@@ -19,29 +17,30 @@ export interface FullscreenViewProps {
...
-      closeText,
-      isContentFlush,
-      onClose,
-      title,
-      subtitle,
-      bannerComponent,
-      headerComponent,
-      cypressSelectorBase = "fullscreenView"
-    } = this.props;
+const FullscreenView = React.memo(
```
mstruebing, line 30: ❓ Why do we `memo` here without an dependency array?

> If no array is provided, a new value will be computed on every render.

https://reactjs.org/docs/hooks-reference.html#usememo

packages/donutChart/components/DonutChart.tsx
```diff
@@ -27,74 +27,67 @@ export interface DonutChartProps {
...
 }
 
-class DonutChart extends React.PureComponent<DonutChartProps, {}> {
-  public render() {
-    const { data, label, text } = this.props;
-    const strokeWidth = 1.5;
-    const radius = 100 / (Math.PI * 2);
-    const diameter = radius * 2 + strokeWidth;
-    const circleCenter = diameter / 2;
+const DonutChart = React.memo(({ data, label, text }: DonutChartProps) => {
```
mstruebing, line 30: Same here, why memo without a dependency array?

mstruebing (commented):
* The `isOpen` for the dropdownable doesn't seem to work - but that's the case even in prod.
* The page is reloaded at `Previous And Next Buttons As Links` so I'm not really able to navigate.

packages/fullscreenView/components/FullscreenView.tsx
vacas5: React.memo !== React.useMemo 

https://reactjs.org/docs/react-api.html#reactmemo

vacas5: React memo recreates the `PureComponent` behavior

packages/fullscreenView/components/FullscreenView.tsx
mstruebing: ❓ Why don't we want breaking changes? If that change would be meaningful we should do it? (With the right commit message to trigger a breaking change release of course)

packages/fullscreenView/components/FullscreenView.tsx
mstruebing: 👍

packages/fullscreenView/components/FullscreenView.tsx
vacas5: breaking changes are fine but I was hoping to batch them in a forthcoming release. Seems like overkill in this case to introduce it when we are not even sure if we want to use this feature of this component in the future. I think the better breaking change would be to just remove this functionality since Kommander UI doesn't leverage it at all

vacas5:
I think the page reloading on `Previous And Next Buttons As Links` is the desired behavior... but maybe I should revert it back to what it was. Just didn't make sense as it was written to only show the urls if disabled

seialkali:
> I think the page reloading on `Previous And Next Buttons As Links` is the desired behavior... but maybe I should revert it back to what it was. Just didn't make sense as it was written to only show the urls if disabled

@vacas5 I think you should keep what you've changed it to. Sure, it isn't exactly how we'd expect it to behave, but from what I can tell, it was broken before. At least it's working now, even if the links don't change depending on what page you're on.

seialkali (approved):
👍

PR #774, opened 2022-09-21
fix: revert module change

vacas5:
This steps back a complete shift from CommonJS to esm modules, and instead makes only the file that imports tablemark use esm. Then, we use a dynamic import statement inside the build script to import that file as a workaround. Thus we're able to leave tablemark as is for now.

Closes D2IQ-92904

# Description

## Checklist


tsconfig.json
```diff
@@ -17,9 +17,7 @@
     "skipLibCheck": true,
     "tsBuildInfoFile": ".tsbuildinfo",
     "lib": ["es6", "dom", "es2017"],
-    "incremental": true,
-    "declaration": true,
-    "emitDeclarationOnly": true // to emit at least something until we can update Typescript to v4 then we will want to use `--noEmit`
```
vacas5, line 22: I moved these back out because the linter used this too and I didn't want any adverse side effects with it, either

mstruebing: 💬 Feels weird that we need an own configuration file for type checks 🤷

packages/design-tokens/scripts/build.js
```diff
@@ -17,18 +15,20 @@ const DOCS_CONFIG = "./packages/design-tokens/config.designGuidelines.json";
 const configs =
   formatCategory === "docs" ? [DOCS_CONFIG] : [CONFIG, DIST_CONFIG];
 
-configs.forEach(path => {
-  const styleDictionary = StyleDictionary.extend(path);
+import("./formats.mjs").then(({ default: formats }) => {
```
mstruebing, line 18: Is that the magic 🪄 ?

mstruebing (approved):
It seems to work fine - thanks for doing this 🥇

tsconfig.json
vacas5: I think it's the github action that needed `emitDeclarationOnly`. So, we need a config file in order to do type checking with a github action, maybe?

packages/design-tokens/scripts/build.js
vacas5: yep, this is it. The trick is that it is a Promise, so I had to find the right place to do it. I tried so many other things.

PR #776, opened 2022-09-27
refactor: typeahead component and stories

vacas5:
This one was more complicated but converted Typeahead from a class component to a functional component. All functionality should be preserved, but I did find that the CustomMenuEmptyState story did not work as expected because the empty state should be shown when there are no items. Also made a decision to remove the filteredlist typeahead because it didn't really demonstrate any functionality of the component itself, just of how a wrapping component could control its state. It seemed superfluous.

# Description

## Testing

This one had pretty good unit test coverage, but a regression of the typeahead stories against prod would be good

## Checklist

### DIFF
--- a/packages/typeahead/components/Typeahead.tsx
+++ b/packages/typeahead/components/Typeahead.tsx
@@ -60,50 +60,139 @@ export interface TypeaheadProps {
   disablePortal?: boolean;
 }
 
-interface TypeaheadState {
-  menuWidth: number | null;
-}
+const defaultItemToString = item => (item ? item.value : "");
 
-class Typeahead extends React.PureComponent<TypeaheadProps, TypeaheadState> {
-  public static defaultProps: Partial<TypeaheadProps> = {
-    menuMaxHeight: 300
-  };
-  private readonly containerRef = React.createRef<HTMLDivElement>();
-
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      menuWidth: null
-    };
-
-    this.getSelectedItems = this.getSelectedItems.bind(this);
-    this.handleSelection = this.handleSelection.bind(this);
-    this.setContainerWidth = this.setContainerWidth.bind(this);
-    this.stateReducer = this.stateReducer.bind(this);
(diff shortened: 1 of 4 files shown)


seialkali (commented):
How come the default for the typeahead has no filtering at all? Surely that defeats the purpose of the typeahead? This is on prod as well, so not something you've introduced, but I'm wondering if we should change it.

packages/typeahead/stories/Typeahead.stories.tsx
```diff
@@ -110,11 +109,8 @@ WithDisabledItem.args = {
...
 
-export const FilterWhileTyping = () => <FilteredListTypeahead items={items} />;
-
-export const CustomMenuEmptyState = () => (
-  <FilteredListTypeahead
-    menuEmptyState={<div className={padding("all")}>Nothing to show</div>}
-    items={items}
-  />
-);
+export const CustomMenuEmptyState = Template.bind({});
```
seialkali, line 112: This now only shows the empty state. I think the purpose of this story is to show an empty state when you type a value in the typeahead that has no matches against the available list. The other options should still be in the list though.

seialkali (changes requested):

packages/typeahead/stories/Typeahead.stories.tsx
vacas5: yeah I guess that's what that helper was for. I could refactor it, but we basically have two stories that demonstrate how a helper works, not how the underlying component works. The behavior of the component is that when you send in empty items, it will display the custom menu state.

vacas5: okay I just reverted the changes and reintroduced the filtered state helper, refactored as a function. Took me only a few minutes and it's easier than making a decision about removing it

nataliepina (approved):
Looking good to me. We can refine these stories in the future as well. 
I think the default state is fine to show it in a basic form, it could have a bit of a different use similar to a select in this form, though I'm we'll most likely implement filtering in most cases.

seialkali (approved):
Looks good!

PR #778, opened 2022-10-04
ci: update code checks and formatting

nataliepina:
# Description

While working on the code checks GitHub actions for DKP UI, I noticed that these checks could use some updates as well. Updates include bumps from v2 to v3. Formatting changes. The semantic version has been set to match the updated version - 19. 

## Testing

Checks and tests pass ✅ 

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,53 +1,34 @@
-# This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
-name: UI Kit Actions
-
+name: UI Kit Code Checks
 on:
+  pull_request:
   push:
     branches: [main]
-  pull_request:
-    branches: [main]
-
 jobs:
   ci:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [16.x]
     steps:
-      - uses: actions/checkout@v2
-
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,53 +1,35 @@
-# This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: UI Kit Actions
-
 on:
-  push:
-    branches: [main]
   pull_request:
+    types: [opened, reopened, synchronize]
```
mstruebing, line 4: ❓ Is there a reason why you list these types explicitly here? It seems like these three are the default after some quick googling: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
I was googling in the first place because I needed to know what the synchronize event is: https://github.com/orgs/community/discussions/24567

mstruebing (approved):

.github/workflows/ci.yml
nataliepina: Glad you've mentioned this, I wasn't sure myself. I was testing this out and going off of this example - https://d2iq.atlassian.net/wiki/spaces/ENG/pages/88834215/Migration+to+Github+Actions#Code-Checks 
@clintonmedbery do you know if there is any reason to set these explicitly?

nataliepina: >  By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword. For more 

Read through the docs some more. 
Based on this I'll remove specifying this here and in DKP.

clintonmedbery: I like the explicit declarations so I know what I am looking at without having to hunt down documentation, but for me it is not a deal breaker wither way.

nataliepina:
Looks like I've encountered the GitHub actions issue again where I cannot merge. Closing this and creating a new PR.

PR #814, opened 2022-10-12
refactor: migrate DropdownMenu Tests to RTL

davidtaing:
# Description
Summary:
- Migrate tests in DropdownMenu.test.tsx from Enzyme to React Testing
  Library. 
- refactored and removed events that had no effect on test results
- "calls onSelect prop with the selected value" passed when all events 
  were removed; added expect to increase test robustness.
  
Some tests were refactored as part of this migration as Enzyme and RTL
have approaches to testing. RTL has more of a test "what the user sees" 
approach.

#810 

## Testing

- Run tests via `npm run test` or `npx jest --watch`.

## Trade-offs

N/A

## Screenshots

Screen Shot 2022-10-13 at 8 50 00 am

## Checklist

- [X] I have reviewed the changes and provided detail to the sections above


packages/dropdownMenu/tests/DropdownMenu.test.tsx
```diff
@@ -50,10 +47,10 @@ describe("Dropdown", () => {
       </DropdownMenu>
     );
 
-    expect(toJson(component)).toMatchSnapshot();
+    expect(baseElement).toMatchSnapshot();
```
davidtaing, line 50: I found baseElement to be more robust that asFragment.

After rebuilding the snapshots, and changing the menu item text, I found the test to still pass.

With baseElement, changes to the menu item text failed the test, as I expected.

packages/dropdownMenu/tests/DropdownMenu.test.tsx
```diff
@@ -150,16 +151,19 @@ describe("Dropdown", () => {
...
 
-    const trigger = component.find(`#${triggerId}`);
-    expect(component.find(DropdownContents).length).toBe(0);
-    trigger.simulate("click");
-    expect(component.find(DropdownContents).prop("isOpen")).toBe(true);
-    trigger.simulate("click");
-    expect(component.find(DropdownContents).length).toBe(0);
+    const trigger = screen.getByText(/dropdown trigger/i);
+
+    expect(screen.queryByText(/edit/i)).toBeFalsy();
```
davidtaing, line 156: Modifed logic due to the differences in testing approaches between Enzyme and RTL. Grab this component programmatically via a CSS selector vs grab "what the user sees".

packages/dropdownMenu/tests/DropdownMenu.test.tsx
```diff
@@ -255,22 +255,18 @@ describe("Dropdown", () => {
...
     trigger.simulate("keyDown", {
       key: "Enter"
     });
-    trigger.simulate("blur");
 
     // using `expect.anything()` because the second parameter is an object
     // that comes from Downshift, and there's no good way to know exactly
     // what to expect
     expect(onSelectFn).not.toHaveBeenCalledWith("edit", expect.anything());
+    expect(onSelectFn).toHaveBeenCalled();
```
davidtaing, line 266: Added this expect as the original expect statement is a false-positive test.

The edit menu item in this test is disabled and cannot be selected by the user. I think we could break this out into another test where we only have disabled menu-items and test for `onSelectFn.not.toHaveBeenCalled()`.

davidtaing: In addition to the above, we could modify this to "selects second item and passes it's value to onSelect".

mstruebing (approved):
This is a great refactor, great work! 🚀

packages/dropdownMenu/tests/DropdownMenu.test.tsx
nataliepina: Do you have more information on why `baseElement` would be preferred here?

I read through this Kent C. Dodds's article recently and he mentioned this, 
> The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.
> 
> The only exception to this is if you're setting the container or baseElement which you probably should avoid doing (I honestly can't think of a legitimate use case for those options anymore and they only exist for historical reasons at this point).
This is the article - https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen 

And from the [RTL docs](https://testing-library.com/docs/react-testing-library/api/#baseelement-1), 

> Note: the queries returned by the render looks into baseElement, so you can use queries to test your portal component without the baseElement.

We're definitely open to learning new ideas and improvements. Would love to hear your thoughts!

nataliepina:
Really detailed and improvement-driven work @davidtaing!

packages/dropdownMenu/tests/DropdownMenu.test.tsx
davidtaing: For demonstration purposes, I've changed the `renders an open dropdown with multiple sections` test and updated the snapshots
asFragment before text edit

This is me changing the Trigger text. It fails, so we're good here.
asFragment - trigger text changed

I originally modified the text in the menu items and the changes weren't caught.

But I just found this a moment. This is really interesting.
Screen Shot 2022-10-14 at 1 22 33 am


### How I arrived at baseElement
Screen Shot 2022-10-14 at 1 11 40 am

This is going to sound funny. But I googled "snapshot testing react testing library", landed on the docs, then searched for "snapshot". I didn't even read the docs 🤣.

I ran the same checks as above and baseElement catches these changes.

davidtaing: I just checked and the DropdownMenu definitely using a portal so the Menu Contents are being rendered in a different subtree.

It's definitely an edge case. This has been a really interested rabbit-hole to dive into.

nataliepina: Love the screenshots. Thank you. That makes sense to me in that we can reach for either asFragment, baseElement, or screen depending on the complexity and the type of element. Really interesting, and I appreciate you sharing your findings!

nataliepina (approved):
Awesome 🎉  Thank you for the effort put into these changes.

packages/dropdownMenu/tests/DropdownMenu.test.tsx
clintonmedbery: This is very informative. Thank you for the thought out comments!

PR #816, opened 2022-10-13
feat: add message control to infobox stories

seialkali:
# Description

This PR adds a control for the message property to the `InfoBox` and `InfoBoxBanner` stories, and also adds a type to the template so we can remove the message prop where the `InfoBox` is used.

## Testing

Go to `InfoBox` and `InfoBoxBanner` stories in Storybook and confirm that they message control works

## Trade-offs

I haven't added the message control to the `InfoBoxInline` stories because one of the stories uses custom markup and I'm not sure what's the best way to allow users to control this in a way that is intuitive, and I'm not sure if it's even possible. 

## Checklist

### DIFF
--- a/packages/infobox/stories/InfoBox.stories.tsx
+++ b/packages/infobox/stories/InfoBox.stories.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import { Story, Meta } from "@storybook/react";
 import { InfoBox } from "../index";
 import { PrimaryAction, SecondaryAction } from "./helpers/actions";
+import { InfoBoxProps } from "../components/InfoBox";
 
 export default {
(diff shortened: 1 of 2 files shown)


packages/infobox/stories/InfoBox.stories.tsx
```diff
@@ -15,20 +16,19 @@ export default {
     },
     className: {
       control: { disable: true }
+    },
+    message: {
+      control: { type: "text" },
```
nataliepina, line 21: I thought storybook was smart enough to handle default controls by now but this does help so that we don't need to wrap the text in a string in the input.

nataliepina (approved):

PR #824, opened 2022-10-17
feat: optional max-width for FullscreenView header

vacas5:
If child content of FullscreenView is wrapped in a container, it would be nice to be able to contain the header of this component also for visual consistency. Opted to keep the padding as is and just account for it in the max width so that consumers can pass in our design tokens as is without having to do the math themselves to account for the padding left and right.

Addresses D2IQ-92266

# Description

https://d2iq.atlassian.net/browse/D2IQ-92266

## Testing

Created a new story for this use case.

## Trade-offs

Accounting for padding from within the component. May be confusing to the user but border-box would be the alternative and I'm not sure if that would have fallout.

## Checklist

### DIFF
--- a/packages/fullscreenView/components/FullscreenView.tsx
+++ b/packages/fullscreenView/components/FullscreenView.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { cx } from "@emotion/css";
+import { cx, css } from "@emotion/css";
 import { ButtonProps } from "../../button/components/ButtonBase";
 import { flex, padding, flexItem } from "../../shared/styles/styleUtils";
 import { modalContent, fullscreenModalHeader } from "../style";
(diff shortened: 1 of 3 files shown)


mstruebing (approved):
👍

packages/fullscreenView/components/FullscreenView.tsx
```diff
@@ -24,56 +25,65 @@ export interface FullscreenViewProps {
   onClose: (event?: React.SyntheticEvent<HTMLElement>) => void;
   /** The base of the `data-cy` value. This is used for writing selectors in Cypress. */
   cypressSelectorBase?: string;
+  /** The css max width around the header contents  */
+  containerMaxWidth?: string;
```
mstruebing, line 29: ❓ Should we allow a certain set of values here or just any string?

vacas5: as a user I think I would want to be able to set this to any value, otherwise it loses usefulness pretty fast. There could be all kinds of situations in which I wouldn't want to use a particular breakpoint

packages/fullscreenView/components/FullscreenView.tsx
```diff
@@ -24,56 +25,65 @@ export interface FullscreenViewProps {
...
-          data-cy={`${cypressSelectorBase}-content`}
-        >
-          {children}
+  let container: string = "";
+  if (containerMaxWidth) {
+    // account for left and right padding
+    container = css`
+      margin-left: auto;
+      margin-right: auto;
+      max-width: calc(${containerMaxWidth} + ${spaceXl} + ${spaceXl});
```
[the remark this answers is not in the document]
vacas5: okay I've opted to use border-box instead because it's cleaner, and should make this comment moot.

nataliepina (approved):
Updates check out to me, and the story is still working well ✅ Nicely done.

PR #828, opened 2022-10-19
fix: contain FullscreenView with classname

vacas5:
In order to support responsive layouts and different container sizes at various widths, this really should just take a classname so that a client has total control. I tried to borrow the container classname from the container component and go with a boolean argument like "contained" but I found that our usage in DKP needed to change the box-sizing anyway to get it to be consistent. Rather than have both props, just went with one, more versatile prop.

# Description

## Checklist


seialkali (approved):
Nice!

mstruebing (approved):

nataliepina (approved):
Makes sense 👍🏻  This should be fine.

PR #829, opened 2022-10-19
refactor: converts ToggleInputList class component into a function co…

leonardomaier:
…mponent

# Description

https://github.com/dcos-labs/ui-kit/issues/825

## Testing

Storybook and by running tests.

## Trade-offs

None.

## Checklist

- [x] This PR contains breaking changes and states in the commit message body ("BREAKING CHANGE: …")

### DIFF
--- a/packages/toggleInputList/components/ToggleInputList.tsx
+++ b/packages/toggleInputList/components/ToggleInputList.tsx
@@ -68,54 +68,28 @@ export interface ToggleInputListProps {
   /**
    * Sets the current appearance of the label component. This defaults to InputAppearance.Standard, but supports `InputAppearance.Error` & `InputAppearance.Success` appearances as well.
    */
-  labelAppearance: InputAppearance;
+  labelAppearance?: InputAppearance;
   /**
    * Whether the inputs are radio buttons
    */
   isRadioGroup?: boolean;
 }
 
-class ToggleInputList extends React.PureComponent<ToggleInputListProps, {}> {
-  public static defaultProps: Partial<ToggleInputListProps> = {
-    showListLabel: true,
-    vertAlign: "center",
-    selectedItems: [],
-    labelAppearance: InputAppearance.Standard
-  };
-
-  constructor(props) {
-    super(props);
-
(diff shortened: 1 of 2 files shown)


packages/toggleInputList/components/ToggleInputList.tsx
```diff
@@ -168,7 +133,7 @@ class ToggleInputList extends React.PureComponent<ToggleInputListProps, {}> {
         hintContent,
         id,
         inputLabel,
-        name: inputListId,
+        name: id,
```
nataliepina, line 136: Could you explain a bit more on why this should be changed from `inputListId` which set name to `radio` as seen in the original snapshots to now it being set to an `id` value?

leonardomaier: Well noticed, that was my mistake :)

There were two conflicting `id` attributes and I put the wrong one at the `name` attribute.

I just fixed it in the last commit.

packages/toggleInputList/components/ToggleInputList.tsx
```diff
@@ -68,54 +68,28 @@ export interface ToggleInputListProps {
   /**
    * Sets the current appearance of the label component. This defaults to InputAppearance.Standard, but supports `InputAppearance.Error` & `InputAppearance.Success` appearances as well.
    */
-  labelAppearance: InputAppearance;
+  labelAppearance?: InputAppearance;
```
mstruebing, line 71: Why do we change this property from being required to being optional?

leonardomaier: Once you turn the class component into a function component it will ask for the required prop even if you have defined a default value.

As `labelAppearence` has a default value `InputAppearance.Standard` I changed the prop to optional.

nataliepina (approved):

packages/toggleInputList/components/ToggleInputList.tsx
mstruebing: Thank you for the explanation, this is fine then. 👍

mstruebing (approved):

PR #843, opened 2022-10-21
fix: forwardrefs to FocusStyleManager children

vacas5:
I am not sure that I did this correctly. This is a tough then where the child component could literally be anything, so typing the ref is very difficult. In some cases its a button, others an a, sometimes a label. In any event, for the ui-kit components that are children of FocusStyleManager, I wrapped them in forwardRef. Resolves the console warning issues for the pages that use FocusStyleManager.

Closes D2IQ-93522

# Description

https://jira.d2iq.com/browse/D2IQ-93522

## Testing

Button stories and unstyled link stories should match production

## Trade-offs

I'd really like some other thoughts here. I am not sure if the change I made to the type of LinkComponent is breaking or not. 

## Checklist

### DIFF
--- a/packages/button/components/ButtonBase.tsx
+++ b/packages/button/components/ButtonBase.tsx
@@ -91,56 +91,24 @@ export interface ButtonBaseProps extends ButtonProps {
   className?: string;
 }
 
-const ButtonBase = (props: ButtonBaseProps) => {
-  const {
-    appearance,
-    children,
-    className,
-    disabled,
-    iconStart,
-    iconEnd,
-    isInverse,
-    isProcessing,
-    isFullWidth,
-    onClick,
-    type = "button",
-    url,
-    openInNewTab,
-    ...other
-  } = props;
-
-  const buttonClassName = cx(
-    buttonReset,
-    button(appearance),
-    buttonBase,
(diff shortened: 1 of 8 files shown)


nataliepina (approved):
This seems to be working well, the stories are good, and no console errors. Thank you for fixing it.

packages/button/components/ButtonBase.tsx
```diff
@@ -91,115 +91,102 @@ export interface ButtonBaseProps extends ButtonProps {
...
+      iconEnd,
+      isInverse,
+      isProcessing,
+      isFullWidth,
+      onClick,
+      type = "button",
+      url,
+      openInNewTab,
+      ariaHaspopup,
+      ...other
```
seialkali, line 145: I found another console warning while testing this out:

> Warning: Invalid ARIA attribute `ariaLabel`. Did you mean `aria-label`?

It's a minor fix if you were happy to add it to this PR. You would need to destructure `ariaLabel` here, and then add a `aria-label` prop in the button element on line 195 and set `ariaLabel` as the value.

nataliepina:
Sorry merged some test updates and gave the snapshot conflicts, after that we can merge away!

PR #846, opened 2022-10-22
refactor: migrate table component tests from enzyme to rtl

manav-06:
# Description
Updates `tablev2.test.tsx` tests from enzyme to react-testing-library

#818 

## Testing
`npm run test -t tablev2.test.tsx`

## Trade-offs
None

## Checklist


packages/tablev2/tablev2.test.tsx
```diff
@@ -260,7 +259,9 @@ describe("Table v2", () => {
       );
 
       expect(sorterFn).toHaveBeenCalledWith("asc", 1);
-      component.find("#name button").simulate("click");
+
+      fireEvent.click(container.querySelector("#name button")!);
```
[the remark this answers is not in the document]
taksuparth: @manav-06 this is because with querySelector operator it is not always certain we will get that element. I would recommend using getByTestId and assigning the proper test id to the element.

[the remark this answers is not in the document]
manav-06: @taksuparth @nataliepina please review.

packages/tablev2/tablev2.test.tsx
```diff
@@ -237,7 +237,7 @@ describe("Table v2", () => {
   describe("Sorting", () => {
     it("calls the function passed to the 'sorter' prop when clicking the table headings", () => {
       const sorterFn = jest.fn();
-      const component = mount(
+      render(
```
nataliepina, line 240:
```suggestion
      const { getByTestId } = render(
```

There may be some examples with screen but I think it's a cleaner implementation to set the const here with the render. and then use it below with the event firing.

packages/tablev2/tablev2.test.tsx
```diff
@@ -260,7 +260,9 @@ describe("Table v2", () => {
       );
 
       expect(sorterFn).toHaveBeenCalledWith("asc", 1);
-      component.find("#name button").simulate("click");
+
+      fireEvent.click(screen.getByTestId("table-headercell-name-button"));
```
nataliepina, line 264:
```suggestion
  fireEvent.click(getByTestId("table-headercell-name-button"));
```
Then you will not need screen.

clintonmedbery (approved):
Thanks for updating and thanks for contributing!

nataliepina (approved):
Great work @manav-06 thank you!

packages/tablev2/tablev2.test.tsx
taksuparth: @nataliepina I found this amazing article by Kent( Owner of react testing library ) https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen
where he states otherwise. I think we should use the screen to query elements moving forward.

nataliepina: I have also read that one! It sparked a discussion on a previous PR even. We talked about it on the UI team and decided to opt for the render queries. They are nearly the same we've just gone with the latter for consistency as the initial bulk refactor of our tests used mostly render queries. We could consider refactoring down the line to use screen if there is a solid reason to other than the man KCD says to 😄 
Happy to hear out if there are any other strong reasons to use screen instead if you have any thoughts! 

I had a read through this one before option for consistency - https://stackoverflow.com/questions/64792767/what-is-the-difference-between-getbytext-and-screen-getbytext-in-rtl

PR #864, opened 2022-10-25
fix: rollback ref on unstyled link

vacas5:
Consuming LinkComponent has been a challenge downstream. We don't actually need the focus style manager to handle unstyled links as far as I can tell, so the console warnings should be resolved without having to keep drilling down into child Link components.

# Description

## Checklist


nataliepina (approved):
This looks good on my part. I've tested with `npm link` and I didn't notice any console warnings. The stories all look to be working well. The implementation looks cleaner even. So it's a ✅  for me.

PR #867, opened 2022-10-26
test: toaster tests with RTL

vacas5:
Some of these tests were testing internal component state. Opted instead to try and use visual cues that would result from such a state. This will make refactoring the component much easier.

# Description

## Checklist

### DIFF
--- a/packages/toaster/components/Toaster.tsx
+++ b/packages/toaster/components/Toaster.tsx
@@ -15,125 +15,97 @@ interface ToasterProps {
   children?: Toast | Toast[];
 }
 
-class Toaster extends React.PureComponent<ToasterProps, { toasts: Toast[] }> {
-  public timeouts: number[] = [];
+const Toaster = ({ children }: ToasterProps) => {
+  const timeouts = React.useRef<number[]>([]);
 
-  constructor(props) {
-    super(props);
+  const [toasts, setToasts] = React.useState<Toast[]>(
+    React.Children.toArray(children) as Toast[]
+  );
 
-    this.cloneToast = this.cloneToast.bind(this);
-    this.dismissToast = this.dismissToast.bind(this);
-    this.clearTimeouts = this.clearTimeouts.bind(this);
-    this.restartTimeouts = this.restartTimeouts.bind(this);
-    this.setTimer = this.setTimer.bind(this);
+  const handleDismissToast = (id: ToastId = "") => {
+    const toastToDismiss = toasts.find(toast => id === toast.props.id);
+    if (toastToDismiss && toastToDismiss.props.onDismiss) {
+      toastToDismiss.props.onDismiss();
+    }
+    setToasts(toasts.filter(toast => id !== toast.props.id));
+  };
 
-    this.state = {
-      toasts: React.Children.toArray<Toast>(this.props.children).map(
-        this.cloneToast
-      )
+  const restartTimeouts = () => {
+    toasts.forEach((toast: Toast, index: number) => {
+      if (toast.props.autodismiss) {
+        timeouts.current.push(
+          window.setTimeout(() => {
+            handleDismissToast(toast.props.id);
+          }, DELAY_TIME + MARGINAL_DELAY * index)
+        );
+      }
+    });
+  };
+
+  const clearTimeouts = () => {
+    timeouts.current.forEach(clearTimeout);
+  };
+
+  React.useEffect(() => {
+    return () => {
+      clearTimeouts();
     };
-  }
+  }, []);
 
-  public componentWillUnmount() {
-    this.clearTimeouts();
-  }
-
-  public componentDidMount() {
-    this.restartTimeouts();
-  }
-
-  public UNSAFE_componentWillReceiveProps(props: ToasterProps) {
-    const children = React.Children.toArray(props.children);
-    const childIds = children.map(toast => toast.props.id);
-    const currentIds = this.state.toasts.map(toast => toast.props.id);
+  React.useEffect(() => {
+    const toastChildren = React.Children.toArray(children) as Toast[];
+    const childIds = toastChildren.map(toast => toast.props.id);
+    const currentIds = toasts.map(toast => toast.props.id);
 
     if (
       !currentIds.every(e => childIds.includes(e)) ||
       !currentIds.length ||
       childIds.length !== currentIds.length
     ) {
-      this.setState(
-        () => ({ toasts: children.map(this.cloneToast) }),
-        () => {
-          this.state.toasts
-            ?.filter(t => t.props.autodismiss)
-            .map(this.setTimer);
-        }
-      );
+      setToasts(toastChildren);
     }
-  }
+  }, [children]);
 
-  public render() {
-    return (
-      <div
-        className={cx(toaster, margin("all"), marginAt.medium("all", "l"))}
-        data-cy="toaster"
+  React.useEffect(() => {
+    restartTimeouts();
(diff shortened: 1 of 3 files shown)


packages/toaster/tests/Toaster.test.tsx
```diff
@@ -1,105 +1,133 @@
...
+    expect(setTimerSpy).toHaveBeenCalledTimes(1);
+
+    const toasterList = screen.getByTestId("toaster-list");
+    await user.hover(toasterList);
+
     expect(clearTimeoutsSpy).toHaveBeenCalled();
-    component.find("ol").simulate("mouseLeave");
-    expect(setTimerSpy).toHaveBeenCalled();
+
+    await user.unhover(toasterList);
```
nataliepina, line 37: **unhover** _nice_

nataliepina (approved):
Looks good to me. I think the use of `screen` is fine here. We discussed opting for the render queries but I think we should come to a group consensus in a sig meeting. For components like this, screen makes sense.

vacas5:
I think I missed that discussion. Happy to refactor if you prefer

vacas5:
a lot changed here so I'd appreciate another review

nataliepina:
☝🏻  Can we get some extra reviews here @seialkali @mstruebing @clintonmedbery?

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
-                    )}
-                  >
-                    {toast}
-                  </li>
-                )}
-              </Transition>
-            ))}
-          </TransitionGroup>
-        </ol>
-      </div>
```
taksuparth, line 98: component wrapped under React.memo should be given displayname to make debugging easier in react devtools. Else it will show up as Anonymous component

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
+      clearTimeouts();
+    };
+  }, []);
 
-  public UNSAFE_componentWillReceiveProps(props: ToasterProps) {
-    const children = React.Children.toArray(props.children);
-    const childIds = children.map(toast => toast.props.id);
-    const currentIds = this.state.toasts.map(toast => toast.props.id);
+  React.useEffect(() => {
+    const childIds = toastChildren.map(toast => toast.props.id);
```
taksuparth, line 56: Can't we use `Set` here?
1. Array.includes() has a time complexity o(n)
2. Set.has() has a time complexity o(1)

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
+    const currentIds = toasts.map(toast => toast.props.id);
 
     if (
       !currentIds.every(e => childIds.includes(e)) ||
       !currentIds.length ||
       childIds.length !== currentIds.length
```
taksuparth, line 62: I feel order of this condition can be improved. 
```
!currentIds.length ||
childIds.length !== currentIds.length ||
!currentIds.every(e => childIds.includes(e))
```
This will save iterating over the loop due to short circuit if there are no child elements

seialkali (changes requested):
There's something funky going on with the Toaster auto dismiss story in storybook. The button is moving as if the toast is appearing, but I'm not seeing it. This is definitely different from the behavior in production.

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
-    this.setTimer = this.setTimer.bind(this);
-
-    this.state = {
-      toasts: React.Children.toArray<Toast>(this.props.children).map(
-        this.cloneToast
-      )
-    };
-  }
+  const handleDismissToast = (id: ToastId = "") => {
+    const toastToDismiss = toasts.find(toast => id === toast.props.id);
```
taksuparth, line 25: this can be improved by keeping a memoized map on toasts storing 
```
Map<string, Toast>
```
i.e. ToastId vs Toast

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
+    if (toastToDismiss && toastToDismiss.props.onDismiss) {
+      toastToDismiss.props.onDismiss();
+    }
+    setToasts(toasts.filter(toast => id !== toast.props.id));
+  };
+
+  const restartTimeouts = () => {
```
taksuparth, line 32: shouldn't this be memoized inside useCallback?

packages/toaster/components/Toaster.tsx
```diff
@@ -15,124 +15,96 @@ interface ToasterProps {
...
-  public render() {
-    return (
-      <div
-        className={cx(toaster, margin("all"), marginAt.medium("all", "l"))}
-        data-cy="toaster"
+  }, [children]);
+
+  React.useEffect(() => {
+    restartTimeouts();
+  }, [toasts]);
```
taksuparth, line 70: effect should be dependent on `restartTimeouts` and not on `toasts`

nataliepina:
I am seeing the same issues with the Autodismiss story. I can't see the toast when I click `make me toast` It would be nice if you could see the full countdown on that story compared to production also.

packages/toaster/components/Toaster.tsx
vacas5: see the other comment above. My main goal here is to get the class migrated to a function, not to refactor all the other logic

packages/toaster/components/Toaster.tsx
vacas5: actually I was off here. This was called in the old componentDidMount, so I am adding it to the useEffect without any dependencies to recreate that.

packages/toaster/components/Toaster.tsx
vacas5: I don't typically memoize unless I have a particular need for it or notice performance issues. Would useCallback here preserve some functionality that the public method on the class had here? https://github.com/dcos-labs/ui-kit/blob/main/packages/toaster/components/Toaster.tsx#L129

packages/toaster/components/Toaster.tsx
vacas5: I am sure there are improvements to be made here but this effort is around refactoring from class to functional component

packages/toaster/components/Toaster.tsx
vacas5: okay looks like this would maintain the same referential integrity that would have existed in the class, so I'll wrap the functions around it https://stackoverflow.com/questions/60344321/how-to-convert-functional-component-using-hooks-to-class-component

packages/toaster/components/Toaster.tsx
vacas5: on second look the reason I did this was because this was called on mount in the class, but then again in the component will receive props. Both seemed to be based on changes to `toasts` state. So, now I am using useEffect to change this when toasts change.

packages/toaster/components/Toaster.tsx
vacas5: the existing one is buggy however because it will create an infinite loop if both are empty arrays, so I'll modify it a bit

vacas5:
@seialkali should be fixed now. Also, this fixes an issue where the initial toasts, when added, fly in with a CSS transition.

seialkali (approved):
Looks good! Thanks for fixing that auto dismiss story 🎉

PR #872, opened 2022-11-04
fix: refactor toggleboxgroup to prevent rerenders

nataliepina:
# Description

We had seen some re-rendering issues with the ToggleBoxGroup throughout DKP UI. This could be seen through element flashes displaying the re-rendering.
Through some testing with UI Kit linked into the UI, we were able to confirm these changes resolve the issues. 
 It seems that the issue was something to do with the inner function to create ToggleBoxes, which is not an ideal pattern. It simplifies this component to have a wrapper element outside of it to pass in the label and children.

## Testing

You could test out npm link with these changes in the UI. 

## Checklist


vacas5 (approved):
yes plz

PR #873, opened 2022-11-07
PageHeader new custom properties

nataliepina:
# Description
Changes TLDR;
- new component properties
- improved stories

These changes allow for a custom `className` and `dataCy` prop to be passed into the component. This can help to set a container on the PageHeader.

I've included some refactoring and improvements to the PageHeader stories. Adding comments around the types will improve the documentation under the Docs tab under Description in the props table. 

## Testing

The best way to test this is with `npm link` and a custom className. 

## Trade-offs
With the changes to the stories, it defaulted set values and users cannot enter custom breadCrumbElements or actions. However, because these are arrays it is already hard with Storybook and IMO it's better if these are hard set instead of breaking when a user enters an object or array that storybook cannot process. 

## Checklist

### DIFF
--- a/packages/pageheader/components/PageHeader.tsx
+++ b/packages/pageheader/components/PageHeader.tsx
@@ -14,8 +14,22 @@ import PageHeaderTabs from "../components/PageHeaderTabs";
 import { fillHeight } from "../style";
 
 export interface PageHeaderProps {
+  /**
+   * Breadcrumbs to display on the left side of the PageHeader.
+   */
   breadcrumbElements: React.ReactNodeArray;
+  /**
+   * Actions to display on the right side of the PageHeader.
+   */
   actions?: Array<React.ReactElement<any>>;
+  /**
+   * Human-readable selector used for writing tests.
+   */
+  "data-cy"?: string;
+  /**
+   * Pass in custom CSS properties.
+   */
+  className?: string;
   children?: React.ReactNode;
 }
 
@@ -24,7 +38,9 @@ export const pageHeaderPaddingSize: SpaceSizes = "l";
 const PageHeader = ({
   breadcrumbElements,
   actions = [],
-  children
+  children,
+  "data-cy": dataCy = "pageHeader",
+  className
 }: PageHeaderProps) => {
   const hasTabsChild = React.Children.toArray(children).some(
     child => React.isValidElement(child) && child.type === PageHeaderTabs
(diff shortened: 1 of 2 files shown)


packages/pageheader/components/PageHeader.tsx
```diff
@@ -14,30 +14,53 @@ import PageHeaderTabs from "../components/PageHeaderTabs";
...
   breadcrumbElements: React.ReactNodeArray;
+  /**
+   * Actions to display on the right side of the PageHeader.
+   */
   actions?: Array<React.ReactElement<any>>;
   children?: React.ReactNode;
+  /**
+   * Contains the header to a specified width.
+   */
+  maxWidth?: React.CSSProperties["maxWidth"];
```
vacas5, line 29: so... just because of weird content box / border box stuff, I'd actually suggest just doing this via className

vacas5 (commented):
think you'll run into weird spacing issues like i did when you go to integrate this

packages/pageheader/components/PageHeader.tsx
nataliepina: FWIW I did test it out with npm link 
Screen Shot 2022-11-07 at 9 34 57 AM

nataliepina: I think the FullscreenView is lacking border-box but the PageHeader does have it set around one of the outer divs.

nataliepina: Is a classname still preferred here? 
It seems it would have to be set on the outermost div instead of this one. The outer cannot have the maxWdith set on it or the bottom border will not span the entire page. 

Then we'll likely need to pass something like this in the UI:
```
  const pageHeaderContainer = css`
    > div {
      max-width: "600px";
      margin: 0 auto;
    }
  `;
```

nataliepina: I'm going to update this to a className along with some other changes.

vacas5 (approved):

vacas5 (commented):
did you want to add a story for this or can I engage it via an existing story?

nataliepina:
> did you want to add a story for this or can I engage it via an existing story?

I decided against it. Now that it's for a classname I don't think it should have a special story for it. It's quite hard to inject classnames via the storybook controls since it's expecting a string. We'll have classNames throughout most of the components eventually as well so I don't think we'll want custom stories for all of those.

mstruebing (approved):

PR #875, opened 2022-11-08
feat: react 18

nataliepina:
includes test and type updates to support react 18

Closes D2IQ-92806

# Description

In order to support React 18 there are some changes that we'll be reliant on. 
Required changes included:
- There is a peer dependency error that will break CI if we do not update react-tabs, which then included some type updates for the Tabs component. 
- There were many locations where types for children had to be updated.
- Test updates were required. Use of `.click()` from react-test-renderer had to be refactored to use fireEvent from testing-library/react instead. I found locations using `create` from react-test-renderer that were broken and I aligned them with the rest of our testing updates using render queries from @testling-library/react. Updating both of these removed the use of react-test-renderer so it has been uninstalled along with the related type dependency. The use of this package has a historical context that dates back to 5 years ago when some tests began migrating. Everything that can be done with it can be used with RTL. Its main use is for snapshot testing which. I opted to remove it to align with our current RTL testing paradigm. 

After this, we will want to follow up and update other dependencies as there are some peer deps warnings. We'll want to update Emotion deps to the latest in particular. 

I tested this in the UI with npm link and did not encounter any errors.

## Testing
- Tests should pass
- Storybook should start up and there should not be any component errors
- CI should be all ✅ 
- Testing with `npm link` in the UI is suggested

## Checklist

### DIFF
--- a/packages/card/tests/ButtonCard.test.tsx
+++ b/packages/card/tests/ButtonCard.test.tsx
@@ -1,7 +1,6 @@
 import React from "react";
 import { createSerializer } from "@emotion/jest";
 import { render } from "@testing-library/react";
-import { create } from "react-test-renderer";
 import { ButtonCard } from "../";
 
 expect.addSnapshotSerializer(createSerializer());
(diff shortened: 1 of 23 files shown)


packages/card/tests/ButtonCard.test.tsx
```diff
@@ -33,34 +32,4 @@ describe("ButtonCard", () => {
...
-    const component = create(<ButtonCard isInput>default</ButtonCard>);
-    const buttonCardNodeProps = component.root.findByProps({
-      "data-cy": "buttonCard"
-    }).props;
-    const propKeys = Object.keys(buttonCardNodeProps);
-
-    accessibilityProps.forEach(prop => {
-      expect(propKeys.includes(prop)).toBe(false);
-    });
-  });
```
nataliepina, line 65: These two tests were really problematic. I am opting to remove because of the other methods for accessibility testing we have in the application such as the storybook addon and the a11y regression tests.  Testing should be focused on behaviors.

clintonmedbery: This makes sense as we shift to testing for behaviors.

clintonmedbery (approved):
I clicked through this and put it in kommander ui and didn't find any issues!

mstruebing (approved):
Great work, thank you 🙏

seialkali (approved):
Awesome work pushing this forward and getting it done!

PR #876, opened 2022-11-10
Dependency Updates and MDX2 support

nataliepina:
# Description

After Storybook was updated, a few issues came up with our mdx file rendering which is used for the Welcome, Colors, and Typography pages.
The issues are common and outlined here - https://github.com/storybookjs/storybook/issues/18094
I followed the thread and the suggested workarounds have fixed the issues. In Storybook 7 (coming soon), mdx2 will be fully supported and suggested so we may be able to update some of these configurations at that point.
Additionally, I've had to remove the fragments wrapping the components in the mdx files which were previously required to render multiple components.

## Testing
1. npm install
2. Start storybook and check the pages
3. Tests pass ✅ 

## Checklist


.storybook/main.js
```diff
@@ -18,5 +18,8 @@ module.exports = {
       }
     },
     "@storybook/addon-a11y"
-  ]
+  ],
+  features: {
+    previewMdx2: true
+  }
```
nataliepina, line 24: Required to support mdx2, mdx files will render as blank pages.

mstruebing (approved):
🥇

PR #891, opened 2022-11-30
fix: allow valid css height/width types

nataliepina:
Allow height and width types to accept any valid
CSS height/width properties such as vw, ems,
and other string values beyond numbers.

# Description

I encountered this when attempting to use a value with `vh` where I got a type error. These values should accept any valid related CSS height or width property, such as ems, rems, vh, vw, etc.  Followed the trail up to the `PopoverBox` and update those types there too. 

Included a few other tweaks in here too such as a typo fix, and a className prop for the Dropdown. 

Note: The `React.CSSProperties` type will handle strings or numbers so there will not be a breaking change. 
Emotion will append `px` if only a number is passed in. 

Screen Shot 2022-12-01 at 1 07 09 PM

## Testing

You can update one of these values in the Storybook story. 
I also tested this out using `npm link` in the UI. 

## Checklist

### DIFF
--- a/packages/popover/style.ts
+++ b/packages/popover/style.ts
@@ -17,7 +17,7 @@ export const menuList = css`
 `;
 
 // Hides unnecessary horizontal scrollbars on Linux and Microsoft
-export const hideHoriztonalScroll = css`
+export const hideHorizontalScroll = css`
   overflow-x: hidden;
 `;
 
--- a/packages/dropdownMenu/components/DropdownMenu.tsx
+++ b/packages/dropdownMenu/components/DropdownMenu.tsx
@@ -21,13 +21,13 @@ export interface DropdownMenuProps {
    */
   initialIsOpen?: boolean;
   /**
-   * Maximum height the menu can grow to
+   * Maximum height the menu can grow to. Can accept any number value and it will add `px` or any valid maxHeight value including vh, ems, or rems.
    */
-  menuMaxHeight?: number;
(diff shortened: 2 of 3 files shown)


vacas5:
should it fall back to a string instead of `number`? if someone passes `80` is there code that changes it to `80px`?

nataliepina:
> should it fall back to a string instead of `number`? if someone passes `80` is there code that changes it to `80px`?

I tried to follow the trail of this yesterday and it seems something to do with props being passed between components and how it's handled inside of the Popper and the Dropdownable. I couldn't quite find the exact piece of code setting the px. 
I left number because that does work and it does apply px, that is the previous functionality so I didn't want to cause a breaking change.
However, when I also allow the other valid CSS values as a type they do work as expected, it does not apply any px.
I tested this with `npm link` in the UI and it allowed me to enter both types of values successfully with no type errors.

packages/popover/style.ts
```diff
@@ -17,7 +17,7 @@ export const menuList = css`
 `;
 
 // Hides unnecessary horizontal scrollbars on Linux and Microsoft
-export const hideHoriztonalScroll = css`
+export const hideHorizontalScroll = css`
```
clintonmedbery, line 20: Took me a minute to see the difference 😂

clintonmedbery (approved):
This is good to me pending Russ' concerns.

vacas5 (approved):

PR #892, opened 2022-12-01
feat: Toast - add a custom timeout

seialkali:
# Description

This PR updates the Toaster component so that it can take a custom timeout value

**NOTE** I'm going to put the story updates for this component in a separate PR because I have questions about some funky behavior I'm seeing and I don't want it to hold up this insights toast ticket.

https://d2iq.atlassian.net/browse/D2IQ-94241

## Checklist

### DIFF
--- a/packages/toaster/stories/Toaster.stories.tsx
+++ b/packages/toaster/stories/Toaster.stories.tsx
@@ -106,6 +106,41 @@ export const AutoDismiss = () => {
   );
 };
 
+export const CustomDismissTimeAutoDismiss = () => {
+  const [toasts, setToasts] = React.useState<number[]>([]);
+
+  const removeToast = (id: number) => {
+    setToasts(toasts => toasts.filter(toast => toast !== id));
+  };
+
+  const handleToastAdd = () => {
+    const newToastId = addedToastId++;
+    setToasts(toasts => [...toasts, newToastId]);
+  };
+
+  return (
+    <div>
+      <Toaster dismissTime={5000}>
+        {toasts.map(toastId => {
+          const handleDismiss = () => {
+            removeToast(toastId);
+          };
+          return (
+            <Toast
+              autodismiss
+              id={`toast-${toastId}`}
+              key={toastId}
+              title={`New message ${toastId} disappearing in 5 seconds`}
+              onDismiss={handleDismiss}
+            />
+          );
+        })}
+      </Toaster>
+      <button onClick={handleToastAdd}>Make me toast</button>
+    </div>
+  );
+};
+
 export const WithDismissCallback = () => (
   <Toaster>
     {[
--- a/packages/toaster/components/Toaster.tsx
+++ b/packages/toaster/components/Toaster.tsx
@@ -5,17 +5,21 @@ import { ToastProps } from "./Toast";
 import { toaster, preTransitionStyle, transitionStyles } from "../style";
 import { margin, marginAt, listReset } from "../../shared/styles/styleUtils";
 
-export const DELAY_TIME = 3000;
+export const DEFAULT_DELAY_TIME = 3000;
 const MARGINAL_DELAY = 1000;
 const animationDuration = 300;
 
 type Toast = React.ReactElement<ToastProps>;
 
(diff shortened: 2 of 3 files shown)


packages/toaster/stories/Toaster.stories.tsx
```diff
@@ -45,6 +45,12 @@ export const Warning = () => (
   </Toaster>
 );
 
+export const Mute = () => (
```
vacas5, line 48: @nataliepina do you know if there is a way to combine all of these appearance stories into one and just let a user change the appearance in Storybook?

vacas5 (approved):
this is groovy. I wonder if you considered a custom icon prop. Seems like the design has gone beyond the parameters of the existing component library. My concern is codifying "mute" as an appearance for just this one use case, because we may not ever use it again anywhere else. Not a big deal, but maybe we should push back on design that for consistency of alerts, we should use one of the existing appearances like "success" and not try and customize the icon.

packages/toaster/stories/Toaster.stories.tsx
nataliepina: Yes @vacas5, you can combine all of the appearances by using controls similar to the Badge component story - https://dcos-labs.github.io/ui-kit/?path=/story/graphic-elements-badge--default

seialkali:
> this is groovy. I wonder if you considered a custom icon prop. Seems like the design has gone beyond the parameters of the existing component library. My concern is codifying "mute" as an appearance for just this one use case, because we may not ever use it again anywhere else. Not a big deal, but maybe we should push back on design that for consistency of alerts, we should use one of the existing appearances like "success" and not try and customize the icon.

@vacas5 Check the trade-offs section on my PR description.

vacas5:
okay I disagree that we should handle separately. If we add `mute` we can't remove it without it as a breaking change.

nataliepina:
I don't think we should add Mute as a separate appearance. It's not quite the same as something with semantic meaning such as success, danger, or warning. It's not very extensible and feels like more of a one-off for a feature vs something that should be integrated into the component library.
I'd suggest adding the mute icon next to the text in the `title` for now since it can take in an element. In the future, it may be helpful to allow at least an `iconStart` property to add any icon. 

Sidenote, we do need to fix up the story for this component to display the code snippet and some other cleanup could be done here.

vacas5:
can the hook for toasts go in separately? seems like a separate bit of work

packages/toaster/components/useAutoDismissToasts.tsx
```diff
@@ -0,0 +1,43 @@
...
+      title={toastProps.title}
+      description={toastProps.description}
+      appearance={toastProps.appearance}
+      onDismiss={() => removeToast(toastProps.id)}
+    />
+  ));
+
+  return [toasts, addToast] as const;
+};
+export default useAutoDismissToasts;
```
vacas5, line 43: I don't love how we abstract out the `Toast` because what if I want to configure those props a certain way.

vacas5: looking at this in code, is this providing much utility? I can see where a consumer of toaster would want to customize their state and how they handle it.

packages/utilities/dataUtilities.ts
```diff
@@ -0,0 +1,19 @@
...
+      r = (d + r) % 16 | 0;
+      d = Math.floor(d / 16);
+    } else {
+      r = (d2 + r) % 16 | 0;
+      d2 = Math.floor(d2 / 16);
+    }
+    return (c == "x" ? r : (r & 0x7) | 0x8).toString(16);
+  });
+  /* eslint-enable */
+};
```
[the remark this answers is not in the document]
clintonmedbery: We could import one, but no.

packages/toaster/components/useAutoDismissToasts.tsx
clintonmedbery: I would say it is a convenience method, and you can configure props to an extent through the function. I am open to suggestion but I feel like we will need some sort of method to create this short lived components one way or another.

PR #894, opened 2022-12-02
docs: add appearance control to toast story

seialkali:
# Description

This PR adds an appearance control to the Toaster story. Since the Toast story needs both the Toaster and Toast component, we had to add the Toaster component as a decorator in order to get the appearance control to work properly. However, it only works for the stories that have the Toast component(s) as a direct child/children of the Toaster component (or wrapped in a fragment). It does not work for the stories where there's maps or the Toast component is in an array. 

Since the appearance control will only work on some of the stories, I've installed the [@storybook/addon-controls](https://www.npmjs.com/package/@storybook/addon-controls) package so that I can disable controls on the stories which the appearance control will not work.

## Testing

Go to the Toaster story in story book and confirm that you can change the appearance control for Default, Description, and Multi Toast story. Also confirm that you can not use the controls for the rest of the stories.

## Checklist


package.json
```diff
@@ -77,6 +77,7 @@
     "@commitlint/config-conventional": "17.2.0",
     "@mdx-js/react": "2.1.5",
     "@storybook/addon-a11y": "6.5.13",
+    "@storybook/addon-controls": "^6.5.14",
     "@storybook/addon-essentials": "6.5.13",
```
nataliepina, line 81: The storybook addon-essentials includes Controls. Was there a reason why you had to install it separately?
https://www.npmjs.com/package/@storybook/addon-essentials

.storybook/main.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
         actions: false
       }
     },
-    "@storybook/addon-a11y"
+    "@storybook/addon-a11y",
+    "@storybook/addon-controls"
```
nataliepina, line 21: Removing controls should solve this but doing it this way also swapped the order where the Accessibility tab is first by default instead of Controls.

nataliepina (changes requested):

package.json
seialkali: Can you point me to an example where you've a control has been disabled for a single story? I wasn't able to find one. This is why I've installed this addon.

.storybook/main.js
seialkali: I'm sorry, I don't quite understand the the importance of the order. Can you please explain?

seialkali: Actually, I think I know what you mean now. In the situation where we're not able to do what I need without the controls addon, would swapping the order here change which tab is first?

package.json
nataliepina: This addon should be included already through the essentials, it's what gave us Controls at all in the first place. 

We don't have any examples of it being disabled for a single story like you are implementing but similarly, it has been used to disable certain controls for all of the stories such as this one in ToggleInputList:
`    selectedItems: {
      control: { disable: true }
    },
    `

.storybook/main.js
nataliepina: Yep the order here matters for which tab is displayed first in the addons pannel. Controls was before Accessibility because it is first in the addons array.

nataliepina: Also when a user lands on a story they should see the Controls addon open by default in the panel, since that is more important and what we're likely to use first compared to Accessibility. Hope that clarifies what I meant by the first comment.

package.json
seialkali: I don't want to disable controls for all stories. I want to disable controls for just some of the stories. I tried it without the addon and it didn't work. If you think it can be done in another way without the addon, please let me know what that way is.

seialkali: I just wasn't work at all, it wouldn't disable the controls. I've tried uninstalling it now, and it's working 🤷‍♀️

nataliepina (approved):
Nice very clean implementation. Glad you've introduced this pattern to disable controls for specific stories and we can use it elsewhere.

PR #908, opened 2022-12-13
fix: make var check ssr friendly

vacas5:
document and window are not defined in node environments, so trying to use ui-kit with an isomorphic framework like next will fail saying document/window is not defined. Adding a check that window is not undefined before trying to operate on document seems to fix the issue. There may be more places like this in ui-kit but this is all I've run up against so far.

# Description

## Checklist


mstruebing (approved):
I bet there are a lot of other places in here which aren't ssr friendly.

PR #916, opened 2023-01-12
[1] refactor: add data-cy and classname props

nataliepina:
# Description

Throughout these changes I am looking to achieve the following:
- Add `data-cy` attributes that we will be able to pass in custom values. Improving DX with our components + Cypress testing. I have opted for consistent use of `"data-cy"` instead of `["data-cy"]` as the [] is unnecessary. 
- Add `className` to an outermost container on components, that we can use for style overrides in the future.
- Consistency in the casing of prop definitions, as this displays to the Storybook UI in the Docs tab.
- Remove unnecessary `defaultProps` and opt for default values set during prop destructuring. 

## Checklist

### DIFF
--- a/packages/accordion/components/AccordionItem.tsx
+++ b/packages/accordion/components/AccordionItem.tsx
@@ -5,13 +5,16 @@ import { Provider as AccordionItemProvider } from "./AccordionItemContext";
 
 export interface AccordionItemProps {
   /**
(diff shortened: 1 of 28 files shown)


mstruebing (approved):
👍

seialkali (approved):

PR #923, opened 2023-01-26
feat: list tag prop is optional

nataliepina:
The tag property has always had a fallback/default value of "ul". Because of this, the property should be optional to set. Moving the from defaultProps to using fallbacks triggered this requirement.

# Description

## Checklist


vacas5 (approved):
do you want this to be a `fix` or `feat` so you can bump?

nataliepina:
> do you want this to be a `fix` or `feat` so you can bump?

It was originally a fix since it was a bugfix ticket but good call, I think it's more of a feat at this point anyway 😄

PR #936, opened 2023-03-03
fix: fix broken alias transformer and update design token docs

seialkali:
# Description

When style-dictionary was upgraded from 2.x.x to 3.x.x the data structure of the design token changed, which meant that the alias transformer was using the value of the token rather than the object path in order to generate the alias.

With the transformer fixed, this PR also updates the design token docs.

https://d2iq.atlassian.net/browse/D2IQ-95943

## Checklist


nataliepina (approved):
Looks like this did the trick! Thanks for getting to the bottom of this.

Issue #926, opened 2023-02-07
Refactor the use of `defaultProps` to instead use fallback default values


nataliepina:
# Description

In older versions of React, we had an explicit defaultProps property we could apply to components. 

We still have some references to defaultProps throughout UI Kit that should be updated to use fallbacks like the other components. This may require some component refactoring. 

For example:
```
const HeadingText1 = (props: HeadingTextProps) => (
  <Text weight="medium" size="xl" data-cy="headingText1" {...props} />
);

HeadingText1.defaultProps = {
  align: "inherit",
  color: themeTextColorPrimary,
  wrap: "wrap",
  tag: "h3"
};
```

can be refactored to:
```
const HeadingText1 = ({
  align = "inherit",
  color = themeTextColorPrimary,
  wrap = "wrap",
  tag = "h3",
  ...props
}: HeadingTextProps) => (
  <Text weight="medium" size="xl" data-cy="headingText1" {...props} />
);
```

## Acceptance Criteria

Update anywhere that `defaultProps` is used and opt for fallback defaults set where props are destructured. 

<hr> 

## How to Contribute 
Please review the [contributing documentation](https://github.com/dcos-labs/ui-kit/blob/main/CONTRIBUTING.md) beforehand. 

Steps to Contribute:
1. [Fork the repository](https://github.com/dcos-labs/ui-kit/fork).
2. Clone it to your local system.
3. Create a new branch.
4. Make your changes. Make sure that tests are still passing. To update snapshot tests, run `npm run test -- -u` and include the updated tests in your commit.
5. Commit your changes.  When committing prefix your commit message with the appropriate conventional commit type. For example, `refactor: <component_name> tests`. 
6. Push up your changes. In GitHub, click the compare & pull request button to begin drafting your pull request.
7. Fill out the pull request form, and it will be reviewed by the UI Team.

derekhassan:
Hi @nataliepina, I'd love to take this issue!

nataliepina:
Assigning it over to you, thanks for your interest in contributing again Derek! 🎉

nataliepina:
Hi @derekhassan how is this task going? Please let us know if you need any help with it!

derekhassan:
Hi @nataliepina, thanks for reaching out! Apologies for the delay, I haven't had much time to work on this issue. Things are going good so far and I expect to have a PR up by Saturday 👍

PR #937, opened 2023-03-04
refactor: change default props to use default values

derekhassan:
# Description

This changes all uses of `defaultProps` to use default parameters. The following components have been changed:
- `PopoverListItemAvatar`
- `PopoverListItemIcon`
- `DangerText`
- `HeadingText1`
- `HeadingText2`
- `HeadingText3`
- `InteractiveText`
- `MonospaceText`
- `SmallText`
- `SuccessText`
- `Text`
- `WarningText`
- `TextInput`
- `TextInputWithBadges`
- `TextInputWithButtons`
- `TextInputWithIcon`

I also changed a few prop interface properties to optional as they have default parameters specified in their respective components.

#926 

## Testing

Storybook and run tests

## Checklist

- [X] I have reviewed the changes and provided detail to the sections above


packages/textInput/components/utils.tsx
```diff
@@ -78,7 +78,7 @@ const getInputElement = (
 };
 
 const getInputAppearance = (props: TextInputProps): string => {
-  return props.disabled ? "disabled" : props.appearance;
```
derekhassan, line 81: Is there a better solution for this? WIthout the `!`, I get the following TypeScript warning:
```
Type 'string | undefined' is not assignable to type 'string'.
```

packages/popover/components/PopoverListItemAvatar.tsx
```diff
@@ -1,20 +1,8 @@
 import * as React from "react";
 import Avatar, { AvatarProps } from "../../avatar/components/Avatar";
 
-interface PopoverListItemAvatarProps extends AvatarProps {
-  /**
-   * Which side of the menu item the avatar appears on
-   */
-  position?: "start" | "end";
-}
```
nataliepina, line 9: Was there a reason you removed this interface?

It looks like we still want this here. The component will use the `position` prop which does not exist on `AvatarProps`. 

To maintain backward compatibility with the `defaultProps`, we'll want to make sure we set a matching fallback for position `position = "start"`

packages/popover/components/PopoverListItemIcon.tsx
```diff
@@ -1,20 +1,8 @@
 import * as React from "react";
 import Icon, { IconProps } from "../../icon/components/Icon";
 
-interface PopoverListItemIconProps extends IconProps {
-  /**
-   * Which side of the menu item the icon appears on
-   */
-  position?: "start" | "end";
```
nataliepina, line 8: Same thing here with the default value for the position prop on this component as the `PopoverListItemAvatar`.

packages/textInput/components/utils.tsx
nataliepina: Great question! We generally try to avoid using `!` in our codebase.

There are [several methods](https://bobbyhadz.com/blog/typescript-argument-type-undefined-not-assignable-parameter-type-string) to fix this. 
I might suggest using some nullish coalescing here, like this:
```
const getInputAppearance = (props: TextInputProps): string => {
  return props.disabled
    ? "disabled"
    : props.appearance ?? InputAppearance.Standard;
};
```

If appearance is undefined or null, explicitly set the fallback to the default defined in `TextInputProps`.
```
  /**
   * Sets the current appearance of the input component. This defaults to InputAppearance.Standard, but supports `InputAppearance.Error` & `InputAppearance.Success` appearances as well.
   */
  appearance?: InputAppearance;
   ```
   
It might be even more readable to use an if statement in there instead. But I'll leave that up to you!

```
const getInputAppearance = (props: TextInputProps): string => 
{
  if (props.disabled) {
    return "disabled";
  }
  return props.appearance ?? InputAppearance.Standard;
};
```

I hope that helps as another option here!

**Comment by nataliepina** [comment]
I am wondering if this should be a `feat:` commit instead of a `refactor:` since we are changing some props from required to optional and it could change how we work with these components downstream. Any thoughts @d2iq/ui-team ? 🤔

**Comment by mstruebing** [comment]
Technically changing the API is a breaking change but as we change it from required to optional you can still use the API as before. So it shouldn't be a breaking change.

If we change the functionality this should be a `feat` commit but as I think everything will work exactly as it did before a `refactor` is fine in my opinion.

File: packages/popover/components/PopoverListItemAvatar.tsx (hunk 2 above)
Comment (derekhassan) [reply packages/popover/components/PopoverListItemAvatar.tsx:9]: Hey @nataliepina, thanks for mentioning this. I removed this interface as I was getting a TypeScript error when passing the `position` prop like this as `position` was not a defined prop for `Avatar`:
```js
const PopoverListItemAvatar = ({
  size = "xs",
  position = "start",
  ...props
}: PopoverListItemAvatarProps) => (
  <Avatar size={size} position={position} {...props} />
);
```

The error was:
```
Property 'position' does not exist on type 'IntrinsicAttributes & AvatarProps'
```
I had assumed the prop `position` was not required to pass here since `Avatar` didn't seem to use it, but apologies for my incorrect assumption 😬

Since the `Avatar` component doesn't have a `position` prop defined, how should this be handled? Should I include it as part of the props spread?

packages/textInput/components/utils.tsx
derekhassan: Hey @nataliepina, thanks so much for this explanation! This was super helpful for me 🙌

packages/popover/components/PopoverListItemAvatar.tsx
nataliepina: No worries, I can totally see where you'd be coming from there, and trying to clean it up as it doesn't initially seem like it's used with the props spread and encountering that error.

It also can be tricky if you don't know these components well as some components are used by others. The `PopoverListItemAvatar` and `PopoverListItemIcon` are used in our `DropdownMenu` component. This is where it could take in a start or end value for the position.
Check out [this component story](PopoverListItemAvatar ) and under the Docs tab if you click "Code" you'll see the example. 
Here is a screenshot of that:
Screenshot 2023-03-08 at 9 57 10 AM

I think as you suggested you could let the props spread handle it as a simple solution. We may want to consider adding position to the prop types for Avatar and Icon in the future but I'm sure we're there yet.

nataliepina: No problem! Once you update this and pull in the changes from main it's a ✅ for me! Thank you for all of your effort on this update.

derekhassan: Hey @nataliepina, I've run into an issue when testing these components using default values instead of `defaultProps`. I noticed that if no `position` prop was passed down to `PopoverListItemAvatar` or `PopoverListItemIcon`, the following variable `itemGraphicStart` in `PopoverListItem` is undefined:
```js
      const itemGraphicStart = itemChildren.find(
        child => isItemGraphic(child) && child.props.position === "start"
      );
```

However if I add a `defaultProps` value for `position`, `itemGraphicStart` correctly finds the item and it works as expected. Would you have some guidance on what could be causing this issue? Is it because React doesn't know what the default values would be but it can determine the `defaultProps`?

I did play around with it a little bit and I found that changing `itemGraphicStart` to the following resolved the issue:

```js
      const itemGraphicStart = itemChildren.find(
        child =>
          isItemGraphic(child) &&
          (child.props.position === "start" || !child.props.position)
      );
```

Let me know your thoughts on this!

nataliepina: This is a pretty tricky one with those functions to set `itemGraphicStart` and `itemGraphicEnd`. I've tested some different methods to avoid having to update these functions, but I think the implementation you suggested will be fine and it does work. Great detective work on this.

nataliepina (approved):
Thanks for making the changes and all of your hard work here.

derekhassan:
Happy to help! Thanks again for the feedback and tips, I learned some new things! 🙌

packages/textInput/components/TextInputWithButtons.tsx
```diff
@@ -25,7 +25,12 @@ export interface TextInputWithButtonsProps
...
 }
 
-const TextInputWithButtons = (props: TextInputWithButtonsProps) => {
+const TextInputWithButtons = ({
+  type = "text",
+  appearance = InputAppearance.Standard,
+  showInputLabel = true,
+  ...props
+}: TextInputWithButtonsProps) => {
   const [hasFocus, setHasFocus] = React.useState(false);
```
nataliepina, line 34: Sorry @derekhassan we've actually noticed another issue with this after some downstream testing. 
There is some broken functionality here. 

If you compare this story in this PR:
Screenshot 2023-03-17 at 2 15 10 PM

You will notice it is missing that icon start in this component. It should look like this:
Screenshot 2023-03-17 at 2 16 34 PM

packages/textInput/components/TextInputWithButtons.tsx
```diff
@@ -86,12 +95,23 @@ const TextInputWithButtons = (props: TextInputWithButtonsProps) => {
                 getInputAppearanceStyle(inputAppearance)
               )}
             >
-              {getIconStartContent(props, getInputAppearance())}
+              {getIconStartContent(
+                {
+                  type,
```
clintonmedbery, line 100: I think this is missing some props. At least for `iconStart`, as the icon is no longer showing up in storybook.

packages/textInput/components/TextInputWithButtons.tsx
```diff
@@ -49,13 +54,17 @@ const TextInputWithButtons = (props: TextInputWithButtonsProps) => {
...
-    return props.appearance;
+    return appearance;
   };
 
   const getInputElementProps = () => {
-    const baseProps = getBaseInputElementProps(props);
+    const baseProps = getBaseInputElementProps({
+      type,
+      appearance,
+      showInputLabel
```
clintonmedbery, line 66: I believe this is missing the placeholder props. Not sure if it missing anything else like hint content.

clintonmedbery:
Lots of great work here! I found a few places that are missing some props around the `TextInputWithButtons` component. Hopefully my comments are helpful and great work!

packages/textInput/components/TextInputWithButtons.tsx
```diff
@@ -126,18 +146,22 @@ const TextInputWithButtons = (props: TextInputWithButtonsProps) => {
...
-        appearance,
-        hidden: !props.showInputLabel,
-        id: getId(props),
+        appearance: calculatedAppearance,
+        hidden: !showInputLabel,
+        id: getId({
+          type,
+          appearance,
+          showInputLabel
+        }),
```
seialkali, line 164: This function is looking for a prop of `id` so I think what we need to pass in `props` here instead.

```suggestion
        id: getId(props),
```

packages/textInput/components/TextInputWithButtons.tsx
derekhassan: Ah, apologies for the broken icon, it looks like I had removed the extra non-destructured props from `getIconStartContent` 😅. I looked into the issue about being unable to change the appearance and it seems to be an issue with the `TextInputWithButtons` story. I think this behavior is related to [this Storybook issue](https://github.com/storybookjs/storybook/issues/17025). If I move the buttons array outside the function, the issue goes away. I'm not sure if this is the ideal solution and was wondering if you had any feedback on this approach? (This is in [TextInputWithButtons.stories.tsx](https://github.com/d2iq/ui-kit/blob/main/packages/textInput/stories/TextInputWithButtons.stories.tsx))

```jsx
const buttonsPlaceholder = [
  <TextInputButton
    key={0}
    color="red"
    shape={SystemIcons.Close}
    onClick={btnClickFn}
    aria-label="Clear input"
  />
];

export const WithCustomIcons = args => {
  const { buttons, ...rest } = args;
  return (
    <TextInputWithButtons
      id="withIcon.colored"
      inputLabel="With colored icon"
      iconStart={<Icon shape={SystemIcons.Search} />}
      buttons={
        buttons && Object.values(buttons).length ? buttons : buttonsPlaceholder
      }
      {...rest}
    />
  );
};
```

nataliepina: It looks like some of these issues may be resolved with Storybook 7 which we plan to upgrade to soon. 
Go ahead and fix the other issues with the component and its props and if there is still an issue with the appearance, don't worry about it. I'll do some research and see how I can resolve that. It's more important that the component functions as expected and I have seen some issues myself with the controls for the stories and plan to improve those in the near future. Thanks for your research on this matter!

derekhassan: Gotcha, thanks!

packages/textInput/components/TextInputWithButtons.tsx
derekhassan: Ah yep, this should be resolved now 👍

packages/textInput/components/TextInputWithButtons.tsx
derekhassan: Good call, thanks for the guidance! 🙌

nataliepina (approved):
I did some further regression testing for the story and I tested it downstream.

mstruebing (approved):

PR #939, opened 2023-03-07
fix: shape of progressbar and avatar

nataliepina:
There are some issues with borders set as % vs px.

# Description

This [issue](https://stackoverflow.com/questions/18794947/capsule-shape-using-border-radius-without-a-set-width-or-height) basically sums up the fix with percentage-based border-radius for a pill/capsule shape.
The Avatar shape issues were less obvious as they did not appear in the story but they will be affected in the UI after some testing, reverted back to 25% as a fix. 

Added a custom data-cy for the progress bar also. 

## Checklist

### DIFF
--- a/packages/avatar/style.ts
+++ b/packages/avatar/style.ts
@@ -1,13 +1,10 @@
 import * as React from "react";
 import { css } from "@emotion/css";
-import {
-  borderRadiusLarge,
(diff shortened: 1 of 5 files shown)


clintonmedbery (approved):
Nice!

PR #951, opened 2023-03-14
fix: remove engines

mstruebing:
# Description

This removes the engines from our `package.json` because it actually restricts the consumer to only use the node version matching. We are providing a frontend library that doesn't export NodeJS code to the consumer, so we should not expect any specific NodeJS version at all.

I think we set this up when we ran into peer dependency issues, which I hope are not the case anymore and should be fixed as soon as possible whenever they occur.

An alternative would be to provide a bigger version range but we would need to manually update this constantly.

## Checklist


nataliepina (approved):
Ahh okay. I think we did add this to help resolve those peer dependency issues but this should be fine to remove now. Good call.

PR #953, opened 2023-03-16
feat: node 18 support

nataliepina:
Closes D2IQ-96159

# Description
Now that we've updated to Storybook 7, theoretically, this should be good to go now.

To break down some of the changes further. This should mimic what we have in kommander-ui where we are able to set up a single file to reference the node version we want to use. It will be easier to update the node version by only having to update this one location. 

I've updated the actions to use a matrix testing strategy.  This configuration creates three parallel jobs, one for each value specified in the node-version array. Each job runs with a different version of Node.js: 16 and 18. The matrix strategy is useful when you want to run the same job with different configurations or environments. In this case, it allows the action to be tested with multiple Node.js versions to ensure compatibility and verify the behavior across different runtime environments.

## Testing
We'll probably need to merge this. 

## Checklist


nataliepina:
I think this will fail until we fix the gh actions using build-storybook, but this can still be reviewed. It may block this from merge. We have a task related to fixing this in the sprint.

nataliepina:
> I think this will fail until we fix the gh actions using build-storybook, but this can still be reviewed @d2iq/ui-team. It may block this from merge. We have a task related to fixing this in the sprint.

.node-version
```diff
@@ -0,0 +1 @@
+18.15.0
```
mstruebing, line 1: Should we use 18.16.0?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16.13.0-slim
+FROM node:18.15.0-slim
```
mstruebing, line 1: Do we even use this `Dockerfile` anywhere? Can we remove this?

.node-version
nataliepina: I wanted to match to what we have in kommander-ui. Should we update it there if I move this to the latest?

Dockerfile
nataliepina: 🤔  Let me check and i'll remove it if not.

nataliepina: It's used by Uffizzi.

.node-version
nataliepina: Updated this to 18.16.0

.github/workflows/release.yml
```diff
@@ -5,11 +5,14 @@ on:
...
+    strategy:
+      matrix:
+        node-version: [16, 18]
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v2
         with:
-          node-version: "16"
+          node-version: ${{ matrix.node-version }}
```
mstruebing, line 15: Isn't one node version enough for release?
I would suggest using the one from `.node-version`

```suggestion
          node-version-file: ".node-version"
```

And removing line 8-10.

nataliepina: @clintonmedbery suggested the matrix for testing. Clint do you have any thoughts on the benefits of the matrix vs a single version?

mstruebing: This is the `release` workflow, not the testing workflow :)

clintonmedbery: Yeah I like the idea of testing on multiple versions (if that is what we want to do) but I think the release only needs one version.

clintonmedbery (approved):
Nice work!

.github/workflows/ci.yml
```diff
@@ -6,11 +6,14 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16, 18]
```
mstruebing, line 11: ❓ Why do we want to run the CI pipeline with node 16 and 18 when we use node 18 in development and we do not even publish a node package but a react component library?

nataliepina: I'm just going to remove this. I'd like to get this merged in before i'm off.

mstruebing (approved):

PR #956, opened 2023-03-24
feat: add footerBar prop to AppChrome

nataliepina:
# Description

I've updated the header and footer sections to use their semantic element counterparts. 

I tested this downstream with the custom banners, screenshot below.

## Testing
View the first AppChrome story.
Add a chunk of Lorem Ipsum or something to the main and set a fixed height on it to see how it will work with scroll.
Even better, test this downstream with the custom banners. 

## Screenshots
Screenshot 2023-03-24 at 11 17 11 AM
Screenshot 2023-03-24 at 11 24 58 AM

## Checklist

- [x] Significant changes have been tested downstream to avoid breaking changes

### DIFF
--- a/packages/appChrome/components/AppChrome.tsx
+++ b/packages/appChrome/components/AppChrome.tsx
@@ -20,27 +20,31 @@ import {
 
 export interface AppChromeProps {
   /**
-   * Allows custom styling
+   * Allows custom styling.
    */
   className?: string;
   /**
-   * JSX to render along the left side
+   * JSX to render along the left side.
    */
   sidebar?: React.ReactNode;
   /**
-   * JSX to render as the navigation bar at the top
+   * JSX to render as the navigation bar at the top.
    */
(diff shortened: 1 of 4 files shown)


packages/appChrome/components/AppChrome.tsx
```diff
@@ -87,6 +92,7 @@ const AppChrome = ({
             {children}
           </main>
         </div>
+        {footerBar && <footer data-cy="footerBar">{footerBar}</footer>}
```
vacas5, line 95: maybe we shouldn't introduce a footerBar here if we're arguing above that headerbar is deprecated?

nataliepina: headerBar is not deprecated, that is mainContent as children was introduced

vacas5: that makes way more sense :)

PR #962, opened 2023-04-19
refactor: TextInputWithButtons story

nataliepina:
# Description
This fixes the story controls so that it doesn't freeze when you change the appearance.

I've refactored this again as I've discovered the intended behavior here is that the  `iconStart` will take on the color of the input. However, the button can take on its own color. I've added a comment to clarify this for the future. 

## Checklist


mstruebing (approved):
Looks good :)

PR #971, opened 2023-05-12
docs: update stories to access all system icons

nataliepina:
# Description

Previously stories that have icon-related controls only had access to a set amount of system icons. With this method, these stories can access all of the system icons. This gives us a better method to test all of the possibilities and is less maintenance. This is also especially helpful for the Icon story so we can test out new icons to make sure they are accurately taking in props like size and color. 

I also included a typo fix for the ToggleBox appearance control. 

## Testing
The stories should all function as before, except now there are more options for icons. 

## Screenshots
Before
Screenshot 2023-05-12 at 12 20 30 PM

After
Screenshot 2023-05-12 at 12 21 04 PM

## Checklist


mstruebing (approved):
This is way better!

PR #972, opened 2023-05-25
Storybook 7

nataliepina:
# Description

These changes introduce the migration from Storybook 6.5 to Storybook 7. 

**Issues Encountered** 

There was a breaking change for the Welcome page. Storybook's styles now will override all story files, including these MDX files. This broke our theming and styles for this page. I discovered a workaround in a community issue. We can use `Unstyled` to remove any storybook style overrides.   Also, for theming to work correctly, it will now come from "@storybook/theming/create". 

The favicon broke when upgrading to storybook 7. I chose a new method for us to integrate the favicon with the manager-preview.html file. The favicon won't appear locally but it will appear in the production build. You can test this out by running `npx http-server storybook_static`.  The favicon as an svg is the more modern way to manage this. 

The codemod took care of the migration of our stories from CSF2 to CSF3. There were a few stories that were broken and required tweaks. The decorators for the Toast component broke the Docs page. And you would see this error - `Error: Rendered more hooks than during the previous render.` I fixed this by refactoring the stories so that Toaster will wrap around the Toast component within the individual stories. It doesn't need to be a decorator.

There were a few other locations that had render errors. To fix this I had to slightly adjust the formatting of the story. PageHeader is an example of a story that I had to adjust.  There also were some errors in the code due to props being overridden in stories. Infobox required a slight refactor to get the message to render.

**Other Notable Changes**

With these changes, I was also able to migrate us to MDX2!  This only impacts the Welcome page, Typography, and Colors. 

I wanted the changes to isolate the update to Storybook 7 primarily so that it’s easier to review. I did a side-by-side comparison of the component docs. I’ve noted a few locations where improvements could be made as a follow-up.

These changes have also been tested downstream.

- https://jira.d2iq.com/browse/D2IQ-92603

## Testing
1. Make sure to run an `npm install`
2. Build storybook `npm run build:storybook`
3. Start Storybook `npm run start`
4. Perform a thorough regression test. 

## Trade-offs
It takes a few seconds longer for storybook to start up. I believe we can improve this by refactoring our webpack configuration and reduce/refactor our stories. Switching between stories seems to be more performant though.

## Screenshots

Screenshot 2023-05-30 at 12 29 26 PM

## Checklist

- [x] Significant changes have been tested downstream to avoid breaking changes


.storybook/webpack.config.js
```diff
@@ -8,9 +8,6 @@ module.exports = ({ config }) => {
   config.resolve.extensions = config.resolve.extensions.concat(
     webpackBase.getExtensions()
   );
-  config.resolve.alias["core-js"] = path.dirname(require.resolve("core-js"));
-  config.module.rules[0].use[0].options.sourceType = "unambiguous";
```
nataliepina, line 12: These items are no longer required, and they will not allow Storybook to build.

seialkali (approved):
Looks good! Great work persevering with all the issues you came up against and finding solutions 💥

PR #982, opened 2023-07-24
feat: new z-index design tokens

nataliepina:
Introduces new z-index tokens and adjustments for pre-existing tokens.

BREAKING CHANGE:
The z-index values for zIndexModal and zIndexDropdownable have increased.

Closes D2IQ-92067

# Description

**Improved Z-Index Design Tokens**

**Intro**

These changes introduce a new and refreshed set of z-index design tokens with the aim of enhancing the token system for future use in our design system. With these changes, we can ensure a more consistent and organized stacking order for our elements and avoid bugs.

**The Problem**

There are several challenges with our current set of z-index design tokens. One issue is that there are not enough options. 
We don't have enough values to cover other common use cases. There is no option to place anything along the z-axis anywhere between a modal and a dropdown. The values are very low which can be problematic. For example, modals are something that should always be very high, and stacked upon all other items our modal is set to a z-index of 1 which is low and more commonly encountered. We have conflicts with components from external libraries because of these chosen values. These conflicts can result in components overlapping inappropriately or appearing behind others, causing usability issues.

Having more options will help with organization and present an easier set of options for engineers without any compromises. Better semantic token options will be beneficial to make decisions on which tokens to use. So rather than using the `zIndexModal` token for an overlay on a table we can use `zIndexOverlay`. More options can help to prevent the use of arbitrarily large numbers at random when an important override is needed.

**The Solution**

To resolve the z-index conflicts and improve the token system, I conducted research on other design systems to identify common practices for z-index management. I found that many systems utilize significantly higher z-index values for their custom components, often ranging from the 100s to the 1000s.

After careful consideration, I decided to adopt a more conservative approach by opting for lower z-index values in the 100s instead of the 1000s. It's worth noting that precise values may not be as critical as having a well-organized and structured hierarchy. I spoke to our design team on this and we aligned these values and the stacking order for our components. 

**Current Z-Index Usage**

Our design system makes limited use of the z-index. This is a good, because we will not have to make updates in many places. The less z-index is used the less we'll encounter conflicts between usage, and the easier it is to keep track of the areas that can be impacted by applying z-index to a component. We primarily utilize it for overlays, modals, dropdownable components, and a few other specific scenarios in the UI. Despite the limited use, it's essential to have a well-defined and consistent z-index system to ensure proper stacking of elements. 

**Establishing a Z-Index Hierarchy**

To address the z-index conflicts and establish a clear hierarchy, I collaborated with the design team to analyze how these components are currently used in our application. We identified the key components that should take precedence over others in terms of stacking order. For instance, we recognized that tooltips and dropdowns could be contained within modals therefore they should use a higher z-index value.

I also hoped to do some "future-proofing" with these changes. For example, I wanted to introduce the "deep" token that is set to -1 even though we do not have an immediate use for it. Using `z-index: -1` is generally reserved for special cases where you want an element to appear behind all other elements on the page. It positions the element behind the default stacking order of the page content, including the background.

I introduced the "sticky" token which could be useful for sticky/fixed elements, and generally as an option for something higher up that can live between the base level and an overlay. 

The overlay was introduced for (you guessed it) overlays! We have a specific use case for this in kommander-ui with the insights table that has the loading overlay.

I kept some of our previous tokens with the same names so there would be fewer code changes, and it will be less of a breaking change in turn. 

**A Note on the Dropdownable Component Z-Index**

Presently, any component that includes the Dropdownable component will inherit the same z-index value. This means that our Tooltip, DropdownMenu, and Popover share the same z-index. I considered creating separate tokens for these items, however, I think it is reasonable for them to be positioned at the same height. There shouldn't be a need to display a Tooltip within a Dropdown or a Dropdown within a Tooltip. These components generally shouldn't have collisions. If we do encounter any issues in the future we can easily introduce higher values beyond `zIndexDropdownable`.

Would love to know if anyone has any thoughts or feedback on this new system.

 - https://jira.d2iq.com/browse/D2IQ-92067

## Testing
View the stories for components that use z-index (Modal, Dropdownable, DropdownMenu, Popover, Tooltip).

It would be best to bring these changes into the UI for a more thorough test.

## Trade-offs
I considered updating the Modal component's "scrim" to use the zIndexOverlay design token, however, because it is position fixed, it will overlap the Modal. This is due to the construction of this component. It would take a greater refactor of the ModalBase component to sort out a better solution for this and would be best to do this in isolation.

## Screenshots
Example of the `CodeEditor` component partially overriding the `Modal`.
image

## Checklist


nataliepina:
@d2iq/ui-team I'd like 2 approvals on this one due to the impact. I want to make sure everyone is on board with these changes.

clintonmedbery (approved):
Thanks for putting so much thought into this. This isn't necessary here but if we continue to have trouble another option is to add a prop that allows you to adjust the z-index. They do this in ant design. https://ant.design/components/modal#api

nataliepina:
> Thanks for putting so much thought into this. This isn't necessary here but if we continue to have trouble another option is to add a prop that allows you to adjust the z-index. They do this in ant design. https://ant.design/components/modal#api

We do have the `className` prop available on most components to override with custom styles which could achieve similar.

PR #986, opened 2023-08-16
Updated Dependencies / Major Prettier update

nataliepina:
# Description
Doing some prep work to make our transition away from fixed versioning successful. 

I've updated to all of the minor and patch versions available here. Then I updated prettier and it's related dependencies. This is a major update, it seemed to trigger some newly requested formatting for the style files. I reviewed all of the requested changes from prettier and they all seemed valid so I used the format command to update them all. Then it needed an update to the tests to pass since those files were reformatted, you can see that it's mostly an updated class for those, and not any actual style changes. 

I tried to update prettier a few months back and ran into some issues. It looks like they resolved those issues. For context, I believe this was the issue I encountered - https://github.com/azz/pretty-quick/issues/164 

## Testing
Make sure everything works as expected. 

## Checklist


vacas5 (approved):

clintonmedbery (approved):
Thanks for keeping up with this!

package.json
```diff
@@ -46,7 +46,7 @@
   ],
   "typings": "dist/packages/index",
   "dependencies": {
-    "@emotion/css": "11.11.0",
+    "@emotion/css": "11.11.2",
```
mstruebing, line 49: Why not start switchting to version ranges? :)

nataliepina: That's the next piece for me. I'd rather do this first and it's less messy so I can isolate it to only version ranges and not upgrades with that too.

PR #990, opened 2023-08-24
feat: tooltipicon component

nataliepina:
This introduces a TooltipIcon component that can be used as a trigger. It will help with consistent tooltip icon usage.
There are new docs that outline how to use the icons.

Closes D2IQ-97013

# Description

While testing out some Tooltip updates in the UI, I realized we are using the Icon component in many places to repeatedly with the same values as a Tooltip trigger. I first thought to create a reusable atom component for this in the UI, which made me come to the realization that this should live in UI Kit with it being many times throughout the app.

Creating a component for this will also help to define the intended use and alleviate some of the consistency issues we've encountered with this icon. 

Screenshot 2023-08-24 at 9 49 23 AM

In the UI, we discovered there has been an inconsistent use of icon color for the tooltip icons. The product design team and I compared color options. We decided on `greyLightDarken1` color for the level of color contrast while still having a nice appearance.

The only variable in it's use should be to choose between the (?) and (i) icons. I've included some documentation to serve as a reference when choosing between the two options. This information is based on what is outlined in the story along with some additional details I've included. 

 - https://jira.d2iq.com/browse/D2IQ-97013

## Testing
View the new story for the tooltip icon. 

## Screenshots
Screenshot 2023-08-24 at 8 54 21 AM

Screenshot 2023-08-24 at 10 01 19 AM

## Checklist

### DIFF
--- a/packages/tooltip/README.md
+++ b/packages/tooltip/README.md
@@ -19,3 +19,13 @@ Tooltips provide additional information or clarification about a UI element when
 
 - Use hint text below the field to provide information about what input is expected.
 - Use a Tooltip to provide additional explanation about the field. Usually, Tooltips are triggered in form fields by an icon in the field's label.
+
+### Using Icons with Tooltips
+
+In the UI we use two distinct icons for tooltips: ( i ) and ( ? ). We have defined the context of each icon, describing when to use them to ensure clarity and consistency in their usage:
+
+( i ) Icon:
+The ( i ) icon is utilized to provide users with a definition of a specific term. These icons are often associated with information includes additional details, explanations, or contextual information about a specific element, feature, or term. This aids in enhancing user comprehension by offering explanations directly within the context. This is the default for the `TooltipIcon` component as it is most commonly used.
+
+( ? ) Icon:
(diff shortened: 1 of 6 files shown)


packages/tooltip/README.md
```diff
@@ -19,3 +19,13 @@ Tooltips provide additional information or clarification about a UI element when
 
 - Use hint text below the field to provide information about what input is expected.
 - Use a Tooltip to provide additional explanation about the field. Usually, Tooltips are triggered in form fields by an icon in the field's label.
+
+### Using Icons with Tooltips
```
mstruebing, line 23: ❓ Should we make it possible to use any icon? Just because it's still an open source library and other people might would want to use these differently.
Additionally, if we ever decide to use other icons we do not need to change the code here.

nataliepina: I did consider this but tooltip icons are really only ever these two options. This is where it helps with the consistency part. Sometimes it's good to limit options to help consumers make the right decisions. If you wanted a completely different icon you could just pass in the `Icon` component into the trigger and i'm not sure how that would make this component much different than using that alone.

mstruebing (approved):

clintonmedbery (approved):
Tested and LGTM!

PR #993, opened 2023-08-25
feat: data-cy for dropdownmenu

nataliepina:
# Description

We need a data-cy for this component for testing. 

If this component is already targeted with it's data-cy the user may have had to use "popover" since the component utilizes that library and doesn't have a data-cy of it's own. This makes it a breaking change as it could potentially break tests for others.

## Checklist


clintonmedbery (approved):
Thanks!

PR #997, opened 2023-09-11
refactor: migrate react-id-generator to useId

nataliepina:
Refactored use of nextId from react-id-generator.
Now for fallback ids we can use React's useId hook. Extra component refactoring was required for hook usage. For example, renderLabel was modernized into a component.

Closes D2IQ-95926

# Description

This is everything required to be able to fully remove `react-id-generator`. 
This library was used to generate fallback id's for some of our components when an id prop is not passed in, giving it a unique value to pair with a description of the component (for example, "copyTooltip-1"). 

With React 18 we have the `useId` hook. This will perform the same function as nextId from react-id-generator. 
Using what is built into React makes the most sense here. It also allows us to remove a dependency, which is always a win when the maintenance burden is decreased. 

With useId it will create a unique id that has a value surrounded by colons, the value has "r" plus a unique number. This can be helpful to denote that the ID is being generated from react. For example, `copyTooltip-1` could become `copyTooltip-:r1:` with useId. 

#### Implementation Details
There are some caveats with using useId.

It is a hook, so it needs to be used at the top level of a component.
For this reason, I refactored the `renderLabel` which is really just returning JSX for a label element. I turned this into a reusable component `InputLabel`. This is a more modern pattern. The label should display in the exact same way when used. 
We have several other locations that use these getter type functions to return reusable elements. I made a note of some of these as I'd like to get those refactored next. I want to isolate the changes in this work to focus on what's required to remove react-id-generator. 

Because useId is a hook it also cannot be used as a fallback while destructuring, which is how we had it set up in many places with nextId. I simply moved them into the top level of the component instead. Also since it's a hook you can't do a one liner with the conditional logic. I set another value to choose between the id from props and the fallback generatedId because it often had repeated use.

I took the opportunity to remove some TODO comments here because they are no longer needed. There are no type errors when removing the non-null assertion so I removed those. With that I did check that id's were being set, they were so I don't think this is required here and it's a very old TODO [referenced by this PR](https://github.com/d2iq/ui-kit/pull/489/files). 

- https://jira.d2iq.com/browse/D2IQ-D2IQ-95926

## Testing
Review the changes in the snapshot tests, this is a great indicator to see that we haven't impacted the style of any components, you should only see a change in the fallback IDs for some of them that now use an ID generated by useId. 

I performed a regression test for these components, it would be helpful to have another set of regression tests for them. 
Especially for the ones with a TextInput that uses the InputLabel component. 

## Checklist


clintonmedbery (approved):
Thanks for improving this!

packages/accordion/components/AccordionItem.tsx
```diff
@@ -22,16 +21,18 @@ export interface AccordionItemProps {
 const AccordionItem = ({
   children,
   "data-cy": dataCy = "accordionItem",
-  id = nextId("accordionItem-"),
+  id,
   className
 }: AccordionItemProps) => {
+  const generatedId = `accordionItem-${React.useId()}`;
+  const accordionItemId = id || generatedId;
```
mstruebing, line 28: No big deal, but this way we always generate an ID even though we pass in an `id`.
It would be a bit more performant if `generateId` is a function which is only executed when we call it - but I guess that's not easily possible with a hook.
We could try to build a custom hook that does that for us.
Or a simple solution may be something like:

```
  const accordionItemId = id || `accordionItem-${React.useId()}`;
```

But anyway, I don't think generating an id is that much of a computing overhead, so decide for yourself if you want to improve that or not.

mstruebing (approved):
Do you want this second commit in there?

packages/accordion/components/AccordionItem.tsx
nataliepina: I wanted to do that method with one line but it will cause this `React Hook "React.useId" is called conditionally. React Hooks must be called in the exact same order in every component render.` so that's why i've broken it out into 2 lines. I will consider the computing overhead a bit more today before merging, glad that it's not much of one.

PR #998, opened 2023-09-11
refactor: create shared text input components

nataliepina:
This updates getter functions returning elements,
to become their own reusable components.

# Description
This work includes a set of refactors for text input components. 
The goal of this work is to refactor the use of shared utils/components.

I want to modern the code by migrating these getter functions to use components when their purpose is to return JSX. I created a shared folder that can house these components. The overall effort is to get rid of more of these `get<Something>` functions. I'd like us to refactor the getHint and getValidationErrors in another set of work. 

The `getInputContent` seems to have no reason to me to have separated out instead of simply including it in the return. 

I discovered this comment while working on these components `// TODO: consider making a component for this wrapper span`.  So I refactored `getButtons` to become a `TextInputButtons` component.

## Testing
Regression test the components, they should look and function the same. 

## Checklist


clintonmedbery (approved):
Thanks!

PR #1039, opened 2023-10-01
refactored the modal stories component issue #1005 

Chirag-Nanda:
# Description

The Modal component's stories are configured so that it is wrapped in the <ModalStoryContainer />. The Story Container is used to manage the state and functionality of the component.
The Code option in the component Docs page serves as a reference for how to use the component. With this HOC component wrapping it we cannot see the example of how to use this component.
To fix this, we can move the helper and move that logic into the component story, so that it allows for the Modal component code to fully display.

## Testing

Test out the Modal component

## Screenshots

Before
After

## Checklist


nataliepina:
Hi @Chirag-Nanda thank you for making these changes. I will review them shortly. 
I do see one thing you need to correct here with the commit message. You'll want to update that commit message to include a prefix, something like `docs: refactored the modal stories`. This is outlined in our guide on How to [Contribute](https://github.com/d2iq/ui-kit/wiki/How-to-Contribute). 
Here is some information on [how to change a commit message](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message).

packages/modal/stories/Modal.stories.tsx
```diff
@@ -26,6 +26,7 @@ import { TextInput } from "../../textInput";
 import { fullscreenModalTitle } from "../../fullscreenView/style";
 import { blue } from "../../design-tokens/build/js/designTokens";
 import { SystemIcons } from "../../icons/dist/system-icons-enum";
+import { returnStatement } from "@babel/types";
```
nataliepina, line 29: Looks like you have an accidental unused import here for `returnStatement` that you can remove.

packages/modal/stories/Modal.stories.tsx
```diff
@@ -66,343 +67,458 @@ export default {
...
-  )
+  render: Template1
 };
 
 const confirmIcon = {
   shape: SystemIcons.CircleInformation,
   color: blue
 };
 
+const Template2 = args => {
```
nataliepina, line 104: Instead of all of these Templates, it would be best to try to work around the new way to write stories that use hooks. 
You can read up on it further here in the [Storybook Docs](https://storybook.js.org/docs/react/writing-stories/introduction#working-with-react-hooks).

Let me know if you have any questions!

Chirag-Nanda: Sure thing!

packages/modal/stories/Modal.stories.tsx
Chirag-Nanda: Sure

packages/modal/stories/Modal.stories.tsx
Chirag-Nanda: Can you be more specific about this request ? Should I remove the templates altogether and directly export the React elements? @nataliepina

packages/modal/stories/Modal.stories.tsx
```diff
@@ -11,7 +11,6 @@ import {
   LargeDialogModalWithFooter,
   FullscreenModal
 } from "../../index";
-import ModalStoryContainer from "./helpers/ModalStoryContainer";
```
nataliepina, line 14: Make sure to remove this ModalStoryContainer file now that we don't need it anymore :)

packages/modal/stories/Modal.stories.tsx
nataliepina: I think what you've done now is fine, we will probably migrate to this new documentation style in the future but this aligns with other stories.

packages/modal/stories/Modal.stories.tsx
Chirag-Nanda: Hi @nataliepina , I have made the required changes

nataliepina:
Could you combine your 2 commits to only be one with the correct formatting please?
The second commit message has the correct formatting.

Chirag-Nanda:
The remaining commits are showing conflict errors when squashing , Can you give any way round this problem? @nataliepina

nataliepina:
> The remaining commits are showing conflict errors when squashing , Can you give any way round this problem? @nataliepina

I might suggest that doing a reset, this would be an easier solution, if you do a `git reset HEAD~3`. 
This will set you back all 3 current commits and uncommit those chnages. You should still have the changes but they will not be commited yet. Then you can do a completely new commit with all of those changes under the correct commit message formatting.

Chirag-Nanda:
Complied with the request @nataliepina. Thank you for the help

nataliepina (approved):
Thank you for all of your effort with this work 🌟

Issue #1012, opened 2023-09-25
Refactor FormFieldWrapper components


nataliepina:
## Description
We have a shared component called _FormFieldWrapper_. There are some parts of it that use old patterns that we'd like to update. 
There are these `get` getter type functions that return JSX. Really these should be components that return JSX in accordance with modern patters. 

This would would be to assess those parts of FormFieldWrapper and refactor and improve them. 

For example, this getHintContent, should be a component called HintContent:

```
  const getHintContent = hintContent => {
    return hintContent ? (
      <span
        id={getHintContentId()}
        className={cx(
          tintContentSecondary,
          display("block"),
          padding("top", "xxs"),
          textSize("s")
        )}
      >
        {hintContent}
      </span>
    ) : null;
  };
```

The same thing for getValidationErrors.

With these updates make sure to do regression tests and everything should work in the exact same way. 

<hr> 

### Contribution Guide
If this is your first time contributing to UI Kit, please review the [How to Contribute](https://github.com/d2iq/ui-kit/wiki/How-to-Contribute) guide beforehand.

Chirag-Nanda:
Can I be assigned this issue?

mstruebing:
@Chirag-Nanda sure - I've assigned you to that issue.

Chirag-Nanda:
can you explain a bit more on what exactly is the job needed to be done? Should I replace the entire getHintContent by HintContent?

nataliepina:
> can you explain a bit more on what exactly is the job needed to be done? Should I replace the entire getHintContent by HintContent?

That's correct. `getHintContent` should become a functional component and you can replace it's usage everywhere possible and then remove the getHintContent function once that is completed.  You can isolate the work for getHintContent into one pull request and you can do the validation errors in a separate pull request if you have interest in refactoring that one as well. Let me know if you have any other questions around that all.

Chirag-Nanda:
@nataliepina I have made the changes can you please verify them?

PR #1058, opened 2023-10-12
Refactored the getHintContent to HintContent issue #1012

Chirag-Nanda:
# Description

We have a shared component called FormFieldWrapper. There are some parts of it that use old patterns that we'd like to update.
There are these get getter type functions that return JSX. Really these should be components that return JSX in accordance with modern patters.

This would would be to assess those parts of FormFieldWrapper and refactor and improve them.

For example, this getHintContent, should be a component called HintContent

## Testing

test the FormFieldWrapper component

## Screenshots

CodeBefore
CodeAfter

## Checklist

### DIFF
--- a/packages/shared/components/FormFieldWrapper.tsx
+++ b/packages/shared/components/FormFieldWrapper.tsx
@@ -49,7 +49,7 @@ const FormFieldWrapper = ({
       : "";
   };
 
-  const getHintContent = hintContent => {
+  function HintContent({ hintContent }) {
     return hintContent ? (
       <span
         id={getHintContentId()}
@@ -96,7 +96,7 @@ const FormFieldWrapper = ({
     <>
       {children({
         getValidationErrors: getValidationErrors(errors, formFieldWrapperId),
-        getHintContent: getHintContent(hintContent),
+        getHintContent: HintContent({ hintContent }),
         describedByIds: getDescribedBy(getHintContentId(), getErrorIds()),
         isValid: !errors || (errors && errors.length === 0)
       })}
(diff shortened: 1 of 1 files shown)


packages/shared/components/FormFieldWrapper.tsx
```diff
@@ -49,7 +49,7 @@ const FormFieldWrapper = ({
       : "";
   };
 
-  const getHintContent = hintContent => {
+  function HintContent({ hintContent }) {
```
nataliepina, line 52: This should be a constant called HintContent to follow the pattern of other components in our application.
`const HintContent = ({hintContent}) => {...}`

Will need to define the prop type for hintContent. 

Then below the render props will likely need to be refactored so that it can take in actual components. I'm not sure all that will entail without digging into it.

Chirag-Nanda:
@nataliepina  will this changes do?

nataliepina:
> @nataliepina will this changes do?

Unfortunately this is not quite what we want, as we want to get rid of the use of the components in this manner.

Chirag-Nanda:
Should I remove getHintContent from the children and replace it by the component?

nataliepina:
> Should I remove getHintContent from the children and replace it by the component?

Apologies @Chirag-Nanda I did not fully dig into this issue when creating it and i've taken some time to look into this, it's going to be a really involved change that would likely require refactoring the complete use of FormFieldWrapper. I think it's best if we close this one and table the changes until later on and I can take the time to research and formulate a well documented issue around this. 
I'll try to add some more issues soon for open source contributions that are better defined soon. Cheers!

PR #1086, opened 2023-11-06
build: upgrade to node lts 20.9.0

nataliepina:
Upgrading to the latest version of node - 20.9.0

# Description
This will align with the changes in kommander-ui to bring us to the latest version of node. 

Encountered this error after the upgrade initially:
`TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Object
`
Found a suggested solution [here](https://stackoverflow.com/questions/61879580/nodejs-data-argument-must-be-of-type-string-buffer-typedarray-dataview-how-to) and it worked to include the conversion to a string. 

## Checklist


mstruebing (approved):
Do we use this Dockerfile even anywhere?

nataliepina:
> Do we use this Dockerfile even anywhere?

You know what, no, it looks like was back from the Jenkins set up for CI but with gh actions and gh pages we should be okay here. I've removed this.

PR #1117, opened 2023-12-13
build: node lts

nataliepina:
# Description
This PR is to ensure I am able to push up changes again. 
I updated us to the latest version of node - 20.10.0 

I also removed some the suggested node/npm versions as these are a maintenance burden and can become behind. 
Devs can check these versions in other ways, such as the node-version file. 

## Checklist


mstruebing (approved):
Looks good 👍 

We could move to devbox here as well but as there are not as many tools to set up here - mainly node - it may not worth it.

